PR #37, opened 2015-11-02
No changed files when running tests

jone:
When running tests, I end up with changed .json-files from the `test_documentation` test.
Maybe I was just doing it wrong, but I thought I'd fix what I can and try making the test run without having unnecessary changes.

My changes:
- Fix error when serializing python datetime object.
- Update `event.json` and `collection.json` with new fields.
- Always lowercase response-headers, since they were uppercase on my machine but commit in lowercase.
- Set timezone to `UTC` in `bin/test` (buildout) in order to have the same serialized dates wherever the machine runs.

Not sure if you want those changes, but I'd like to not have changed files when running the tests :wink:

### DIFF
--- a/buildout.cfg
+++ b/buildout.cfg
@@ -30,6 +30,8 @@ eggs =
 [test]
 recipe = zc.recipe.testrunner
 eggs = ${instance:eggs}
+initialization =
+    os.environ['TZ'] = 'UTC'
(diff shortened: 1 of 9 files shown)


tisto:
@jone thanks for the fix! Haven't had time to look into this...

Issue #4, opened 2015-04-17
Use Dexterity Schema and Behaviors to get the Object Schema.


tisto:
Instead of iterating over all possible interfaces, just iterate over the dexterity type schema and all behaviors.

davisagli:
Probably use this: https://github.com/plone/plone.dexterity/blob/master/plone/dexterity/utils.py#L69

tisto:
I created a branch that uses the plone.dexterity helper functions to look up the schema + the behaviors:

https://github.com/plone/plone.restapi/commit/291f8c424f7058dcaab95206d49ec14fa4035a76

It somehow seems title, description, dates, etc. are not looked up properly.

PR #39, opened 2015-11-06
Field serialization adapters

### LINKED ISSUE #38: provide dependency injection for schema serialization
The schema serialization is currently quite hard-coded for the standard fields.
When having custom fields, it is only possible to provide serialization by copying the serialization function and adapting it.

I'd like to change the serialization mechanism so that I'm able to change how a certain field value is represented by registering an adapter and easily provide custom serializer for my custom fields.

Maybe something like that:
```python
class IFieldSerializer(Interface):
    """The field serializer multi adapter serializes the field value into
    a JSON compatible python data. """

    def __init__(context, request, field):
        """Adapts context, request and the field.
        """

    def serialize():
        """Returns JSON compatible python data.
        """
```

Although the request is not really relevant here, I usually like multi adapters to also adapt the request so that I'm able to easily register a more specific adapter by using a (project-) request-layer as discriminator. Without adapting the request I'm more often forced to customize with an `overrides.zcml`, which I do not like. This is my personal taste and experience though :wink: 

Some questions:

1. Is there already some serialization mechanism which produces JSON-compatible output (!) for dexterity?
2. Where should this go? Into `plone.restapi` or into another package?
3. Any comments about my approach?

jone:
Closes #38 

As discussed in #38 this PR introduces field serializer adapters for dexterity based content types in order to make the serialization better customizable and extendable and provide a dependency injection entry point.

Closes #38, #42, #4

#### ``IFieldSerializer``
The field serializer adapter adapts the dexterity field, the context and the request. It's duty is to convert the value of the adapted field on the current context to a data structure which is supported by the JSON module.
This provides the entry point for customizing how a certain field or a type of fields should be represented in the JSON.

#### ``IJsonCompatible``
The json compatibel adapter converts values. It does not know anything about fields and has no context. The idea here is that is possible to recursively convert data. This is especially important for supporting list fields, object fields, data grid fields etc.

#### Discussion

I tried to make as few changes as possible but do a fully working implementation. I'd like to continue with supporting relations (#43), using the ``IJsonCompatible`` converters for Plone site representation too, and more.

Here some notes about this PR:
- I switch to using dexterity's `iterSchemata` and removed our custom code. See #42, #4 
  - non-field values, such as ``meta_type`` are no longer included
  - some fields such as ``effective`` did not use the field value but direct attribute access, which had a different result, which is fixed now
- I have moved image scaling mechanism to a field serializer and moved everything into a dict, acting as a value for this field. This seems more consistent to me.
- I could eliminate the custom field- and image-content-converters (``ISerializeToJson``) and use the general dexterity converter as logic was moved to field serializers.
- Richtext-Values are represented with only the output. We might change that with frames but I think that deserves a separate issue / PR. Discussion needed.


src/plone/restapi/tests/test_field_serializer.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def test_richtext(self):
+        document = createContentInContainer(
+            self.portal, u'Document',
+            text=IRichText['text'].fromUnicode(u'<p>Some Text</p>'))
+        value = self.serialize(document, u'text')
+        self.assertEquals({'raw': u'<p>Some Text</p>',
+                           'mimeType': 'text/html',
+                           'outputMimeType': 'text/x-html-safe',
+                           'encoding': 'utf-8'}, value)
```
jone: How should richtext field be represented in the JSON output?

I see three possibilites:

1. Provide all infos of the richtext object, as above, so that the client knows which mimetype / encoding the value has.
2. Just return the raw output, so that we do not have a nested dict but a string value as with textline fields (e.g. `title`)
3. Same as 2 but convert the string to the `outputMimeType` (transforms).

What do you think is best?

jone:
Builds are broken because of a version conflict for `setuptools`, see https://github.com/plone/plone.restapi/pull/40

src/plone/restapi/tests/test_field_serializer.py
lukasgraf: I don't think 2. really is an option. The client would be lacking necessary information it needs to deal with the data.

I'm leaning towards 1.

jone: I have currently implemented 1., but I think that 3. is better.
Example:
```python
# 1. all infos:
{
 "text": {
   "encoding": "utf-8", 
   "mimeType": "text/plain", 
   "outputMimeType": "text/html", 
   "raw": "If you're seeing this instead of the web site you were expecting, the owner of this web site has just installed Plone. Do not contact the Plone Team or the Plone mailing lists about this."
  }
}

# 3. Converted to configured output type (`text/html` in this case):
{
  "text": "<p>If you're seeing this instead of the web site you were expecting, the owner of this web site has just installed Plone. Do not contact the Plone Team or the Plone mailing lists about this.</p>",
}
```

As an API client I don't care how the data is stored, but I want to have the data as it is meant to be used (which is `text/html` as declared in `outputMimeType`).
A client also may not have a transformation engine to convert from `text/plain` to `text/html`, which may be necessary in the above example in order to have the correct output mimetype.

I think it is better to either just return the transformed value (3.) or extend the current dict (1.) with the converted value.
We probably want to hide the internal data structure (raw value) from the public since it is not relevant, we want them to use the value converted to the output mime type, right?

jone: I think we should go with 3., which is:
```python
{
  "text": "<p>If you're seeing this instead of the web site you were expecting, the owner of this web site has just installed Plone. Do not contact the Plone Team or the Plone mailing lists about this.</p>",
}
```

Discussion:
- `encoding` is not necessary, since the whole document is encoded with the same encoding (`utf8`).
- `raw` is internal, the client does not nedd to know *what* we store
- `mimeType` is internal, the client does not need to know *how* we store it
- `outputMimeType` should be defined in the schema (e.g. `text` is always `text/html`)
- we therefore only require the actual transformed value.

davisagli: If a client wants to fetch the value so it can write an updated value, then it needs the raw data and the stored mimetype, not the transformed data. I agree using the transformed data and outputMimeType is a good default, but we might need an option or different API to get values for editing instead (like the distinction between get and getRaw in Archetypes)

tisto: I would prefer to keep things simple and hide the internal encoding as an implementation detail for now. If a client really needs a different representation it should ask for a different format through the Accept HTTP header (even though we are talking about an embedded format here). We will have an "editing" frame that can provide more information if necessary. This is just the "view/content" frame, right?

davisagli: Okay, if there's a separate frame for editing that's perfect.

David Glick
(mobile)

jone:
I've finished this PR about field serializer.
@tisto @bloodbare @lukasgraf @buchi I'd appreciate your feedback!

src/plone/restapi/serializer/converters.py
```diff
@@ -0,0 +1,114 @@
...
+def default_converter(value):
+    if value is None:
+        return value
+
+    if type(value) in (unicode, bool, int, float, long):
+        return value
+
+    raise TypeError(
+        'No converter for making'
+        ' "{0}" ({1}) JSON compatible.'.format(value, type(value)))
```
lukasgraf: I would probably use the `repr()` of the `value` here:
```
'No converter for making'
' {0!r} ({1}) JSON compatible.'.format(value, type(value)))
```

src/plone/restapi/serializer/converters.py
```diff
@@ -0,0 +1,114 @@
...
+    return IJsonCompatible(value, None)
+
+
+@adapter(Interface)
+@implementer(IJsonCompatible)
+def default_converter(value):
+    if value is None:
+        return value
+
+    if type(value) in (unicode, bool, int, float, long):
```
lukasgraf: Hmm. What about `isinstance(value, (unicode, bool, int, float, long))`? The `json` module seems to be able to serialize subclasses of those types just fine:

```python
>>> import json
>>>
>>> class MyUnicode(unicode): pass
...
>>> value = MyUnicode('foobar')
>>> json.dumps(value)
'"foobar"'
```

lukasgraf: Actually, scratch that - this would mean we would discard information, because those subtypes couldn't be deserialized to the same types any more.

src/plone/restapi/tests/test_field_serializer.py
```diff
@@ -0,0 +1,141 @@
...
+    def test_tuple(self):
+        document = createContentInContainer(self.portal, u'Document')
+        ICategorization(document).subjects = (u'Foo', u'Bar')
+        value = self.serialize(document, u'subjects')
+        self.assertEquals([u'Foo', u'Bar'], value)
+        self.assertEquals(value, json.loads(json.dumps(value)))
+
+    def test_named_image(self):
+        image_asset_path = os.path.join(
+            os.path.dirname(__file__),
```
lukasgraf: Using [`pkg_resources. resource_filename()`](https://pythonhosted.org/setuptools/pkg_resources.html#resource-extraction) would be better here (relying on `__file__` breaks in zipped eggs and some other scenarios). [Example usage](https://github.com/4teamwork/ftw.crawler/blob/52884a1241b8ca99468620bb9c00a46c9b9fb3dd/ftw/crawler/tests/test_configuration.py#L12).

src/plone/restapi/serializer/converters.py
```diff
@@ -0,0 +1,114 @@
...
+from zope.component import adapter
+from zope.component.hooks import getSite
+from zope.interface import implementer
+from zope.interface import Interface
+
+
+def json_compatible(value):
+    """The json_compatible function converts any value to JSON compatible
+    data when possible, raising a TypeError for unsupported values.
+    This happens by using the IJsonCompatible.
```
lukasgraf: `the IJsonCompatible...` converters?

src/plone/restapi/serializer/__init__.py
```diff
@@ -0,0 +1,101 @@
...
+from zope.component import adapter
+from zope.component import getMultiAdapter
+from zope.component import queryMultiAdapter
+from zope.schema import getFields
+
+
+def lookup_field_serializer(context, field):
+    serializer = queryMultiAdapter((field, context, context.REQUEST),
+                                   IFieldSerializer,
+                                   name=field.interface.__identifier__)
```
lukasgraf: Maybe drop three one-line comments here explaining the purpose of this lookup logic?
```
# Look for a serializer that serializes this field type for this specific schema
...
# Try to find a serializer that serializes this field type with this specific name
...
# Fall back on a general purpose serializer for this field type
```

(Or something like that)

docs/source/customization.rst
```diff
@@ -0,0 +1,68 @@
...
+
+
+    @adapter(ICustomField, IDexterityContent, Interface)
+    @implementer(IFieldSerializer)
+    class CustomFieldSerializer(DefaultFieldSerializer):
+
+        def __call__(self):
+            value = self.get_value()
+            if value is not None:
+                value = value.output()
```
lukasgraf: Maybe insert a comment above this line `# Do something with value` to distinguish it a bit from the rest of the code, which is more or less boilerplate.

src/plone/restapi/interfaces.py
```diff
@@ -15,3 +20,21 @@ class IAPIRequest(Interface):
...
+
+
+class IJsonCompatible(Interface):
+    """Convert a value to a JSON compatible data structure.
+    """
+
+
+class IFieldSerializer(Interface):
+    """The field serializer multi adapter serializes the field value into
+    a JSON compatible python data. """
```
lukasgraf: `a JSON compatible` -> `JSON compatible`

src/plone/restapi/serializer/__init__.py
```diff
@@ -0,0 +1,101 @@
...
+                    portal.absolute_url(),
+                    '/'.join(member.getPhysicalPath())
+                ),
+                'title': member.title,
+                'description': member.description
+            }
+            for member in context.results()
+        ]
+
+    for schemata in iterSchemata(context):
```
lukasgraf: `schemata` -> `schema` (singluar). (yep, it's been copy and pasted all over the place like that, for that reason we could also just leave it be)

docs/source/customization.rst
```diff
@@ -0,0 +1,68 @@
...
+
+Content serialization
+=====================
+
+Dexterity fields
+----------------
+
+The API automatically converts all field values to JSON compatible
+data, whenever possible.
+However, you might use fields which store data which cannot automatically
```
lukasgraf: `store data which cannot automatically be converted` -> `store data that cannot be automatically converted,`

src/plone/restapi/tests/test_serializer_converters.py
```diff
@@ -0,0 +1,167 @@
...
+        self.assertEquals(True, json_compatible(True))
+        self.assertEquals('true', json.dumps(json_compatible(True)))
+
+    def test_False(self):
+        self.assertEquals(False, json_compatible(False))
+        self.assertEquals('false', json.dumps(json_compatible(False)))
+
+    def test_unicode(self):
+        self.assertEquals(u'foo', json_compatible(u'foo'))
+        self.assertEquals(unicode, type(json_compatible(u'foo')))
```
lukasgraf: Maybe [`assertIsInstance(json_compatible(u'foo'), unicode)`](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertIsInstance) ?

src/plone/restapi/serializer/__init__.py
lukasgraf: Hmm. `field.interface.__identifier__` is the dotted name of the *schema interface*. But your docs currently state that you can also register a custom serializer using the full dotted name of the *field*, i.e. `plone.app.dexterity.behaviors.exclfromnav.IExcludeFromNavigation.exclude_from_nav`. Shouldn't this therefore be something like
```
'.'.join((field.interface.__identifier__, field.__name__))
```

lukasgraf:
Awesome work! :tada:

Besides the thing with the first adapter lookup in `lookup_field_serializer()` these are really just minor points.

Other than this, looks good to me! :+1:

docs/source/customization.rst
```diff
@@ -0,0 +1,68 @@
+
```
lukasgraf: @jone could you also summarize these changes in a changelog entry?

docs/source/customization.rst
tisto: +1

src/plone/restapi/interfaces.py
```diff
@@ -1,4 +1,9 @@
 # -*- coding: utf-8 -*-
+
```
tisto: Shouldn't we do this globally for the entire project?

jone: The pylint-ignores are only necessary in order to comply with how zope.interface works:
zope.interface requires method declarations to not have the `self` attribute, but pylint isn't happy about that:
> E0211: Method has no argument
> W0221: Arguments number differs from overridden '__call__' method

I think we should not ignore those globally, because in "normal" code those cheks make perfectly sense, but it just does not work with zope.interface; that's why I've ignored them here locally.

src/plone/restapi/serializer/__init__.py
jone: I've refactored the code and adde comments. I have also dropped the simple field name lookup, because I don't think it makes sense.
There is a standard adapter for a type of fields and a specific named adapter with the full dottedname. The full dottedname prevents naming conflicts of fields on multiple behaviors. I think that's enough of dependency injection entry points.

lukasgraf: I agree. I see "specific field" and "entire field type" as the two main customization use cases. If it turns out people actually need something in between, we can always add that later.

src/plone/restapi/tests/test_serializer_converters.py
jone: :+1: :smile:

docs/source/customization.rst
jone: There is not yet a release, should we really write changelogs for the initial release? There is a git log :wink:

jone:
@lukasgraf Thanks for the feedback!
I've updated most things or commented in the discussion :stuck_out_tongue_winking_eye:

docs/source/customization.rst
lukasgraf: Ah, I got `plone.restapi` mixed up with `plone.rest`, which already has an (alpha) release. :+1: for no CL from me :wink:

tisto: We don't need a changelog before we have an initial release...

lukasgraf:
If there are no objections, I'll be merging this one tomorrow.

src/plone/restapi/interfaces.py
tisto: I usually disable those warnings globally because I don't think it is worth the effort. Though, that's just my personal opinion...

tisto:
@jone @lukasgraf I briefly looked over the pr and it looks good to me. Good works guys!!!

I'm ok with merging, we could squash the commits before we do so if you have time...

jone:
:+1: I've squashed everything into one commit.

PR #41, opened 2015-11-08
Implement deserialization

buchi:
Implements deserialization into Dexterity and Archetypes content types.
Two new REST services are provided, allowing to create new content (`FolderPost`) and to update existing content (`ContentPatch`).

Deserialization of JSON data is delegated to an `IDeserializeFromJson` adapter which handles deserialization for a specific content type. Two generic deserializers are provided, one for Dexterity and one for Archetypes content types. Those should be able to handle most content types. For special cases it's possible to provide a more specific adapter.

Deserialization of each field value is delegated to an `IFieldDeserializer` adapter. Field deserializer implementations are provided for all common field types.

There's a decent set of tests using custom sample types that have all supported field types.

Error handling needs some cleanup and unification, but I would like to do that separately as this also involves `plone.rest`.

So in my opinion this on is ready to be merged...

/cc @jone @lukasgraf @tisto @bloodbare


buchi:
@tisto any idea whats wrong with landscape-bot?

tisto:
@buchi It seems landscape is down since a few days. I don't care that much about it. It's just an experiment. If it does not work reliably, we can turn it off again. That's no blocker for a merge...

tisto:
@buchi nice work! Any reason why you did not implement DELETE yet? I will add a few notes inline...

src/plone/restapi/__init__.py
```diff
@@ -1,3 +1,32 @@
 # -*- coding: utf-8 -*-
 from AccessControl import allow_module
 allow_module('json')
+
+from Products.Archetypes.ArchetypeTool import process_types, listTypes
```
tisto: This is a hard dependency on Archetypes. I would really like us to keep Archetypes optional when people start with new projects from the scratch. I don't mind Archetypes support in plone.restapi but I don't want to force people to install legacy packages. :)

src/plone/restapi/__init__.py
```diff
@@ -1,3 +1,32 @@
...
+from Products.CMFCore import permissions
+
+import pkg_resources
+
+try:
```
tisto: @buchi why can't we just use a testing.zcml that we only register for the test fixture here?

src/plone/restapi/configure.zcml
```diff
@@ -15,6 +17,15 @@
       provides="Products.GenericSetup.interfaces.EXTENSION"
       />
 
+  <genericsetup:registerProfile
```
tisto: I prefer to set up testing content types with python code because I think this makes tests far more readable than those obscure zcml profiles. :)

Though, since I don't have much time to help rewriting this code I will just shut up (for now). ;)

buchi:
@tiso DELETE should be an easy one. I wanted to have something merged asap, as this one is already turning into a long running branch, something i don't like. PUT is also missing but I'm not sure if it's really needed, so a low prio one. Next I would like to do some refactoring in the services and serializer parts...

src/plone/restapi/deserializer/dxcontent.py
```diff
@@ -0,0 +1,109 @@
...
+
+        for schema in iterSchemata(self.context):
+            write_permissions = mergedTaggedValueDict(
+                schema, WRITE_PERMISSIONS_KEY)
+
+            for name, field in getFields(schema).items():
+
+                field_data = schema_data.setdefault(schema, {})
+
+                if field.readonly:
```
lukasgraf: Shouldn't an attempt to write to a read-only field raise an error instead of silently discarding the attempted update?

buchi: Right now the behavior is to fail silently if possible, wich is a personal preference of me. This also affects updating an inexistent field or a field you don't have write permission for. I agree that in some cases it may be desirable to have strict error checking. So maybe we should provide an option for strict checking with a `strict=True` URL parameter.

I'd like to discuss error handling separately as it involves other parts than deserialization.
We should also determine first in what format errors should be reported. One of `application/api-problem+json` or `application/vnd.error+json`?

lukasgraf: :+1: for considering a `strict` flag (as a separate change / discussion).

lukasgraf:
Great work! :tada:

I tested this pretty extensively on a local installation, works nicely.

Tested:
- `POST` / `PATCH`
- Validation: Checks for required fields and validation of constraints
- Error reporting in case of validation failures: Works well from a human readability point of view, machine readability will be a different story (needs some more discussion, as you pointed out)
- An example of `NamedBlobFile` creation, including filename, content-type and base64 encoded data
- Roundtrip `GET` -> `POST`. For a DX content type with a moderately complex schema I was able to basically take the JSON response from a `GET`, and feed it back into a `POST` (with minimal modifications), recreating an equivalent object. The only thing I noticed there was that a relation field with no relations will be serialized to `null`, which isn't a valid value for deserialization. Not sure yet whether that should be addressed on the serialization or deserialization side.

:+1: for merging.

lukasgraf:
Since this has been open for a while now, I'm going to go ahead and merge this one.

PR #47, opened 2015-11-20
Refactor tests: Introduce API session in functional tests

lukasgraf:
Introduces a helper class `RelativeSession` that can be used to 

- Set up commonly used HTTP headers like `Accept: application/json` and `Authentication` *once* in the test setup
- Make requests using relative URLs. Avoids the need to always concatenate strings with `portal.absolute_url()`

This then allows to write tests like this

```python
response = self.api_session.get('non-existing-resource')
```

instead of

```python
        response = requests.get(
            self.portal_url + '/non-existing-resource',
            headers={'Accept': 'application/json'},
            auth=(SITE_OWNER_NAME, SITE_OWNER_PASSWORD)
        )
```

which eliminates a ton of repetition and line noise, and makes the tests more expressive.

@tisto @jone @buchi

### DIFF
--- a/src/plone/restapi/testing.py
+++ b/src/plone/restapi/testing.py
@@ -1,13 +1,21 @@
 # -*- coding: utf-8 -*-
+
+# pylint: disable=E1002
+# E1002: Use of super on an old style class
+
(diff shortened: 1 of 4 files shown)


jone:
Awesome :+1:

PR #48, opened 2015-11-26
Search

lukasgraf:
This implements a named `GET` service `/search` that can be invoked on any context to search for Plone content.

Usage examples:

```
GET /plone/search?SearchableText=lorem
GET /plone/search?created.query=1949-01-01&created.query=1951-01-01&created.range=min%3Amax
GET /plone/folder/search?sort_on=path&metadata_fields=_all
```

- Search is **contextual** by default (constrained by path of context, unless an explicit `path` query is specified)
- Nested dictionaries can be specified via **dotted notation** (`path.query=foo&path.depth=1`) similar to ZPublisher's [record arguments](http://docs.zope.org/zope2/zdgbook/ObjectPublishing.html?highlight=record#record-arguments), but without the need for the `:record` suffix
- Optional **type information** (if required because of data type ambiguity) can be supplied in ZPublisher's [argument conversion](http://docs.zope.org/zope2/zdgbook/ObjectPublishing.html#argument-conversion) style (`numeric_index=42:int`), but should rarely be required
- Search results are represented as compact **summaries** by default, additional metadata columns that should be retrieved can be specified via the `metadata_fields` parameter

See the [documentation update](https://github.com/plone/plone.restapi/blob/lg-search/docs/source/searching.rst) included in this PR for more details on user-facing usage.

### DIFF
--- a/docs/source/_json/search.json
+++ b/docs/source/_json/search.json
@@ -1,0 +1,21 @@
+GET /plone/search?sort_on=path
+Accept: application/json
+
+HTTP 200 OK
+content-type: application/json
+
+{
+  "items_count": 2, 
+  "member": [
+    {
+      "@id": "http://localhost:55001/plone/front-page", 
+      "description": "Congratulations! You have successfully installed Plone.", 
+      "title": "Welcome to Plone"
+    }, 
+    {
+      "@id": "http://localhost:55001/plone/robot-test-folder", 
+      "description": "", 
+      "title": "Test Folder"
+    }
+  ]
+}
--- a/docs/source/_json/search.json
+++ b/docs/source/_json/search.json
@@ -5,17 +5,12 @@ HTTP 200 OK
 content-type: application/json
 
 {
-  "items_count": 2, 
+  "items_count": 1, 
   "member": [
     {
       "@id": "http://localhost:55001/plone/front-page", 
       "description": "Congratulations! You have successfully installed Plone.", 
       "title": "Welcome to Plone"
-    }, 
-    {
-      "@id": "http://localhost:55001/plone/robot-test-folder", 
-      "description": "", 
-      "title": "Test Folder"
     }
   ]
 }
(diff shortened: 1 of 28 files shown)


lukasgraf:
@tisto @buchi @bloodbare

This one is ready for review.

FYI: After some back and forth, I decided to drop support for `POST` + JSON body for the following reasons:

- `"There should be one-- and preferably only one --obvious way to do it."`
- JSON only helps to retain *some* typing information - datetimes still need to be dealt with
- No way to build links to search queries with `POST` - not what you want in a hypermedia API
- Implementation is trivial, but we'd still need to test, document and maintain two HTTP verbs instead of one

However, if a real need arises, it would be trivial to add support for `POST` later.

ebrehault:
@lukasgraf, using a GET with querystring parameters is what REST APIs usually do and not POST (one of the reasons is POST requests cannot be cached, as caching mechanisms are always based on the URL).

buchi:
There's a hard dependency on Products.DateRecurringIndex in `query.py` which breaks compatibility with Plone 4.3. Can you make this optional? See also #72 

Didn't have time for a in-depth review yet, will come back later...

lukasgraf:
@buchi will do, thanks for the hint!

lukasgraf:
Made `DateRecurringIndexQueryParser` and rebased onto `master`.

tisto:
Hey @lukasgraf, sorry for messing up your pull request. The build was broken and I tried to fix it by upgrading to the latest Plone versions (4 and 5). It turned out the plone.app.contenttypes fixtures got a major refactoring by @pbauer. I pinned plone.app.contenttypes to older versions for now because I could not solve all the issues and I did not want to further mess up things. It already took me a few hours to figure out what went wrong...I will try to have another look after the merge...

lukasgraf:
@tisto no worries 😉  Let me know if I can help.

PR #55, opened 2016-02-07
[WIP] Test setup with plone5

buchi:


tisto:
@buchi Awesome! Thanks! That was on my todo list for the next days. :)

buchi:
@tisto thanks for merging so fast, maybe a little bit too fast this time :smile: 

There's still one test failure with Plone 5, see https://travis-ci.org/plone/plone.restapi/jobs/107759959
However it's not reported as a failure. Maybe you see why?

The error comes up because in Plone 5 a document has the IShortName behavior, which isn't the case in Plone 4.3. Thus we have different JSON serializations.

tisto:
Travis should be red now. :)

(bin/test-coverage does not return the correct status codes)

PR #57, opened 2016-02-08
Remove undeclared dependency on p.a.contenttypes

### LINKED ISSUE #51: Undeclared dependency on plone.app.contenttypes
The [import of `ICollection` in the serializer code](https://github.com/plone/plone.restapi/blob/17b585a4c76704a0a6f89524f8a87792c519ddfe/src/plone/restapi/serializer/__init__.py#L8) introduced in #39 adds a new, implicit dependency on `plone.app.contenttypes`.

Until now, `plone.app.contenttypes` is [only a test dependency](https://github.com/plone/plone.restapi/blob/17b585a4c76704a0a6f89524f8a87792c519ddfe/setup.py#L46-L47), and I'd really like to keep it that way.

Can we use an alternative to the `ICollection` interface to recognize collections?

lukasgraf:
This removes an undeclared dependency on `plone.app.contenttypes` by making `ICollection` a conditional import.

[`test_serialize_to_json_collection`](https://github.com/plone/plone.restapi/blob/9dc3548077ddaf24ea52344ce1320343bfac9625/src/plone/restapi/tests/test_serializer.py#L174-L214) still passes, and including `plone.restapi` as dependency in a project without `plone.app.contenttypes` now also works.

Closes #51

/cc @tisto

### DIFF
--- a/src/plone/restapi/__init__.py
+++ b/src/plone/restapi/__init__.py
@@ -12,6 +12,14 @@ else:
     REGISTER_TEST_TYPES = True
 
 
+try:
+    pkg_resources.get_distribution('plone.app.contenttypes')
+except pkg_resources.DistributionNotFound:
+    HAS_PLONE_APP_CONTENTTYPES = False
+else:
+    HAS_PLONE_APP_CONTENTTYPES = True
+
+
 def initialize(context):
     if REGISTER_TEST_TYPES:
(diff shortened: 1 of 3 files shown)


tisto:
@lukasgraf looks good to me. We have to fix travis before we can merge though. :(

PR #59, opened 2016-02-09
Fix Plone 5 test by using assertDictContainsSubset

lukasgraf:
Fixes the failing Plone 5 test by using `assertDictContainsSubset` instead of `assertDictEquals`.

We use `assertDictContainsSubset` in order not to have the Plone 5 tests fail because the automatic ID generation for the stock `Document` type works differently in Plone 5 than Plone 4.

A better solution for this (to get stable test behavior across Plone versions) is to create our own FTIs for tests instead of using the stock types from Plone (which may be named the same, but behave differently).

### DIFF
--- a/src/plone/restapi/tests/test_serializer_dexterity.py
+++ b/src/plone/restapi/tests/test_serializer_dexterity.py
@@ -18,7 +18,17 @@ class TestDexteritySerializers(TestCase):
             text=IRichText['text'].fromUnicode(u'<p>Some Text</p>'))
 
         self.maxDiff = None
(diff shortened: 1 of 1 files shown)


tisto:
@lukasgraf +1 for creating our own FTIs. This makes tests a lot more readable IMHO. We should keep some basic tests to actually test the p.a.contenttypes types though.

lukasgraf:
@tisto agreed, I'd still keep *some* tests using the stock p.a.contenttypes, just make them more basic and less brittle.

PR #61, opened 2016-02-11
Check read permissions

casdr:
Closes #56

PR #62, opened 2016-02-12
Serialization refactoring

buchi:
This mainly brings the serialization part more in line with the deserialization part.

- Merge services for getting content into one GET service implementation.
- Check View permission in GET content service.
- `ISerializeToJson` adapters also adapt the request. This makes it easier to customize them and also provides proper access to the request.
- Check read permissions of Dexterity fields (duplicate of #61, already implemented it before I saw it)
- Include metadata not stored in schema fields (creation and modification date, UID) in serialization
- Add missing converters and tests for Dexterity field serialization.
- Implement Archetypes serialization


tisto:
@buchi ready for review? btw: we can just use the labels "in progress", "pr ready". :)

buchi:
@tisto yep, it's ready. not sure if "pr ready" or "pr review" is the right label...

buchi:
@tisto rebased.

tisto:
@buchi thx! That was quick. :)

I just had a quick look over the pr and it looks good to me. Maybe somebody else has time for a second more in-depth review. Will try to take a closer look tomorrow though...

erral:
I have just made a quick test on a Plone 4 site of ours with AT, and aparently looks good. I found two issues though:

- Something happens with some content-types, I found an issue with a Collage page (https://github.com/collective/Products.Collage/blob/master/Products/Collage/content/_collage.py#L28, the traceback is "**<BaseUnit at /eibarkoudala/es/inicio/description> is not JSON serializable**"  The error happens on a folder where the Collage page is, but not for the Collage page itself. 

- Blob types from plone.app.blob are not covered and plone.app.blob is included Plone 4. It should be easy, though, to enable them, just copy the actual FileField and ImageField adapters.

src/plone/restapi/serializer/atcontent.py
```diff
@@ -0,0 +1,66 @@
...
+@adapter(IBaseFolder, Interface)
+class SerializeFolderToJson(SerializeToJson):
+
+    def __call__(self):
+        result = super(SerializeFolderToJson, self).__call__()
+        result['member'] = [
+            {
+                '@id': member.absolute_url(),
+                'title': member.title,
+                'description': member.description
```
erral: The "BaseUnit" error explained in the comment is thrown here, it looks like "member.description" is a BaseUnit instance for Collage types and I don't know if for others too...

Perhaps this two title and descripton can be changed with something like

```python
 'title': member.getField('title').get(member), 
 'description': member.getField('description').get(member)
```

or 

```python
 'title': member.Title(), 
 'description': member.Description()
```

just guessing.

buchi:
@erral Thanks for testing. We definitely should use accessors instead of attribute access for `title`and `description`. I've pushed a fix.

buchi:
@erral Support for Archetypes blobs added.

erral:
I checked it and it works Ok :+1:

lukasgraf:
I also gave this a test locally, looks good to me! :+1: 

Tested:
- View permission is enforced for GET
- `ISerializeToJson` adapter can be customized via browser layer
- Additional metadata (created, modified, UID) is included (tested for DX types)
- Additional relation value converter

PR #68, opened 2016-02-20
Use IContentListing adapters to produce content listings

lukasgraf:
This change unifies the generation of content listings by introducing adapters for the [`IContentListing`](https://github.com/plone/plone.app.contentlisting/blob/06cc9124362aaccb9f63edef7b5069053da1983c/plone/app/contentlisting/interfaces.py#L6-L7) and [`IContentListingObject`](https://github.com/plone/plone.app.contentlisting/blob/06cc9124362aaccb9f63edef7b5069053da1983c/plone/app/contentlisting/interfaces.py#L10-L88) interfaces provided by [`plone.app.contentlisting`](https://github.com/plone/plone.app.contentlisting/#listing-and-working-with-plone-content-objects-using-ploneappcontentlisting).

Using adapters makes the listings themselves and the particularly the summarized representations of objects **customizable**, and avoids code duplication.

This concerns the listings for:

- Children of the site root
- Children of Dexterity containers
- Children of Archetypes containers
- Collection members
- Relations

For everything except Relations the actual representation didn't change - it's just the implementation that has been unified.

In addition to that, I also made the serialization of `RelationValues` use the `IContentListing` adapters, so their serialization now includes title and description as well, and follows the same structure as other content listings.

---

One thing I should point out is that the `RealContentListingObject` implementation in `p.a.contentlisting` [proxies attribute access to the acquisition **unwrapped** object](https://github.com/plone/plone.app.contentlisting/blob/06cc9124362aaccb9f63edef7b5069053da1983c/plone/app/contentlisting/realobject.py#L37). That's fine for attributes, but because the `Title()` and `Description()` accessors aren't implemented on `RealContentListingObject` or its superclass, they will be proxied and therefore won't have an acquisition context (no access to parent object, ...).

This is [different for `getURL()` and `getPath()`](https://github.com/plone/plone.app.contentlisting/blob/06cc9124362aaccb9f63edef7b5069053da1983c/plone/app/contentlisting/realobject.py#L53-L57), which obviously wouldn't work otherwise, and therefore delegate to the respective methods on the acquisition wrapped object.

If this is a concern, we could simply do the same for `Title()` and `Description()`.

### DIFF
--- a/src/plone/restapi/interfaces.py
+++ b/src/plone/restapi/interfaces.py
@@ -56,3 +56,36 @@ class IFieldDeserializer(Interface):
     def __call__(value):
         """Convert the provided JSON value to a field value.
         """
+
+
+class IContentListingSerializer(Interface):
+    """Adapter to serialize an IContentListing into a JSON compatible
+    representation.
+
+    This is used to produce compact content listings that only contain the
+    most basic information about the listed objects.
+    """
+
+    def __init__(listing, request):
+        """Adapts an IContentListing and the request.
+        """
+
+    def __call__():
+        """Returns a JSON compatible Python list containing summarized
+        representations of the objects wrapped by the adapted IContentListing.
+        """
+
+
+class IContentListingObjectSerializer(Interface):
+    """Adapter to serialize an IContentListingObject into a compact JSON
+    compatible representation for use in content listings.
+    """
+
+    def __init__(listing_obj, request):
+        """Adapts an IContentListingObject and the request.
+        """
+
+    def __call__():
+        """Returns a summarized representation of the object wrapped by the
+        adapted IContentListingObject (as JSON compatible Python data).
+        """
--- a/src/plone/restapi/serializer/collection.py
+++ b/src/plone/restapi/serializer/collection.py
@@ -1,11 +1,13 @@
 # -*- coding: utf-8 -*-
+from plone.app.contentlisting.interfaces import IContentListing
 from plone.app.contenttypes.interfaces import ICollection
+from plone.restapi.interfaces import IContentListingSerializer
 from plone.restapi.interfaces import ISerializeToJson
 from plone.restapi.serializer.dxcontent import SerializeToJson
 from zope.component import adapter
 from zope.interface import Interface
+from zope.component import getMultiAdapter
 from zope.interface import implementer
-from zope.site.hooks import getSite
 
 
 @implementer(ISerializeToJson)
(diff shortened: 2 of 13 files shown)


src/plone/restapi/interfaces.py
```diff
@@ -56,3 +56,36 @@ def __init__(field, context, request):
     def __call__(value):
         """Convert the provided JSON value to a field value.
         """
+
+
+class IContentListingSerializer(Interface):
+    """Adapter to serialize an IContentListing into a JSON compatible
+    representation.
+
+    This is used to produce compact content listings that only contain the
```
buchi: Should be also perfectly usable for less compact listings, although I'm not yet sure about the reasonable use cases of having collection members in an object's representation.

src/plone/restapi/serializer/collection.py
```diff
@@ -14,16 +16,19 @@ class SerializeCollectionToJson(SerializeToJson):
...
-                    portal.absolute_url(),
-                    '/'.join(member.getPhysicalPath())
-                ),
-                'title': member.title,
-                'description': member.description
-            }
-            for member in self.context.results()
-        ]
+
+        # XXX: collection.results() claims to return an IContentListing
```
buchi: Ist this still true? Looking at the `ContentListingCollection` view implementation in Plone 5, it simply calls `self.context.results()`. Thus calling the view or calling `results()` directly with the same parameters should give the same result.

lukasgraf: Ah, you're right - the `@@contentlisting` view does indeed call `results()`, but with different defaults for its keyword arguments - specifically `batch=False`. That's why I was seeing different results.

So to summarize:
- `collection.results(batch=False)` returns an `IContentListing`
- `collection.results()` wraps that content listing in a `Batch` before returning it

I'll update the comment, and simplify the code to something along the lines of `listing = self.context.results(batch=False)`

src/plone/restapi/interfaces.py
lukasgraf: That's true, I'll update the docstring accordingly.

tisto:
@lukasgraf what's the status here? Is this ready for a final review/merge?

lukasgraf:
@tisto I have given this some more thought, and I currently believe this might not exactly be the right approach.

The fact that I basically had to "document my way out" of a weird indirection, and the comment by @buchi that "Content Listing" doesn't necessarily imply "Summary / Compact" raised some doubt about this particular implementation.

I'm planning on proposing a different approach that really just focuses on making the *summary* representation unified but customizable through an adapter that directly adapts the context. The fact that content listing objects would be used to access metadata via an uniform API then really would just become an implementation detail.

So I'll close this PR and open up a new one soon that will supersede it.

PR #71, opened 2016-03-29
Make summary representations customizable

lukasgraf:
This **unifies** the implementation for producing **summary representations** of objects, and makes it **customizable** by introducing an `ISerializeToJsonSummary` adapter (adapts `context` and `request`).

In addition to unifying the implementation (which doesn't change the behavior, pure refactoring), I also used the summary representation for serializing relation values - this means their representation now also includes `title` and `description` instead of just the URL.

The `ISerializeToJsonSummary` adapter is used to create representations in the following places:

- Children of the Plone site root
- Children of folderish AT objects
- Children of folderish DX objects
- Parent of AT objects
- Parent of DX objects
- Members of collections
- Relation values

Once search is implemented, this will also be used for catalog brains.

Supersedes #68

### DIFF
--- a/src/plone/restapi/interfaces.py
+++ b/src/plone/restapi/interfaces.py
@@ -22,6 +22,12 @@ class ISerializeToJson(Interface):
     """
 
 
+class ISerializeToJsonSummary(Interface):
+    """Adapter to serialize an object into a JSON compatible summary that
+    contains only the most basic information.
+    """
+
+
 class IJsonCompatible(Interface):
     """Convert a value to a JSON compatible data structure.
     """
--- a/src/plone/restapi/serializer/atcontent.py
+++ b/src/plone/restapi/serializer/atcontent.py
@@ -1,10 +1,13 @@
 # -*- coding: utf-8 -*-
+from Acquisition import aq_inner
 from Acquisition import aq_parent
 from Products.Archetypes.interfaces import IBaseFolder
 from Products.Archetypes.interfaces import IBaseObject
 from plone.restapi.interfaces import IFieldSerializer
 from plone.restapi.interfaces import ISerializeToJson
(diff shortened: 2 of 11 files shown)


lukasgraf:
@tisto @buchi please review - this is ready as far as I'm concerned.

tisto:
@lukasgraf looks good to me!

buchi:
With this approach it's possible to have different representations for different types in the same summary listing. I assume this is an intentional design decision.
I would have expected that it's possible to customize the summary listing for a given type, not for it's summarized items.

Example: We have a folder with two items in it, a document and file.
The default summary looks like this:
```json
"member": [
    {
      "@id": "http://localhost:55001/plone/folder/document", 
      "description": "", 
      "title": "A document"
    }, 
    {
      "@id": "http://localhost:55001/plone/folder/file", 
      "description": "", 
      "title": "A File"
    }
  ]
```

If I want to customize the summary to include the modification date, I have to write adapters for documents and files.

I'm ok with this approach, just not what I expected. :wink:

lukasgraf:
@buchi correct - this is a more limited approach in order to keep it simple for now. Customizing the *listing* for a type right now would require to overwrite the serialization adapter.

In order to be able to customize the listing separately, we would need an adapter that sits in between serialization and the summary representation of a single object, like I attempted in #68 - however, such an adapter wouldn't do anything more than calling the `ISerializeToJsonSummary` adapter for each item in a sequence and build a list of results for now.

Because I currently didn't have a clear use case for that, but do for customizing summary representations on a global level (including `@type` for example), I left that out for now. But I'm happy to add in yet another level of indirection at a later point ;) Just wanted to have something with a clear scope that we can get merged quickly because I'll be needing this for my search PR (catalog brain summaries).

PR #84, opened 2016-05-16
Implement workflow endpoint

buchi:
- `GET /:path/workflow` (get workflow info)
- `POST /:path/workflow/:transition` (trigger workflow transition)

### DIFF
--- a/docs/source/_json/workflow_get.json
+++ b/docs/source/_json/workflow_get.json
@@ -1,0 +1,27 @@
+GET /plone/front-page/workflow
+Accept: application/json
+
+HTTP 200 OK
+content-type: application/json
+
+{
+  "history": [
+    {
+      "action": null, 
+      "actor": "test_user_1_", 
+      "comments": "", 
+      "review_state": "private", 
+      "time": "2016-05-16T13:42:42+00:00"
+    }
+  ], 
+  "transitions": [
+    {
+      "@id": "http://localhost:55001/plone/front-page/workflow/publish", 
+      "title": "Publish"
+    }, 
+    {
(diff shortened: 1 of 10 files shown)


docs/source/_json/workflow_get.json
```diff
@@ -0,0 +1,27 @@
...
+HTTP 200 OK
+content-type: application/json
+
+{
+  "history": [
+    {
+      "action": null, 
+      "actor": "test_user_1_", 
+      "comments": "", 
+      "review_state": "private",
```
lukasgraf: Can the current workflow state only be determined by looking at the most recent history entry? Or am I missing something?

docs/source/workflow.rst
```diff
@@ -0,0 +1,19 @@
+.. include:: /alert-noindex.rst
```
lukasgraf: This shouldn't be included here (inserts the pre-alpha disclaimer and excludes page from google robots)

docs/source/_json/workflow_get.json
buchi: The current workflow state will also be included in the object's GET response.
Regarding the workflow endpoint, i guess getting the complete history is enough. What else would you expect?

lukasgraf:
Some minor points, LGTM otherwise 👍

docs/source/_json/workflow_get.json
lukasgraf: I guess both having in `GET /obj` as well as `GET /obj/workflow` is somewhat redundant. It's just that the current workflow state is one of the infos I would expect from the `workflow` endpoint as a user.

PR #101, opened 2016-05-19
Fix permissions for @types and @registry endpoints

lukasgraf:


buchi:
`cmf.ManagePortal` for registry is fine.
Protecting the `@types` endpoint with `cmf.ManagePortal` makes it pretty useless.

I think we should keep it public, but do some sophisticated security checks in code.
We need to find out if the user is allowed to add/edit content. But these permissions can be granted locally...

Idea: Check if the user has permission to add/edit content on the site root. If not, check if the user has permission to add/edit content on the context defined by the referrer header.

lukasgraf:
As discussed:
Will just check whether user is authenticated or not for the `@types` endpoint, and only expose the information to authenticated users.

lukasgraf:
@buchi updated accordingly and rebased onto master

PR #103, opened 2016-05-20
Service base class

buchi:
Implements a base class for all services defined in plone.restapi that basically renders the results to JSON. This functionality has been removed in plone.rest (https://github.com/plone/plone.rest/pull/52) an now needs to be done in plone.restapi.

Services now have to implement the `reply()` method to process the request. This method should return a JSON serializable data structure or the no content marker.


lukasgraf:
Looks good to me.

PR #106, opened 2016-05-20
Add review_state to content serializations

lukasgraf:
Includes the workflow state in the `review_state` attribute in content serializations.


lukasgraf:
@buchi please review

PR #109, opened 2016-05-20
Authentication

buchi:
Provides a PAS plugin for JWT based authentication.
Provides endpoints `/@login`, `/@logout` and `/@login-renew`

### DIFF
--- a/docs/source/authentication.rst
+++ b/docs/source/authentication.rst
@@ -32,3 +32,90 @@ Or the same example using ``curl``:
 .. code:: python
 
     curl -u username:password -H 'Accept:application/json' $URL
+
+
+JSON Web Tokens (JWT)
+---------------------
+
+``plone.restapi`` includes a Plone PAS plugin for authentication with JWT. The
+plugin is installed automatically when installing the product. 
+
+A JWT token can be acquired by posting a users credentials to the ``@login`` 
+endpoint.
+
+.. example-code::
+
+  .. code-block:: http-request
+
+    POST /@login HTTP/1.1
+    Accept: application/json
+    Content-Type: application/json
+
+    {
+        'login': 'admin',
+        'password': 'admin',
+    }
+
+  .. code-block:: curl
+
+    curl -i \
+    -X POST \
+    -H "Accept: application/json" \
+    -H "Content-type: application/json" \
+    --data-raw '{"login":"admin", "password": "admin"}' \
+    http://localhost:8080/Plone/@login
+
+  .. code-block:: python-requests
+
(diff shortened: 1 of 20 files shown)


tisto:
@sneridagh you already reviewed that pull request, right? Do we have an Angular 2 client implementation from the sprint already?

sneridagh:
@tisto The only thing I miss is the documentation, the rest is good for me.

@buchi could you provide it? Do we have a way to delete expired tokens?

tisto:
@buchi do you have time to finish the docs in the next days? If not it would be great if you could provide us with a few basic examples so we can finish the docs (or maybe just point us to a few tests?).

buchi:
Will add the documentation asap, although I'm quite busy the next week because we (4teamwork) are moving to a new office.

@sneridagh Currently there's no automatic cleanup of stored tokens. I think we don't need that very much, because I see mainly two use cases:

1. JS client authentication: In this case I would not store the token on the server at all, but let it expire after some time (e.g. 12h). That's the default configuration of the PAS plugin.

2. Long term authentication token for apps: In this case the token should never expire, thus we need the ability to invalidate it. For this we have to store the token on the server. You can change the configuration of the PAS plugin to behave like this. The token can be manually deleted by calling the `@logout` endpoint.

sneridagh:
@buchi If you're busy next week I can write it down for you if you wish.

buchi:
@sneridagh I've added minimal documentation now but thanks anyway.

Should be ready for merging now.

lukasgraf:
Tested the described functionality according to the documentation, works like a charm 🎉  (tested login, renewal, invalidating stored tokens, error response when trying to invalidate non-stored token, error response when trying to invalidate tokens twice).

docs/source/authentication.rst
```diff
@@ -31,4 +31,91 @@ Or the same example using ``curl``:
...
+The server returns a JSON object with a new token:
+
+.. literalinclude:: _json/login_renew.json
+   :language: js
+
+
+The ``@logout`` endpoint can be used to invalidate tokens. However by default
+tokens are not persisted on the server and thus can not be invalidated. To enable
+token invaldiation, activate the ``store_tokes`` option in the PAS plugin. If you
+need tokens that are valid indefinitly you should also disalbe the use of Plone's
```
lukasgraf: `indefinitly ` -> `indefinitely`, `disalbe` -> `disable`

lukasgraf:
The renewal endpoint name in the pull request body should be updated from `/@refresh-login` to `/@login-renew`

buchi:
Fixed typos in documentation

Issue #36, opened 2015-10-19
CSRF handling in angular frontends


saily:
After short discussion on #plone i want to raise this up as an issue here. Currently my post-requests to backend fail after upgrading to `plone.protect >= 3.0` because i do not ship a token within the request. 

How do you guys handle this?

buchi:
We simply disable CSRF protection for API requests that write to the database. Because API requests require a specific HTTP header (`Accept: application/json`) we should be safe from CSRF attacks. It shouldn't be possible to set that header in a link.

PR #112, opened 2016-05-24
CSRF Fixes

erral:
I am not completely sure about this pull-request because I don't think I understand correctly the whole CSRF thing, but I see that in the default content-creation service the CSRF check is done, and if I don't do the check to the workflow service too, I can't get the workflow updated.

### DIFF
--- a/src/plone/restapi/services/registry/update.py
+++ b/src/plone/restapi/services/registry/update.py
@@ -2,8 +2,10 @@
 from plone.registry.interfaces import IRegistry
 from plone.restapi.services import Service
 from zope.component import getUtility
+from zope.interface import alsoProvides
 
 import json
+import plone.protect.interfaces
 
 
 class RegistryUpdate(Service):
@@ -11,6 +13,11 @@ class RegistryUpdate(Service):
     def reply(self):
         records_to_update = json.loads(self.request.get('BODY', '{}'))
         registry = getUtility(IRegistry)
+
+        # Disable CSRF protection
+        if 'IDisableCSRFProtection' in dir(plone.protect.interfaces):
+            alsoProvides(self.request,
+                         plone.protect.interfaces.IDisableCSRFProtection)
 
         for key, value in records_to_update.items():
             if key not in registry:
--- a/src/plone/restapi/services/workflow/transition.py
+++ b/src/plone/restapi/services/workflow/transition.py
@@ -8,6 +8,9 @@ from zope.i18n import translate
 from zope.interface import implements
 from zope.publisher.interfaces import IPublishTraverse
 from zope.publisher.interfaces import NotFound
+from zope.interface import alsoProvides
+
+import plone.protect.interfaces
 
 
 class WorkflowTransition(Service):
@@ -41,6 +44,11 @@ class WorkflowTransition(Service):
                 message='Invalid body'))
 
         wftool = getToolByName(self.context, 'portal_workflow')
+
+        # Disable CSRF protection
+        if 'IDisableCSRFProtection' in dir(plone.protect.interfaces):
+            alsoProvides(self.request,
+                         plone.protect.interfaces.IDisableCSRFProtection)
 
         try:
             wftool.doActionFor(self.context, self.transition, **data)


lukasgraf:
Interesting, I can't reproduce a CSRF protection being triggered for either of these endpoints (`PATCH` with `@registry`, `POST` with `@workflow`).

This is with Plone 4.3 and `plone.protect == 3.0.17`. Which versions are you using?

erral:
I have plone.protect == 3.0.16; but could this be related to have plone4.csrffixes? I have it installed (1.0.9).

Anyway, I can check with the latest plone.protect too.

erral:
I checked with plone.protect 3.0.17 and 3.0.18 and the error stands:

```
2016-05-24 20:22:47 ERROR plone.protect Error checking for CSRF. Transaction will be aborted since the request is now unsafe:
  File "/var/csmant/downloads/eggs/plone.protect-3.0.17-py2.7.egg/plone/protect/auto.py", line 177, in check
    return self._check()
  File "/var/csmant/downloads/eggs/plone.protect-3.0.17-py2.7.egg/plone/protect/auto.py", line 201, in _check
    check(self.request, manager=self.key_manager)
  File "/var/csmant/downloads/eggs/plone.protect-3.0.17-py2.7.egg/plone/protect/authenticator.py", line 113, in check
    raise Forbidden('Form authenticator is invalid.')
Forbidden: Form authenticator is invalid.

2016-05-24 20:22:47 ERROR plone.transformchain Unexpected error whilst trying to apply transform chain
  File "/var/csmant/downloads/eggs/plone.transformchain-1.1.0-py2.7.egg/plone/transformchain/transformer.py", line 43, in __call__
    newResult = handler.transformIterable(result, encoding)
  File "/var/csmant/downloads/eggs/plone.protect-3.0.17-py2.7.egg/plone/protect/auto.py", line 158, in transformIterable
    if not self.check():
  File "/var/csmant/downloads/eggs/plone.protect-3.0.17-py2.7.egg/plone/protect/auto.py", line 177, in check
    return self._check()
  File "/var/csmant/downloads/eggs/plone.protect-3.0.17-py2.7.egg/plone/protect/auto.py", line 201, in _check
    check(self.request, manager=self.key_manager)
  File "/var/csmant/downloads/eggs/plone.protect-3.0.17-py2.7.egg/plone/protect/authenticator.py", line 113, in check
    raise Forbidden('Form authenticator is invalid.')
Forbidden: Form authenticator is invalid.

```

My query is a plane POST to the `@workflow` endpoint:

```python
        AUTH = ('user', 'pass')
        HEADERS = {'Accept': 'application/json'}
                                                                                                                          
        url = API_CREATE_URL_ES                                                                                                           
                                                                                                                                              
        # create the content                                                                                                                  
        result = requests.post(url, json=data, auth=AUTH, headers=HEADERS)                                                                    
        if result.status_code == 201:                                                                                                         
            location = result.headers.get('Location')                                                                                         
            API_WORKFLOW_TRANSITION = '{}/@workflow/publish'.format(location)                                                                 
            wfresult = requests.post(API_WORKFLOW_TRANSITION, auth=AUTH, headers=HEADERS)                                                       
```

Moreover, the endpoint returns a HTTP 200 code, with the relevant worfklow information:

```python
(Pdb) result.json()
{u'action': u'publish', u'review_state': u'published', u'comments': u'', u'actor': u'erral', u'time': u'2016-05-24T20:31:38+02:00'}
```

But Plone aborts the transaction later on and the recently created content is still private.

lukasgraf:
There's nothing in `plone.protect` 3.0.16 -> 3.0.17 that could possibly make a difference, no need to retest just for that. `plone4.csrffixes` could very well be involved though, but I also had that included when I did my test (`1.0.8`).

At this point I'm rather wondering why it *doesn't* seem to trigger for me.

At any rate, disabling the protection for requests that cause a DB write *should* be safe (see issue #36), and we're already doing it in other places as you noticed. However, I'd like to investigate this a bit more and get to the bottom of why this wasn't an issue while we implemented those endpoints during the sprint, and why I'm currently not able to reproduce this.

But in principle I don't see a reason we couldn't merge this fix. However, it would be great if you could update your commit messages (using `git` interactive rebase): They currently say "add CSRF" check, but what they do is actually *disable* the CSRF check 😉

lukasgraf:
@erral ah, now I can reproduce it with the workflow endpoint - the `200 OK` response fooled me, I didn't look closely enough. If you'd like I can do the rebasing for you and push to this branch & merge.

erral:
OK, yes please.

lukasgraf:
@erral done, will merge as soon as tests have finished running. Thanks for your fix and the detailed info 😃

PR #113, opened 2016-05-25
Serializers for Archetypes new style collections

erral:
Archetypes based plone.app.collection serializer was missing and thus the QueryField serializer for Archetypes too.


buchi:
Thanks!

Issue #10, opened 2015-04-17
Implement Image API


tisto:
```
{
    contentType
    size
    data/download
    filename
    scale: {
        'mini': {
            href: ...
            width: 400,
            height: 200,
        }
    }
}
```

lukasgraf:
Image fields currently get serialized as a mapping of `scale_name: download_url` (see docs).

I would propose the following serialization format instead:

```json
{
  "@id": "http://nohost/plone/my-image",
  "@type": "Image",
  "title": "My Image",
  "...": "...",
  "image": {
    "filename": "image.png",
    "content-type": "image/png",
    "size": 1185,
    "download": "http://nohost/plone/my-image/@@images/image",
    "scales": {
      "icon": "http://nohost/plone/my-image/@@images/image/icon",
      "large": "http://nohost/plone/my-image/@@images/image/large",
      "listing": "http://nohost/plone/my-image/@@images/image/listing",
      "mini": "http://nohost/plone/my-image/@@images/image/mini",
      "original": "http://nohost/plone/my-image/@@images/image",
      "preview": "http://nohost/plone/my-image/@@images/image/preview",
      "thumb": "http://nohost/plone/my-image/@@images/image/thumb",
      "tile": "http://nohost/plone/my-image/@@images/image/tile"
    }
  }
}
```

This doesn't include the dimensions for each scale as in your initial suggestion, but including those would make the response really rather bulky. I don't feel strongly about this, but I probably wouldn't include those until we have framing in place and a client can request to have them omitted.

tisto:
Closed by https://github.com/plone/plone.restapi/pull/115

PR #115, opened 2016-05-26
Change file/image field serializers to return a dict with additional metadata

lukasgraf:
The Archetypes part of this needs a thorough review, I was struggling slightly with the different File / Image field type's APIs.

---

This changes the serialization format for file and image fields from returning a simple string to returning dictionaries that contain the download URL in a `download` key, according to #9 and #10.

There's also a change in behavior regarding **empty file / image fields**: Previously, those were still serialized to an URL that would have returned a 404 (or 500, not exactly sure). Now they'll be serialized to `None`.


src/plone/restapi/tests/test_atfield_serializer.py
```diff
@@ -70,12 +73,18 @@ def test_lines_field_serialization_returns_list(self):
...
-                               filename='spam.txt', mimetype='text/plain')
-        self.assertTrue(isinstance(value, unicode), 'Not an <unicode>')
-        self.assertEqual(u'http://nohost/plone/doc1/@@download/testFileField',
-                         value)
+                               mimetype='text/plain')
+        self.assertTrue(isinstance(value, dict), 'Not a <dict>')
+
+        url = u'http://nohost/plone/doc1/@@download/testFileField'
+        self.assertEqual(
+            {u'filename': '',
```
lukasgraf: This part confused me - I wasn't able to set a filename for an `atapi.FileField` for the life of me. Does this field type actually support filenames? I wasn't able to find a clear reference for or against that. @buchi you probably know this.

tisto:
@lukasgraf Awesome!!! I will make a release as soon as this pr has been merged!

src/plone/restapi/tests/test_atfield_serializer.py
erral: To set the filename, you need to do (also works for AT blob fields):

```python
file.getField('file').get(file).setFilename('newname.txt')
```

lukasgraf: @erral thank you, that works. Though after looking at it again, it also seems to work for that field type by passing a [keyword argument to the mutator](https://github.com/plone/plone.restapi/pull/115/files/1d49315bd5533d6640a7f536a406ea76b800a8f6#diff-06974b9abd48f281685ec4494e725f08R28) - which I somehow couldn't get to work at first for that field type.

buchi:
I strongly recommend to include the image dimensions in the output. Without the dimensions it's not possible to render "good" image tags.

See also Google Page Speed Guide: "If no dimensions are specified in the containing document, or if the dimensions specified don't match those of the actual images, the browser will require a reflow and repaint once the images are downloaded. To prevent reflows, specify the width and height of all images, either in the HTML tag, or in CSS."

src/plone/restapi/tests/test_atfield_serializer.py
erral: We've been hit several times by the filenames (specially when running large imports), I think it's a bit obscure :wink:

docs/source/_json/image.json
```diff
@@ -20,14 +20,20 @@ content-type: application/json
...
-    "icon": "http://localhost:55001/plone/image/@@images/image/icon", 
-    "large": "http://localhost:55001/plone/image/@@images/image/large", 
-    "listing": "http://localhost:55001/plone/image/@@images/image/listing", 
-    "mini": "http://localhost:55001/plone/image/@@images/image/mini", 
-    "original": "http://localhost:55001/plone/image/@@images/image", 
-    "preview": "http://localhost:55001/plone/image/@@images/image/preview", 
-    "thumb": "http://localhost:55001/plone/image/@@images/image/thumb", 
-    "tile": "http://localhost:55001/plone/image/@@images/image/tile"
+    "content-type": "image/png", 
+    "download": "http://localhost:55001/plone/image/@@images/image",
```
buchi: I would only include either the "download" or the "original" link, because both point to the same url.

lukasgraf:
Ok, I'll update the PR to also include image dimensions with the scales.

docs/source/_json/image.json
lukasgraf: Ok, I'll drop the `original` scale.

tisto:
@buchi +1 for including the image dimensions. I had the same feeling, but did not want to complain before looking into it. :)

As soon as we have framing in place we can minimize the output if we need to...

lukasgraf:
@buchi @tisto updated:
- Also got setting filename for `atapi.FileField` via mutator kwargs to work, simplified mutator call (always pass kwargs) (*squashed*)
- Dropped `original` scale from list of scales (*squashed*)
- Included dimensions in image scales, updated docs & JSON dumps accordingly
- Factored out the scale / imaging related helper functions into `plone.restapi.imaging`, since duplication got out of hand

src/plone/restapi/serializer/atfields.py
```diff
@@ -38,7 +39,13 @@ def __call__(self):
         url = '/'.join((self.context.absolute_url(),
                         '@@download',
                         self.field.getName()))
-        return json_compatible(url)
+        result = {
+            'filename': self.field.getFilename(self.context),
+            'content-type': self.field.getContentType(self.context),
+            'size': self.field .get_size(self.context),
```
buchi: unnecessary space after `self.field`

src/plone/restapi/serializer/atfields.py
```diff
@@ -58,10 +65,21 @@ def __call__(self):
...
+        if image is None:
+            return None
+
+        download_url = '/'.join((
+            self.context.absolute_url(), '@@images', self.field.__name__))
+        scales = get_scales(self.context)
+        result = {
+            'filename': self.field.getFilename(self.context),
+            'content-type': self.field.getContentType(self.context),
+            'size': self.field .get_size(self.context),
```
buchi: unnecessary space after `self.field`

buchi:
The returned image dimensions should be the real dimensions of the scaled image and not the scales defined in image scaling. E.g. when scaling an image of size 500 x 400 to an image scale of 200 x 200 the resulting image will have the size 200 x 160, not 200 x 200 because the scaling is maintaining aspect ratio.

lukasgraf:
@buchi good point, I'll look into this.

lukasgraf:
@buchi @tisto updated:

- Included the original image's width and height as top-level attributes
- Return the actual dimensions of the scales (maintaining aspect ratio)
- Updated docs and JSON dumps
- Used a test image closer to real life in tests to demonstrate aspect ratio handling
- Fixed unnecessary spaces

---

Using the ``@@images`` ImageScaling implementations for getting at the actual scale
dimensions would in theory be possible, but has some downsides:

 - Accessing a scale might trigger on-demand rendering, killing performance and possibly causing writes on GET.

   I haven't had the time to definitely confirm this, but depending on what ``ImageScaling`` implementation is being used, accessing a scale might not be free of side-effects.

- Different ``ImageScale`` implementations seem to have slightly different APIs, and while a first attempt at using it seemed to work, getting a clean, abstracted solution for the different field types, Archetypes vs. Dexterity and Plone 4 vs 5 turned out to be non-trivial.

I briefly discussed this with @buchi, and I went with a solution that basically just **applies the aspect ratio of the original image to the scale's (usually square) bounding box**, which should result in the final scaled image's actual dimensions.

There's a couple boundary conditions to handle, but nothing too major:

- How to deal with fractional pixel dimensions - truncate or round. Plone seems to truncate, so I did the same.

- Don't produce dimensions with zero pixels on one side as a result of integer truncation

- How to deal with scales that exceed the original image's dimensions:
  For the default named image scales, Plone doesn't upscale images, it just returns the image in it's original dimensions. We do the same.

docs/source/serialization.rst
```diff
@@ -48,32 +48,64 @@ will be serialized to
...
+      },
+      "large": {
+        "download": "http://localhost:55001/plone/image/@@images/image/large",
+        "height": 768,
+        "width": 576
+      },
+      "..." : {}
+    }
+
+(For non-square images, those will be the actual scaled images' dimensions,
```
buchi: I would simply drop this remark because maintaining aspect ratio is not related to square or non-square images. One could also define non-square image sizes (e.g. 600 x 400).

lukasgraf:
@buchi updated

tisto:
Wohoo! I will go over the docs one last time and do some final checks and then make a first alpha release!

PR #117, opened 2016-06-03
Use latest Plone versions in tests

buchi:
- Update test setup to use Plone 5.0.4 and Plone 4.3.9 without downpinning p.a.contenttypes
- Create `SITE_OWNER_NAME` user in test layer (was previously done in p.a.contenttypes)
- Grant `Manager` role to `TEST_USER_ID` (was previously done in p.a.contenttypes)


tisto:
@buchi thanks for the fix!!!

PR #121, opened 2016-06-19
Basic implementation of a @users endpoint.

tisto:
@lukasgraf I guess it would make sense to create a serializer for users instead of doing this in the views, right?


lukasgraf:
@tisto without having looked at the exact implications of this - yes, I'd say so.

docs/source/_json/users_created.json
```diff
@@ -0,0 +1,16 @@
+POST /plone/@users
+Accept: application/json
+
+HTTP 201 Created
+content-type: application/json
+location: http://localhost:55001/plone/@users/noamchomsky
+
+{
+  "description": "", 
+  "email": "noam.chomsky@mit.edu",
```
lukasgraf: Hmm - use `...@example.org` instead of real domain names / addresses?

docs/source/users.rst
```diff
@@ -0,0 +1,163 @@
...
+Read User
+---------
+
+To retrieve all details for a particular user, send a GET request to the ``/@users`` endpoint and append the user id to the URL:
+
+.. example-code::
+
+  .. code-block:: http-request
+
+    GET /@user/noam HTTP/1.1
```
lukasgraf: `@user` (example request) vs. `@users` (documentation just above)

docs/source/users.rst
```diff
@@ -0,0 +1,163 @@
...
+The server will respond with a 200 OK status code and the JSON representation of the user in the body:
+
+.. literalinclude:: _json/users_noam.json
+   :language: js
+
+
+Update User
+-----------
+
+To update the settings of a usersend a PATCH request with the user details you want to amend to the URL of that particular user, e.g. if you want to update the email address of the admin user to:
```
lukasgraf: Typo: `usersend`

docs/source/_json/users.json
```diff
@@ -0,0 +1,26 @@
+GET /plone/@users
+Accept: application/json
+
+HTTP 200 OK
+content-type: application/json
+
+[
```
lukasgraf: No JSON-LD? Ideally these responses would follow the same structure as other collection-like endpoints, including batched results and having a dereferenceable JSON-LD `@id`, no?

src/plone/restapi/services/users/get.py
```diff
@@ -0,0 +1,51 @@
...
+            return [
+                {
+                    'id': user.id,
+                    'email': user.getProperty('email'),
+                    'username': user.getUserName(),
+                    'fullname': user.getProperty('fullname'),
+                    'home_page': user.getProperty('home_page'),
+                    'description': user.getProperty('description'),
+                    'location': user.getProperty('location')
+                } for user in api.user.get_users()
```
lukasgraf: There's a caveat here - Zope's `Products.PluggableAuthenticationService` does not support an interface that *lists* all users, it only provides an interface [`IUserEnumerationPlugin`](https://github.com/zopefoundation/Products.PluggableAuthService/blob/master/Products/PluggableAuthService/interfaces/plugins.py#L292) that can be used to *query / search* users.
`PlonePAS` on the other hand introduces [`IUserIntrospection`](https://github.com/plone/Products.PlonePAS/blob/master/src/Products/PlonePAS/interfaces/plugins.py#L6) which allows for a complete listing, and which is what `plone.api.user.get_users()` uses.

Unfortunately, `Products.LDAPUserFolder` does not implement an `IUserIntrospection` plugin, which is why on a site with all your users in LDAP, `api.user.get_users()` will return an empty list.

From what I understood when I looked into this surprising (to me) behavior a while ago, this was a deliberate decision. Enumerating all users can, depending on the backend and amount of users, absolutely kill performance, and that's why it was considered "an API that shouldn't be used, and therefore isn't provided" by the authors of `Products.PluggableAuthService` (IIRC). There's a rather heated discussion on the details of this [here (dzug ML, german)](http://mail.dzug.org/mailman/archives/zope/2006-December/002431.html).

I still think exposing the behavior of `plone.api` is the right thing to do, but we might want to point out this caveat in the documentation.

src/plone/restapi/tests/test_users.py
```diff
@@ -0,0 +1,100 @@
...
+        )
+        transaction.commit()
+
+    def test_list_users(self):
+        response = self.api_session.get('/@users')
+
+        self.assertEqual(200, response.status_code)
+        self.assertEqual(3, len(response.json()))
+        user_ids = [user['id'] for user in response.json()]
+        self.assertTrue('admin' in user_ids)
```
lukasgraf: Use `assertIn('admin', user_ids)` -> better assertion failure message than `False is not True` 😉

lukasgraf:
@tisto some first comments - haven't had time to do an in-depth review yet.

I moved my comments from the closed PR to this one, sorry for the double notifications.

tisto:
@lukasgraf fyi: I refactored the code and added a user serializer.

docs/source/users.rst
tisto: Fixed.

docs/source/users.rst
tisto: Fixed.

docs/source/_json/users.json
tisto: @lukasgraf of course! I started with a proof-of-concept implementation. Sorry, should have marked this PR as WIP.

docs/source/_json/users_created.json
tisto: Right, fixed.

src/plone/restapi/services/users/get.py
tisto: @jensens @sneridagh opinions?

docs/source/_json/users.json
tisto: @lukasgraf I added the @id attribute to the responses. I did not want to implement batching because I'm aware of the problem with get users you mentioned below. My idea was to push a basic implementation and then discuss the details. I should have made that more clear. :)

lukasgraf: @tisto I see, no worries ;-) Are you planning on merging the basic implementation for `1.0a1` already? Because I feel we've got a pretty stable state in `master` right now, and that's what I recently pushed to production. So I'd love it if we could lock in that state by doing a first release, and then track further changes with a proper change log.

docs/source/users.rst
```diff
@@ -0,0 +1,163 @@
...
+    Host: localhost:8080
+    Accept: application/json
+    Content-Type: application/json
+
+    {
+        'username': 'noam',
+        'email': 'noam.chomsky@mit.edu',
+        'password': 'colorlessgreenideas',
+        'username': 'noamchomsky',
+        'fullname': 'Noam Avram Chomsky',
```
lukasgraf: Noticed during testing: Properties like `fullname` and `home_page` need to be passed as a separate dictionary under the `properties` key - also see `test_documentation_users_created` and the corresponding output in `users_created.json`.

In the `GET` representation however these fields are all serialized to a flat structure, so passing back in what came out doesn't currently result in the same object. It would be nice if we could get a more consistent behavior here. Not sure though how much it's `plone.restapi`'s responsibility to abstract away Plone (PAS)'s layered member property concepts.

lukasgraf:
@tisto changes look good, and I gave this a test drive on a local setup (with `LDAPUserFolder`). Works well 👍  - was able to do `GET` and `PATCH` for LDAP users, `POST` for adding a user to Plone's local `source_users`, and `DELETE` to remove it again.

- As anticipated, when listing users with `GET`, those from LDAP were omitted. Others worked fine.
- `DELETE` should probably `404` for a non-existing user. Currently returns `204 No content` unconditionally.
- `GET` should `404` for a non-existing user. Currently fails with a `TypeError: 'NoneType' object is not callable`. Same for `PATCH`.

docs/source/users.rst
tisto: @lukasgraf Yes. That was a conscious design decision. A REST API should not expose implementation details of the underlying system IMO. I'm not sure why plone.api did not abstact that away already.

I guess PAS is storing those params separately because you might want to store the credentials separated from the user properties sometimes (e.g. if you want "users" as objects). I'd consider that an implementation detail that should not be exposed to the user of an API. 

@sneridagh @jensens opinions?

@zupo do recall why for plone.api it was decided to pass the "properties" param when creating a user instead of providing all user properties as a flat structure (named params)?

tisto:
@lukasgraf thankt for the review! I was planing to work on the edge cases once we got the basic implementation ready. Still WIP (excuses, excuses... ;)).

docs/source/users.rst
lukasgraf: @tisto
I agree with not exposing that implementation detail through the API.

sneridagh: @tisto @lukasgraf I also agree with do it flat. We already follow it in other endpoints during the serialization process.

jensens: +1 for not exposing PAS implementation details. 

If it comes to userdata-as-json I like like how Authomatic handles the different data returned by several OAuth2 providers. https://peterhudec.github.io/authomatic/reference/providers.html#oauth2-providers

buchi:
@tisto your implementation is depending on plone.api. If I remember correctly we decided to not depend on plone.api because of possible circular dependencies if plone.restapi is included in the core. Is this still something we have to keep in mind?

tisto:
@buchi absolutely! I just needed a quick solution for a client. I will remove the plone.api dependency before I merge this pr.

tisto:
@lukasgraf @buchi @sneridagh I removed the plone.restapi dependency from the code (it is still a test dependency though). Please do a final review and merge.

@lukasgraf @buchi btw: are you guys going to be in Boston?

lukasgraf:
@tisto we are, we just booked our flights this morning 😄  I'd definitely be up for a quick ad-hoc sprint, or at least some restapi-beers 😉 

I'll try to do another review ASAP, though it's at least going to be later tonight until I get around to it.

tisto:
@lukasgraf sounds great! :)

PR #122, opened 2016-06-19
Users endpoint

tisto:


docs/source/_json/users_get.json
```diff
@@ -0,0 +1,15 @@
+GET /plone/@users/noam
+Accept: application/json
+
+HTTP 200 OK
+content-type: application/json
+
+{
+  "description": "Professor of Linguistics", 
+  "email": "noam.chomsky@mit.edu",
```
lukasgraf: Hmm - use `...@example.org` instead of real domain names / addresses?

docs/source/users.rst
```diff
@@ -0,0 +1,163 @@
...
+Read User
+---------
+
+To retrieve all details for a particular user, send a GET request to the ``/@users`` endpoint and append the user id to the URL:
+
+.. example-code::
+
+  .. code-block:: http-request
+
+    GET /@user/noam HTTP/1.1
```
lukasgraf: `@user` (example request) vs. `@users` (documentation just above)

lukasgraf: `@user` (example request) vs. `@users` (documentation just above)

docs/source/users.rst
```diff
@@ -0,0 +1,163 @@
...
+The server will respond with a 200 OK status code and the JSON representation of the user in the body:
+
+.. literalinclude:: _json/users_noam.json
+   :language: js
+
+
+Update User
+-----------
+
+To update the settings of a usersend a PATCH request with the user details you want to amend to the URL of that particular user, e.g. if you want to update the email address of the admin user to:
```
lukasgraf: `usersend` -> `user, send`

docs/source/_json/users.json
```diff
@@ -0,0 +1,26 @@
+GET /plone/@users
+Accept: application/json
+
+HTTP 200 OK
+content-type: application/json
+
+[
```
lukasgraf: No JSON-LD? Ideally these responses would follow the same structure as other collection-like endpoints, including batched results and having a dereferenceable JSON-LD `@id`, no?

src/plone/restapi/services/users/get.py
```diff
@@ -0,0 +1,51 @@
...
+            return [
+                {
+                    'id': user.id,
+                    'email': user.getProperty('email'),
+                    'username': user.getUserName(),
+                    'fullname': user.getProperty('fullname'),
+                    'home_page': user.getProperty('home_page'),
+                    'description': user.getProperty('description'),
+                    'location': user.getProperty('location')
+                } for user in api.user.get_users()
```
lukasgraf: There's a caveat here - Zope `PAS` does not support an interface that *lists* all users, it only provides an interface [`IUserEnumerationPlugin`](https://github.com/zopefoundation/Products.PluggableAuthService/blob/master/Products/PluggableAuthService/interfaces/plugins.py#L292) that can be used to *query / search* users.
`PlonePAS` on the other hand introduces [`IUserIntrospection`](https://github.com/plone/Products.PlonePAS/blob/master/src/Products/PlonePAS/interfaces/plugins.py#L6) which allows for a complete listing, and which is what `plone.api.user.get_users()` uses.

Unfortunately, `Products.LDAPUserFolder` does not implement an `IUserIntrospection` plugin, which is why on a site with all your users in LDAP, `api.user.get_users()` will return an empty list.

From what I understood when I looked into this surprising (to me) behavior a while ago, this was a deliberate decision. Enumerating all users can, depending on the backend and amount of users, absolutely kill performance, and that's why it was considered "an API that shouldn't be used, and therefore isn't provided" by the authors of `Products.PluggableAuthService` (IIRC). There's a rather heated discussion on the details of this [here (dzug ML, german)](http://mail.dzug.org/mailman/archives/zope/2006-December/002431.html).

I still think exposing the behavior of `plone.api` is the right thing to do, but we might want to point out this caveat in the documentation.

src/plone/restapi/tests/test_users.py
```diff
@@ -0,0 +1,100 @@
...
+        )
+        transaction.commit()
+
+    def test_list_users(self):
+        response = self.api_session.get('/@users')
+
+        self.assertEqual(200, response.status_code)
+        self.assertEqual(3, len(response.json()))
+        user_ids = [user['id'] for user in response.json()]
+        self.assertTrue('admin' in user_ids)
```
lukasgraf: Use `assertIn('admin', user_ids)` -> better assertion failure message than `False is not True` 😉

PR #130, opened 2016-07-31
get components implementation from barcelona-mocks + add tests

ebrehault:


tisto:
@ebrehault please rebase. :)

ebrehault:
@tisto done

tisto:
@ebrehault Merci! Will have a look and do a review later today.

lukasgraf:
Could we get some documentation for this?

src/plone/restapi/services/components/get.py
```diff
@@ -0,0 +1,73 @@
...
+class ComponentsGet(Service):
+
+    implements(IPublishTraverse)
+
+    def __init__(self, context, request):
+        super(ComponentsGet, self).__init__(context, request)
+        self.params = []
+
+    def publishTraverse(self, request, name):
+        # Consume any path segments after /components_ as parameters
```
lukasgraf: That would be `/@components` now

src/plone/restapi/services/components/get.py
```diff
@@ -0,0 +1,73 @@
...
+        for component_id in name.split(','):
+            self.params.append(component_id)
+        return self
+
+    @property
+    def _component_ids(self):
+        return self.params
+
+    def _frame_component_items(self, items, component_id):
```
lukasgraf: I hurriedly picked the name for this method during the sprint, where the aim was to just cobble together a quick mock. Now I think it might cause confusion with the concept of JSON-LD framing. Maybe `_wrap_component_items()`?

src/plone/restapi/services/components/get.py
ebrehault: true

src/plone/restapi/services/components/get.py
ebrehault: you are right, that's confusing

ebrehault:
@lukasgraf I have added the doc

lukasgraf:
@ebrehault awesome, thank you! 🎉  Docs look good to me. 👍

tisto:
First thing that comes to mind when reviewing this pr: We need to restrict the number of languages in the test setup and we have to get rid of the workflow timestamp. Those two things will be included in every pull request we review. This has nothing to do with this pr itself of course.

tisto:
@ebrehault fyi: I rebased your branch. Is this still work in progress or ready for review/merge?

docs/source/_json/breadcrumbs.json
```diff
@@ -0,0 +1,19 @@
+GET /plone/front-page/@components/breadcrumbs
+Accept: application/json
+
+HTTP 200 OK
+content-type: application/json
+
+[
+  {
+    "data": {
```
tisto: @ebrehault is there a specific reason why "items" is wrapped into "data"? Shouldn't we be consistent with our batching/folder/collection API here? cc @lukasgraf

tisto:
@ebrehault Apart from the open question regarding items/data this pull request looks good to me and can be merged.

ebrehault:
It is not work in progress, but let's wait news from @lukasgraf regarding the data/items thing before merging.

docs/source/_json/breadcrumbs.json
lukasgraf: @ebrehault @tisto Regarding wrapping "items" into "data", that really was just an ad-hoc decision during the sprint when I cobbled together that mock implementation. Someone from the frontend team requested a mock for breadcrumbs, and someone else needed one for navigation. The expected example responses they gave me varied quite a bit in terms of format, but not in semantics. So I tried to identify a common outer frame that they shared, and what you're seeing here is what we went with 😉 

I can't imagine anyone from the frontend team is attached in any way to this particular structure. I'd say eliminate `data`. As for making it consistent with the overall structure of collection-like resources: I'd say "yes, where it makes sense". So let's also have a top-level `items`, maybe later `items_total` if it serves any purpose. But I wouldn't implement actual batching, IMHO that would just make our lives unnecessarily hard (we also didn't do that for relations).

JSON-LDfication? Maybe `id` could be promoted to `@id` and contain the fully qualified URL of the endpoint, which would still carry the same information. I simply don't know what `id` is being used for right now on the frontend, if at all.

So I would probably eliminate `data`, but I don't feel strongly about any of the other points.

docs/source/_json/breadcrumbs.json
```diff
@@ -0,0 +1,19 @@
...
+      "items": [
+        {
+          "title": "Welcome to Plone", 
+          "url": "http://localhost:55001/plone/front-page"
+        }
+      ]
+    }, 
+    "id": "breadcrumbs"
```
tisto: @ebrehault as @lukasgraf mentioned I would recommend to use the JSON-LD "@id" notation with the URL as value here. Do we use "id" for anything?

ebrehault:
ok, the @@components response format is fixed

docs/source/_json/breadcrumbs.json
```diff
@@ -6,14 +6,12 @@ content-type: application/json
...
-    "data": {
-      "items": [
-        {
-          "title": "Welcome to Plone", 
-          "url": "http://localhost:55001/plone/front-page"
-        }
-      ]
-    }, 
-    "id": "breadcrumbs"
+    "@id": "breadcrumbs",
```
tisto: @ebrehault if we use "@id" the value needs to contain the full URL of the object. Otherwise we have to stick with "id".

ebrehault:
@tisto the @id value is fixed

PR #135, opened 2016-08-03
Tests: Use `freezegun` to freeze hard to control timestamps

### LINKED ISSUE #133: Set workflow timestamps in tests to a specific date/time.
This is also annoying because every commit or pull request will update that file:

https://github.com/plone/plone.restapi/pull/131/files#diff-aacab0b0e85d224bdb693ec2f6cc0036L14

Any takers?

lukasgraf:
Use [`freezegun`](https://github.com/spulec/freezegun) to freeze hard to control timestamps in response dumps used for documentation.

This really just addresses the constantly changing timestamps in the `@workflow` response dumps. I considered rewriting all the other places where we set creation / modification dates manually, but I don't think it currently adds any value - so I suggest to just use `freezegun` to target the timestamps that would otherwise be hard to control.

@tisto
 
Closes #133

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,10 @@ Changelog
 1.0a2 (unreleased)
 ------------------
 
+- Tests: Use `freezegun` to freeze hard to control timestamps in
+  response dumps used for documentation.
+  [lgraf]
+
 - Tests: Limit available languages to a small set to avoid excessive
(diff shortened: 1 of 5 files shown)


tisto:
@lukasgraf thanks again! Freezegun looks like a very good solution to those kind of problems.

PR #140, opened 2016-08-09
Disable automatic CSRF protection for @login and @login-renew endpoints

lukasgraf:
This disables automatic CSRF protection for the `@login` and `@login-renew` endpoints.

If persisting tokens server-side ("Store Tokens" option on the `jwt_auth` Plugin) is enabled, those requests need to be allowed to cause DB writes, otherwise the transaction might be aborted by `plone.protect`.

**Steps to reproduce**:

- Use `plone.protect` >= 3.0 (and `plone4.csrffixes` if on Plone 4.3)
- Create a fresh Plone Site and install the `plone.restapi:default` profile
- Enable "Store Tokens" in the `acl_users/jwt_auth` Plugin
- Do a `POST` to `@login` while already authenticated (`plone.protect` doesn't (need to) protect requests from Anonymous users)

**Security Considerations**

I've chosen to blanket whitelist those requests using the `IDisableCSRFProtection` request interface. Ideally, it would be better to simply whitelist the specific DB writes that are allowed to occur (storing the tokens) using [`plone.protect.utils.safeWrite()`](https://github.com/plone/plone.protect/blob/75634b413d7fe26e55b2859c25dc16269e8398ff/plone/protect/utils.py#L115).

However, that helper function did not always exist at that location, or even at all across all `plone.protect 3.0.x` versions. So we'd have to use some conditional imports and fallbacks to make sure we don't break `plone.restapi` with some older versions of `plone.protect 3.0.x` - that's why I chose the simpler approach of just whitelisting the entire requests.

From what I can tell, that shouldn't be a security issue:
- For `@login`, we require the user to provide a JSON body with their credentials. I can't see a way how an attacker could force a request upon the user that includes those credentials.
- For `@login-renew`, we require the existing token in an `Authorization: Bearer ...` header. A browser won't send that by itself either.

@tisto

/cc @buchi (*currently on holiday, we've previously discussed this*)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,11 @@ Changelog
 1.0a2 (unreleased)
 ------------------
 
+- Disable automatic CSRF protection for @login and @login-renew endpoints:
+  If persisting tokens server-side is enabled, those requests need to be allowed
+  to cause DB writes.
+  [lgraf]
+
 - Fixed parameter 'data' to JSON format in JWT Authentication documentation
   [lccruz]
 
--- a/src/plone/restapi/services/auth/login.py
+++ b/src/plone/restapi/services/auth/login.py
@@ -1,11 +1,13 @@
 # -*- coding: utf-8 -*-
 from Acquisition import aq_inner
 from Acquisition import aq_parent
-from Products.CMFCore.utils import getToolByName
-from Products.PluggableAuthService.interfaces.plugins import (
-    IAuthenticationPlugin)
 from plone.restapi.deserializer import json_body
 from plone.restapi.services import Service
+from Products.CMFCore.utils import getToolByName
+from Products.PluggableAuthService.interfaces.plugins import IAuthenticationPlugin  # noqa
+from zope.interface import alsoProvides
+
+import plone.protect.interfaces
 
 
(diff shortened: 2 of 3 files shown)


tisto:
@vangheem thoughts? opinions? (sorry don't have time to look into that right now :))

vangheem:
Looks fine to me.

tisto:
@lukasgraf pls rebase.

lukasgraf:
@tisto rebased, tests running

tisto:
@lukasgraf me too :)

tisto:
@lukasgraf Shall we do another alpha release after that merge?

lukasgraf:
@tisto I see - go run, I'm not in a hurry with this one ;-)

I'm very open to the idea of another alpha soon (also: we could bump `plone.rest` to beta soon I think).

These a couple documentation fixes (very few things that are actually incorrect or broken code samples, more general pass with spelling and formatting fixes) I'd like to work on next tough, and maybe convert some example responses in the docs to `test_documentation.py` dumps, so that they don't get out of date.

But that really doesn't have to be part of that release, so if you have some current use for it, my all means go ahead 👍

PR #147, opened 2016-10-15
do not break if jwt.ExpiredSignatureError

ebrehault:
When token expires, PAS plugin should return an empty credential


tisto:
Looks good to me! @buchi might want to have another look...

PR #160, opened 2016-11-25
Add test_invalid_token_returns_400.

tisto:


tisto:
I will close this merge request because it turns out robot framework is not a good fit for this kind of tests (at least not without using a better library). I will keep the branch in case I want to get back to this though.

PR #163, opened 2016-11-26
Robotframework test setup

tisto:
I created a basic robot framework test setup that uses the Python requests library to make calls to the backend. Problem is that I don't think the robot syntax is very intuitive.

Another issue that bugs me is that our test setup is extremely slow. plone.app.testing takes 12 seconds, plone.app.contenttypes and plone.app.event take 5 sec...Though, this seems to be a general issue...

Robot tests are a bit slower than using the raw requests library with plone.app.testing though.


tisto:
@bloodbare this could be an option to decouple the tests from the underlying implementation and being able to test both plone.restapi and plone server at the same time. Opinions please. :)

FYI: I'm not saying this is my favourite option, I'm just trying to explore possible solutions to our problem.

tisto:
I will close this merge request because it turns out robot framework is not a good fit for this kind of tests (at least not without using a better library). I will keep the branch in case I want to get back to this though.

PR #167, opened 2016-12-04
Make login endpoint accessible without UseRESTAPI permission

### LINKED ISSUE #166: New permission plone.restapi.UseRESTAPI should exclude @login
The permission is now enforced internally as:

https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/services/__init__.py#L24

But the thing is that the @login endpoint should be able to access by anonymous always. The tests should have been assuming some user logged in the first place, as they are not failing...

Maybe other endpoints should bypass this permission. I would personally only use the zcml permission instead of hardcoding it but... it's my 5 cents.

buchi:
Fixes #166

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,11 @@ Changelog
 1.0a7 (unreleased)
 ------------------
 
-- Nothing changed yet.
+Bugfixes:
+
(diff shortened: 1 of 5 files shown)


sneridagh:
Looks fine for me! @tisto ?

tisto:
Thanks @buchi!

Issue #170, opened 2016-12-08
Search users endpoint


tisto:
We need:

1) API design (resource modeling)
2) Implementation (should be efficient with ldap and other PAS sources)

sneridagh:
I'm trying to implement an users search endpoint for using it in autocomplete widgets. I did a first attempt of implementation here:
https://github.com/plone/plone.restapi/pull/171

The first approach I tried is to use the acl_users.searchUsers function since it searches through all the IUserEnumerationPlugin plugins.

@tiso @buchi @lukasgraf Do you think that this is any other good approach for search?

I'm really concerned about the performance of waking up the full member info. I'm not in favor of doing a dangerous potential operation like calling getMemberById that could easily escalate in terms of load. I've been there using big LDAPs, of course this happened also because the LDAP plugin is poorly written. Any insights about this?

tiso:
OT: guys, please, don't mention me.

PR #171, opened 2016-12-08
Add new search users filtering to @users endpoint

sneridagh:


sneridagh:
Related to: https://github.com/plone/plone.restapi/issues/170

src/plone/restapi/services/users/get.py
```diff
@@ -2,11 +2,14 @@
 from plone.restapi.interfaces import ISerializeToJson
 from plone.restapi.services import Service
 from Products.CMFCore.utils import getToolByName
+from zExceptions import BadRequest
 from zope.component.hooks import getSite
 from zope.component import queryMultiAdapter
 from zope.interface import implements
 from zope.publisher.interfaces import IPublishTraverse
 
+DEFAULT_SEARCH_RESULTS_LIMIT = 25
```
tisto: @sneridagh this is fine for a first implementation. Though, we have to think about if we either make this configurable, or we add batching and stuff...

tisto:
@sneridagh pls add a changelog entry. The implementation is fine. The only thing that prevents me form merging are general API design considerations. We are currently using:

/@users?query=noa

which is fine. Though, there are many other options we could consider. I just want to make sure we go with a pattern that we can use consistently in the future for other endpoints. I will do some more research and then get back to you.

@buchi @lukasgraf @bloodbare opinions?

PR #177, opened 2017-01-02
[WIP] Add serializer for dexterity Object fields

cedricmessiant:
Object fields were not supported by the library, I added serialization for such fields.


cedricmessiant:
I will probably write the corresponding deserializer too

src/plone/restapi/serializer/dxfields.py
```diff
@@ -73,3 +94,31 @@ def __call__(self):
...
+    def get_value(self, default=None):
+        result = {}
+        fieldname = self.field.getName()
+        field_value = getattr(self.context, fieldname, None)
+        if field_value:
+            read_permissions = mergedTaggedValueDict(
+                self.field.schema, READ_PERMISSIONS_KEY)
+            for name, field in getFields(self.field.schema).items():
+
+                if not self.check_permission(read_permissions.get(name)):
```
buchi: Why are you doing the permission checks at all? In your implementation `check_permission()` is always returning true. Thus I would omit all the permission check code (L102-103, L110-111, L114-115).

src/plone/restapi/serializer/dxfields.py
```diff
@@ -29,6 +36,20 @@ def get_value(self, default=None):
                        default)
 
 
+@adapter(IRichText, Schema, Interface)
+@implementer(IFieldSerializer)
+class RichTextFieldSerializer(DefaultFieldSerializer):
```
buchi: Is this serializer really needed? Shouldn't the `SchemaFieldSerializer` below also work perfectly for `IRichText`?

Aside from that, somehow I don't like to have this serializer with this name in this module because it may be confusing and would lead to naming conflicts if we would need to implement a `RichTextFieldSerializer` for `IDexterityContent`.

cedricmessiant: I had to add this serializer because IRichText implements IObject. I haven't find another way to manage this issue

src/plone/restapi/serializer/dxfields.py
cedricmessiant: of course !

tisto:
@cedricmessiant what's the status of this pr? Is this still WIP as the title suggests? Sorry for not handling this in a more timely manner...

cedricmessiant:
@tisto I think this PR is ok to merge. Unfortunately I wont be able to work on the deserializer for now

tisto:
@buchi what do we do with this? Would you suggest to close/merge/fix? 

@vincero you also run into this issue, correct? Did you check if this pr/branch solved your problem?

tisto:
I guess an option would be to just merge the ObjectFieldSerializer, this is the only one that has a test anyways. The other two serializers would need more discussion/work.

buchi:
@tisto I'll try to look at this in the next time.

tisto:
@buchi awesome! thank you! FYI: I am thinking about doing a beta release soon and then going for a final 1.0 release next week. If you have any blockers or anything that should go in in your opinion, let me know.

jensens:

PR #185, opened 2017-01-17
Crud Documentation: Update examples

jone:
1) Use `json=` for requests library.
   -> `data=` sends multipart/form-data encoded data
   -> `params=` sends params as URL params
   ->  `json=` encodes the dict as json and sends that as payload
That means for non-GET-requests we usually want `json=`, for GET-requests we want `params=` (e.g. for search filters).

2) Sync `POST` and `PATCH` examples:
Make sure that the `http-request`, `curl`, `httpie` and `python-requests`  examples send the same data in the examples.
I have not changed `PUT`, although it is not in sync, because it is not yet implemented and I do not know which example is the correct one.

---

Is a changelog required for such changes? It does not change any behavior.

### DIFF
--- a/docs/source/crud.rst
+++ b/docs/source/crud.rst
@@ -45,7 +45,7 @@ If we want to create a new document within an existing folder, we send a POST re
 
   .. code-block:: python-requests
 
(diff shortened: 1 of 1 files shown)


tisto:
Thanks @jone!

PR #192, opened 2017-02-13
Sharing / Local Roles

tisto:
Let's try a documentation first approach.


docs/source/sharing.rst
```diff
@@ -0,0 +1,93 @@
+Sharing
```
polyester: Yes, the discussion came up because "sharing" means 'send to social media, or email, or whatever' these days, and therefore is confusing. "Access" would be a better term, but I am not quite sure what is the right way to implement that. Does it need a PLIP?
Probably since it's a change that needs to be reflected also in documentation, and that at least would make it visible in changelogs, release notes and the like. 

But I've discussed it only from a user/documentation standpoint, and actually haven't looked at Plone codebase if the change would lead to confusion there and therefore should only be done towards 6, or if it's a label-only thing that can be changed in a minor release.

tisto: @polyester if there would be a PLIP and the framework team would signal their consent, I guess we could decide to rename the "sharing" endpoint to "access" in plone.api. Though, right now this seems to be a bit premature, what do you think?

polyester: I agree it would be premature; it makes more sense to have plone.api terminology in line with plone proper. Should the change happen later, there's probably ways to have one endpoint be an alias to any new terminology that may come in.

tisto:
@csenger Tests for Plone 5 fail due to inconsistencies between the Plone 4 testing layer and plone.app.contenttypes (I guess). We can look into this during the sprint...

tisto:
@bloodbare would you mind having a look at our implementation and let us know if this is in sync of what plone server does?

tisto:
@lukasgraf sorry to ask again but would you mind rebasing again? I have no idea how to fix this myself...

tisto:
@lukasgraf seems like I'm the only one with the bad karma (I tried the same on OS X). @csenger fixed it...

docs/source/sharing.rst
```diff
@@ -0,0 +1,106 @@
...
+=======
+
+Plone comes with a sophisticated user management system that allows to assign users and groups with global roles and permissions. Sometimes this in not enough though and you might want to give users the permission to access or edit a specific part of your website or a specific content object. This is where local roles (located in the Plone sharing tab) come in handy.
```
sneridagh: I would start wording things like if the reader doesn't know anything or had any previous contact with Plone, the CMS. ;) BTW, it's ok for me now, only wanted to mention it.

docs/source/sharing.rst
```diff
@@ -0,0 +1,106 @@
...
+
+
+Retrieve Local Roles
+--------------------
+
+In plone.restapi, the representation of any content object will include a hypermedia link to the local role / sharing information in the 'sharing' attribute:
+
+.. code:: json
```
sneridagh: The docs haven't been updated to the new sphinx plugin, isn't it?

tisto: No, not yet.

lukasgraf: @sneridagh @tisto that's ok, I'll do that in a second pass where I convert all remaining literal code examples to ones generated by `test_documentation.py` and displayed using `sphinxcontrib-httpexample`.

PR #195, opened 2017-02-15
[WIP] Add reset-password action to the user endpoint.

tisto:
We currently protect the entire users endpoint with cmf.ManagePortal. If we remove that to allow a user to reset her password, we have to make sure we do not open a possible attack vector.


src/plone/restapi/services/users/add.py
```diff
@@ -86,3 +112,84 @@ def reply(self):
...
+                return self._error(403, 'Unknown Token',
+                                   'The reset_token is unknown/not valid.')
+            except ExpiredRequestError:
+                return self._error(403, 'Expired Token',
+                                   'The reset_token is expired.')
+            return
+
+        if old_password:
+            # Check permissions
+            # FIXME: Check manager permission
```
tisto: @csenger what's the status here? can we check permissions properly? Otherwise we can not merge this pr into master anytime soon.

PR #204, opened 2017-02-21
Include mode attribute in json schema / types endpoint.

tisto:
This is just a proof-of-concept right now. Here is the problem: The "mode" is part of the schema, not of the field. Therefore we can not use the serialization adapters (at least not without passing the schema to each adapter). This somehow breaks the adapter pattern, because integrators will have to make a difference between the schema adapters and the form "properties". Before I continue to refactor my solution to improve it I would like to hear some feedback. @davisagli @lukasgraf @buchi?

### DIFF
--- a/src/plone/restapi/types/utils.py
+++ b/src/plone/restapi/types/utils.py
@@ -103,6 +103,8 @@ def get_fields_from_schema(schema, context, request, prefix='',
             fields_info[fieldname] = adapter.get_schema()
         if fieldname in hidden_fields.keys():
             fields_info[fieldname]['mode'] = hidden_fields[fieldname]
+        else:
+            fields_info[fieldname]['mode'] = u'input'
 
     return fields_info
 
--- a/src/plone/restapi/types/utils.py
+++ b/src/plone/restapi/types/utils.py
@@ -9,6 +9,7 @@ from zope.i18n import translate
 from zope.schema import getFieldsInOrder
 from plone.autoform.interfaces import IFormFieldProvider
 from plone.autoform.interfaces import MODES_KEY
+from plone.autoform.utils import mergedTaggedValuesForForm
 from plone.behavior.interfaces import IBehavior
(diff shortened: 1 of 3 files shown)


src/plone/restapi/types/utils.py
```diff
@@ -76,13 +76,20 @@ def get_ordered_fields(fti):
     return (fields, ordered_fieldsets_fields)
 
 
+def get_hidden_fields_from_schema(schema):
```
lukasgraf: @tisto does this respect schema interface inheritance? (Think schema inheritance chain where a field is omitted at some point, but later down the chain gets un-omitted by yet another directive).

It might be appropriate to use [`mergedTaggedValuesForIRO()` or `mergedTaggedValuesForForm()` from `plone.autoform.utils`](https://github.com/plone/plone.autoform/blob/master/plone/autoform/utils.py#L40-L67) here - these should respect IRO (Interface Resolution Order) and merge tagged values accordingly.

src/plone/restapi/types/utils.py
```diff
@@ -76,13 +76,20 @@ def get_ordered_fields(fti):
...
+
+
 def get_fields_from_schema(schema, context, request, prefix='',
                            excluded_fields=None):
     """Get jsonschema from zope schema."""
     fields_info = OrderedDict()
     if excluded_fields is None:
         excluded_fields = []
-
+    hidden_fields = get_hidden_fields_from_schema(schema)
```
tisto: Here I look up the hidden fields from schema.getTaggedValue(). We might want to refactor this to be more generic and catch all possible plone.autoform MODE_KEYS.

src/plone/restapi/types/utils.py
tisto: @lukasgraf you are reviewing faster than I can type. :) Good point. I will look into that if we figured out the more general issues (adapter pattern vs tagged values).

tisto:
Next try: https://github.com/plone/plone.restapi/pull/216

Issue #187, opened 2017-01-26
Null value representation. WAS: How to clear a datetime field with json?


pabo3000:
It is Plone site with plone.restapi installed. There is a dx content type with a zope.schema.Datetime field. It is declared in this way:

```
<model xmlns="http://namespaces.plone.org/supermodel/schema">
  <schema>
    <field name="due" type="zope.schema.Datetime">
      <description/>
      <title>due</title>
      <required>False</required>
    </field>
  </schema>
</model>
```
Let's assume the value of the field is any Datetime. But now the value should be unset by a json rest call (Update, Patch. After that call the field should be unset. How does the json call look like?

I have tested some possibilties which were not successful:

```
PATCH {}
PATCH {"due":""}
PATCH {"due":"None"}
PATCH {"due":null}
PUT {}
```
I think there should be a possibility to reset a Datetime field via rest. Therefore I append this issue.

tisto:
@pabo3000 we indeed need a way to represent null values in the API. I will look into it...

tisto:
I did some initial research and I think using PATCH {'my-field': null} is the way to go. Since a PATCH by definition only contains a partial resource, not sending a field is not an option here.

@robgietema @ebrehault what do you think? How does plone-react and the angular sdk currently handle this?

I plan to work on this during the sprint next week...

davisagli:
`PATCH {"my-field": null}` makes sense to me as the way to set a field's value to `None` in Python.

lukasgraf:
In the case of Dexterity, to **unset** a field the stored value being written should actually be `field.missing_value` I believe, which is not necessarily Python `None`.

See [`missing_value` in `zope.schema`](https://github.com/zopefoundation/zope.schema/blob/919388ca4e26346f1a4463e0fecc8ad2fd9f871b/src/zope/schema/_bootstrapfields.py#L88-L95) and  [p.a.dexterity docs on field properties](https://docs.plone.org/external/plone.app.dexterity/docs/reference/fields.html#field-properties)

Unfortunately, that is semantically different from just setting a field's value to `None`, which one might also want to do.

Either way, I think it's imporant that we're really clear about the exact semantics of the operation in question: Is it supposed to **unset** a field, or **set it to a null value**?

csenger:
I added support for null/None in #513 . It will set the value to field.__missing_value__. It's dx only.

[...]

PR #205, opened 2017-02-21
Make types endpoint return 'string' for the widget attribute by default.

tisto:


tisto:
@ebrehault I would like to hear your opinion before I put effort into adapting our tests to pass. Would you be ok if we would always include the widget attribute and make "string" the default value for it?

ebrehault:
@tisto why do you think it is better to do provide a default widget in the schema rather than let the mechanism in charge of rendering the form to choose by itself a default widget if none is provided in the schema? (I don't say I am opposed to your change, I am just curious)

tisto:
@ebrehault it was actually a request made by my front end devs. I think whatever we choose to do, we should try to be consistent. So here are the options (in general):

1) Always show all possible fields and force a default value
2) Always show all possible fields and allow None as a value
3) Omit fields that don't have a value set

We currently do 3) here but not for other endpoints (e.g. content). I need to investigate this a bit further, but I think we currently do 2). If this is the case we should do the same here as well (and not use the first option that I proposed initially).

ebrehault:
+1 for consistency, and my preference would either 1 or 3 (I think 2 is quite useless)

tisto:
@ebrehault could you elaborate why you think 2) is useless? I would say this is what we currently do for the content endpoint. We iterate over existing the attributes (zope schema, behaviors, etc.) and then show None if there is no value set (and no default value defined, which can be the case). This matter is also related to the question how we represent null values in plone.restapi: https://github.com/plone/plone.restapi/issues/187

ebrehault:
@tisto if we have:
```json
{"id": 12, "title": "my title", "subjects": null}
```
or:
```json
{"id": 12, "title": "my title"}
```
This code will work the same with both:
```javascript
.subscribe(response => {
    if(!response.subjects) { ... }
});
```
but the second is shorter and easier to read.

tisto:
@ebrehault convinced me. Will close this pr.

PR #206, opened 2017-02-21
Improve types endpoint tests.

tisto:
Expected value comes first in assertions, be as explicit as possible, one empty line between precondition and assertions.


ebrehault:
cannot merge because changelog is not ok

tisto:
@ebrehault merci! :) 

Will merge it myself now. I don't think we should add a changelog note for a minor test refactoring.

PR #210, opened 2017-02-22
@groups endpoints implementation

sneridagh:
This is the implementation of the @groups endpoints allowing to make CRUD operations. Very much the same as the @users endpoints.


tisto:
@sneridagh LGTM!

PR #220, opened 2017-03-01
@principals endpoint

sneridagh:


docs/source/_json/principals.resp
```diff
@@ -0,0 +1,16 @@
...
+    {
+      "@id": "http://localhost:55001/plone/@groups/ploneteam", 
+      "description": "We are Plone", 
+      "email": "ploneteam@plone.org", 
+      "groupname": "ploneteam", 
+      "id": "ploneteam", 
+      "title": "Plone Team"
+    }
+  ], 
+  "users": []
```
tisto: @sneridagh shouldn't there be at least the test users? I mean this is auto-generated from the default p.a.testing fixture, right?

src/plone/restapi/tests/test_documentation.py
```diff
@@ -608,3 +608,23 @@ def test_documentation_navigation(self):
         response = self.api_session.get(
             '{}/@components/navigation'.format(self.document.absolute_url()))
         save_request_and_response_for_docs('navigation', response)
+
+    def test_documentation_principals(self):
```
tisto: @sneridagh ah, now I get it. Maybe we should use a query (or no query at all) here that returns at least one user.

docs/source/_json/principals.resp
sneridagh: @tisto It uses a query that the test user doesn't match. I can create a test for it to match also for users (or both, users/groups).

src/plone/restapi/tests/test_documentation.py
sneridagh: @tisto yeah, sure, I will add it. (in fact in the default tests it's already there).

sneridagh: @tisto I started to document this as well, but documentation wise, I think it will be pretty much redundant. So for now I will leave it as it is.

tisto:
@lukasgraf not sure if this is just bad karma or my stupidity...but I have the exact same problem with this pr: https://travis-ci.org/plone/plone.restapi/jobs/208645688

lukasgraf:
@tisto I think something with the merge strategy your (version of) `git` is using might be off - the merge commit only shows conflicts in `test_documentation.py`, but none of the other files. So git was thinking it was able to "solve" those other conflicts, even though it apparently couldn't.

But it could also be a silly user error, not ruling that out at this point 😜 

Pushed another fix-commit, sorry for the fallout from my docs-update!

tisto:
@lukasgraf thank you again!

PR #226, opened 2017-03-03
Make @types endpoint include an 'omitted' attribute.

lukasgraf:
This will serialize Tagged Values from the `omitted` and `no_omit` `plone.autoform` directives in the `@types` endpoint.

---

One case that isn't handled yet is the ability to omit fields only for a particular type of form:

```
    form.omitted('edit_only')
    form.no_omit(IEditForm, 'edit_only')
    edit_only = schema.TextLine(
        title = u'Only included on edit forms',
        )
```

This is not specific to omitted fields though, this applies to other directives as well, `mode()` for example also allows for this.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,11 @@ Changelog
 1.0a10 (unreleased)
 -------------------
 
+New Features:
+
+- Make @types endpoint include an 'omitted' attribute.
+  [lgraf]
+
 Bugfixes:
 
 - Add missing id to the Plone site serialization, related to issue #186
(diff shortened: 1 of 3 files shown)


buchi:
Wouldn't it be better to just omit fields if they are tagged with the omitted directive? Same question applies to permissions (#227). This way we're exposing more information.

@lukasgraf PR needs rebase

lukasgraf:
@buchi good point.

My initial though was that the `omitted` flag is a form-level hint, and doesn't necessarily apply to the schema as a whole. Fields can still be present in the schema, but the hint suggests they should be omitted from a particular form. This way, the client would get the full information, and it's left up to the what exactly to do with that.

But yes, we do indeed expose more information that way then a regular Plone would.

I don't feel to strongly about this either way - @ebrehault @tisto thoughts?

lukasgraf:
Superseded by #237

PR #227, opened 2017-03-03
Make @types endpoint include field permissions (read/write).

lukasgraf:
⚠️ (Based on #226, merge that one first) ⚠️ 

---

Includes field level permissions in `read_permission` and `write_permission` for the `@types` endpoint.


lukasgraf:
Superseded by #237

PR #231, opened 2017-03-03
Add review_state on summary serializer

sneridagh:
Since this is already provided by IContentListingObject, we can easily add it to the summary serializer at no cost. By doing this, it will be easier to build basic listings from the summary of a folder.

Maybe there's still some other IContentListingObject information that we could include...
https://github.com/plone/plone.app.contentlisting/blob/master/plone/app/contentlisting/interfaces.py#L10

Bottom line is that I had to amend 17 tests failures! ;)

What do you think? /cc @lukasgraf @buchi @tisto @bloodbare

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,11 @@ Bugfixes:
 - Add missing id to the Plone site serialization, related to issue #186
   [sneridagh]
 
+New Features:
+
+- Add review_state to the folderish summary serializer.
+  [sneridagh]
+
 1.0a9 (2017-03-03)
 ------------------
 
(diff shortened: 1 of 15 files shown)


lukasgraf:
`+1` from me for `review_state` in summary representations 👍

erral:
Could this be handled with some kind of framing/expansion?

If we start providing additional values apart from `@id` `title` `description` and `@type` anyone could argue "why the attribute XXX is not included"? 

I don't know if I get the framing/expansion right, but if it's used to provide aditional information about the item, this could be use for this too, perhaps something like: `?expand_summary=review_state,added,Subject`

erral:
Already discussed with @lukasgraf and @sneridagh:

- The idea of Framing/Expansion is to provide additional information about existing endpoints and not about providing extra dynamic data. Something like: "we also have workflow information -> Include workflow information".

- If you want to get additional information, just use the `@search` endpoint

PR #232, opened 2017-03-03
Added API conventions.

tisto:


docs/source/index.rst
```diff
@@ -31,6 +31,7 @@ Contents
    serialization
    searching
    customization
+   conventions
 
 .. include:: ../../README.rst
```
svx: Including the README into the docs is **not** best practice ! :)

There is a difference between a README and documentation !
Also with you current README you will include for example all the widgets ov travis and so oninto you docs, this is not really cool or a benefit.

svx: Oh and please check your docs, currently this file here is cutting sentences ! This is also not coll and really hard for people who use for example translation tools.

Rule of thumb:

One sentence per line, use line breaks which makes sense, line should not longer than around 130.

Please do not use pep8 or so on docs !

:)

docs/source/conventions.rst
```diff
@@ -0,0 +1,113 @@
+Conventions
+===========
+
+Nouns vs Verbs
+--------------
+
```
svx: Please check your reST, you are mixing here and on other place headers, this leads currently to failing tests for me, because your reST is not valid !

:)

PR #233, opened 2017-03-04
Implement @copy and @move endpoints

### LINKED ISSUE #180: Implement copy/paste or move.
We need to think about how we would want to implement copy/paste or moving items via rest api. One could argue that can already be handled by the frontend (GET => copy, POST => paste, DELETE => move). Though, especially for larger folder structures or larger files, we might want to think about keeping all this on the backend.

Opinions?

buchi:
Fixes #180


tisto:
@buchi wohoo! That was quick. Is this ready for review or still WIP as the label says?

lukasgraf:
@buchi for solving the docs rebase conflicts you'll need to:
- Use the new `save_request_and_response_for_docs` function in `test_documentation.py` (has been renamed, previously `save_response_for_documentation`)
- only pass it the name, without a `.json` extension (so just `copy` and `move`)
- include e.g. the produced `move.req` and `move.resp` separately in the `.rst` file, something like this:

```
Request examples:

..  http:example:: curl httpie python-requests
    :request: _json/move.req

Response:

.. literalinclude:: _json/move.resp
   :language: http
```

src/plone/restapi/services/copymove/copymove.py
```diff
@@ -0,0 +1,96 @@
...
+                else:
+                    parents_ids[parent] = [obj.getId()]
+
+        results = []
+        for parent, ids in parents_ids.items():
+            result = self.context.manage_pasteObjects(
+                cb_copy_data=self.clipboard(parent, ids))
+            for res in result:
+                results.append({
+                    'old': '{}/{}'.format(
```
tisto: @buchi I was wondering if it would be better to use "source"/"target" here instead of "old"/"new" to be consistent with request that initializes the copy/move (which uses "source").

tisto:
@buchi @lukasgraf @sneridagh here is a follow up on the discussion that we started on the Beethoven Sprint:

We could implement the API in a way that we would support async operations. The server would respond with the HTTP response "202 Accepted" instead of 200. The client then can check the progress header to wait until the operation has been succeeded (see for http://restful-api-design.readthedocs.io/en/latest/methods.html#asynchronous-requests for a full example).

Though, the question is if we would want to make this the default behavior or not. I could imagine that we might want to make this configurable. We certainly do not want to hard-code any async backend implementation and even the async API I briefly described might be overkill to implement for a lot of use cases. If you try to move a large portion of content with vanilla Plone the request will also run into a timeout. In any case, an async API and even more the implementation needs some serious design thoughts. I'm sure we will do that at some point. Though, right now this would just hold us back. Therefore I'd suggest to merge the current design and implementation. We can offer an alternative async implementation at a later point without breaking the API.

lukasgraf:
@tisto I agree: Async implementation would be very much desirable in the future, but this shouldn't hold us back going forward with the current implementation.

And yes, while async might be really desirable for *these* specific operations, it's not nearly as relevant for other endpoints that already exist. So I'd lean towards creating entirely new endpoints for the async variants, like `@move-async`, *when* we need them. They'll have to be consumed in a totally different way anyway.

tisto:
@lukasgraf sure, a second endpoint would be an option. Let's discuss this when someone actually has the need for this and is willing to work on an implementation (as said, that could be very likely us).

tisto:
@lukasgraf the Plone 4.3 travis job currently fails with a docs diff failure. I already ran the test locally and there is no difference. Would you mind to have a look? https://travis-ci.org/plone/plone.restapi/jobs/208617965

lukasgraf:
@tisto the difference is in the casing of the request header names (old: `content-type`, new: `Content-Type`), and the location of the `Authorization` header.

So somehow your merge commit (b9b53b10) introduced those old lines. Running `bin/test -t test_documentation` locally for the `copy-move` branch results in exactly the diff for me that the test failure complains about.

lukasgraf:
@tisto if you'd like, I can push the necessary changes to this branch to make tests pass.

tisto:
@lukasgraf yes please. I just did a git pull and merge into copy-move, run the tests and do not see any changes....

lukasgraf:
@tisto here we go, tests should pass soon. I'd recommend merging without squashing to retain proper attribution (not 100% sure how GitHub handles that with multiple authors).

tisto:
@buchi to not further mess with merge conflicts I will merge this now. Please comment on the new/old vs. source/target issue.

src/plone/restapi/services/copymove/copymove.py
buchi: @tisto yes, that makes sense.

tisto: @buchi ok, we should change that then before making a new release...

tisto: @buchi FYI: I changed it here: https://github.com/plone/plone.restapi/commit/05dd972e981a58051ce25578c8b4920c09c693ef

PR #237, opened 2017-03-04
Overhaul JSON schema generation for @types endpoint

lukasgraf:
This new implementation instantiates a minimal `z3c.form` in order to allow `plone.autoform` to do all the heavy lifting (particularly processing fields, fieldset directives and field move directives).

This makes sure we correctly process and respect (or at least have the option to respect):

- Schema interface inheritance (IRO)
- Additional schemata from behaviors
- `plone.autoform` directives / form hints:
  - Fieldsets
  - Field moves
  - Omitted fields
  - Field modes
  - Field level permissions

Basically all the heavy lifting is done in (or triggered from) [`plone.autoform.base.AutoFields. updateFieldsFromSchemata()`](https://github.com/plone/plone.autoform/blob/1b24ddfa8cf6e68914772d5d1c5fe89653c7a7c3/plone/autoform/base.py#L34-L135)

So instead of attempting to do these complex tasks ourselves in `plone.restapi` like in the [previous implementation](https://github.com/plone/plone.restapi/blob/075ef20823418bbdeead69c168f49c8d30460b73/src/plone/restapi/types/utils.py#L33-L79), I instantiate [a minimal `z3c.form`](https://github.com/plone/plone.restapi/blob/tgv-fieldsets/src/plone/restapi/types/utils.py#L15) inheriting from `AutoExtensibleForm` (and therefore indirectly from `AutoFields`), and make sure  `form.updateFieldsFromSchemata()` gets called on it.

TODO:
- [ ] Extend tests with some schema interface inheritance testcases
- [ ] Extend docs for `@types` endpoint 

---

This also brought to the surface another issue / aspect: If we're talking about schemas with the purpose of rendering a specific form, we'll need to tell the implementation that produces the JSON schema which form this is gonna be. Different fields can be omitted / hidden / ... for edit forms than for add forms. For now I went with defaulting to delivering a schema appropriate for an add form, but I plan to further extend this implementation to support different form modes as well.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,10 @@ Changelog
 
 Bugfixes:
 
+- Overhaul JSON schema generation for @types endpoint. It now returns
+  fields in correct order and in their appropriate fieldsets.
+  [lgraf]
+
 - Add missing id to the Plone site serialization, related to issue #186
   [sneridagh]
 
--- a/docs/source/_json/types_document.json
+++ b/docs/source/_json/types_document.json
@@ -10,13 +10,6 @@ content-type: application/json+schema
       "fields": [
         "title", 
         "description", 
-        "subjects", 
-        "language", 
-        "effective", 
-        "expires", 
-        "creators", 
-        "contributors", 
-        "rights", 
         "text", 
         "changeNote"
       ], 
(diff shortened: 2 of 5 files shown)


lukasgraf:
@davisagli I'd love your 2 cents on whether or not this approach is sensible.

Instantiating a [minimal `z3c.form`](https://github.com/plone/plone.restapi/blob/tgv-fieldsets/src/plone/restapi/types/utils.py#L15) with the appropriate `schema` and `additionalSchemata`, and then calling `form.updateFieldsFromSchemata()` on it seems to work. In particular, theres two things I'm unsure about:
- Is there any state that `form.updateFieldsFromSchemata()` relies on, in the sense that some other form initialization should happen before it's called?
- Could there be any side effects to `request` (or possibly `context`) from invoking this method?

davisagli:
@lukasgraf I think that approach will work, and it seems like a good way to make sure that fields are processed the same as for a server-rendered form.

lukasgraf:
@davisagli thanks for the quick feedback!

Issue #253, opened 2017-03-08
p.a.event: Cannot compare offset-naive and offset-aware datetimes


sneridagh:
On the deserializer: 
```
  Module plone.restapi.services.content.update, line 22, in reply
  Module plone.restapi.deserializer.dxcontent, line 78, in __call__
TypeError: can't compare offset-naive and offset-aware datetimes
```

In fact, you cannot feed a PATCH request with the GET response of the same object... it's broken with the same error.

sneridagh:
Here we are flattening the TZ information of any Datetime:
https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/deserializer/dxfields.py#L46

@buchi @lukasgraf @tisto There is any reason for this? This is what is causing that the later comparision when doing a PATCH fails for p.a.event (since it forces it to have TZ):

What would be the better approach in this case?

tisto:
@sneridagh I think we should create a test class that makes sure all content objects we GET from the API can be used in a POST to create an object.

@thet opinions on the datetime representation on plone.restapi?

thet:
@tisto @sneridagh tnx for asking :)

I suggest representing dates as ISO8601 string converted to UTC and eventually transmitting the display timezone as a string in a separate field.

Example: `2017-03-08T15:18:52+00:00`

But since plone.app.event 2.0, event type start and end dates are stored with their original timezone information and the timezone is a portal wide setting, which can be overwritten by user settings. So the target timezone can also be set as a site wide setting by the API client. However, the API client needs to know about a "default timezone" to display times.

Regarding timezones : The ISO representation only knows timezone offsets from UTC, which is not very helpful. E.g. you cannot know, which standard and daylight saving time rules apply for the offset. For that, you need to know the timezone name, like "Europe/Vienna".
But when we transmit dates converted to UTC and let the client/server decide into which zone to convert, this isn't actually a big deal.


## JavaScript

```
// Parsing
var dt = new Date('2017-03-08T15:18:52+00:00')

// Serializing
dt = dt.toISOString()  // '2017-03-08T15:18:52.000Z'

// better bring in same format like python datetime isoformat and W3C examples
dt = dt.substring(0, dt.length-5) + '+00:00'  // '2017-03-08T15:18:52+00:00'
```


## Python

```
from dateutil.parser import parse
dt = parse('2017-03-08T15:18:52.000Z')
[...]

PR #258, opened 2017-03-08
Propose a fix for the issue #253 by not comparing IDatetime fields wh…

sneridagh:
…en patching and set the value always.


sneridagh:
@tisto @buchi @lukasgraf It's the simplest and less intrusive solution that I could think of to solve this problem. Let me know what do you think.

lukasgraf:
I haven't been involved with this part of the original implementation, so I'm afraid I can't really give any meaningful feedback without fully diving into it.

buchi:
This will only work around the error you currently get when trying to create an event through the REST API. However you get an event with start/end dates without tzinfo which is inconsistent to events created through the web. Most likely this will lead to errors when viewing/editing the event with Plone's default views/forms. See discussion in #253

sneridagh:
Closing in favor of: #394

PR #260, opened 2017-03-11
Add verbose message to avoid confusion for creating document under a folder docs

prakharjoshi:
Here are few confusion that I faced while trying out APIs.

1) When I see the url `/folder	Creates a new document within the folder` this looks like a url something like `http://localhost/Plone/folder` but here folder means the name of the folder in which we want to create a document. 

Also if we show APIs like `/folder` it signifies that this is the endpoint but the reality is it is something like this `/{{folder_name`.

So the quick fix that pop up to me is to add a note before declaring these endpoints that it means the folder in which you want to create a document. Later may be we can update the endpoints and their responses in generic way, something like this `/{{folder_name}}`

### DIFF
--- a/docs/source/content.rst
+++ b/docs/source/content.rst
@@ -5,6 +5,9 @@ plone.restapi does not only expose content objects via a RESTful API. The API co
 
 Manipulating resources across the network by using HTTP as an application protocol is one of core principles of the REST architectural pattern. This allows us to interact with a specific resource in a standardized way:
 
+.. note::
+        Here folder implies to name of the `folder name` in which you want to create the new document.
+
 ======= ======================= ==============================================
 Verb    URL                     Action
 ======= ======================= ==============================================
--- a/docs/source/content.rst
+++ b/docs/source/content.rst
@@ -5,17 +5,14 @@ plone.restapi does not only expose content objects via a RESTful API. The API co
 
 Manipulating resources across the network by using HTTP as an application protocol is one of core principles of the REST architectural pattern. This allows us to interact with a specific resource in a standardized way:
 
-.. note::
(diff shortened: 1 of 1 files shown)


tisto:
@prakharjoshi thanks for your fix! I agree this is confusing. Though, I would rather rewrite the paragraph and the table instead of adding a note.

docs/source/content.rst
```diff
@@ -28,7 +25,7 @@ If we want to create a new document within an existing folder, we send a POST re
 
   .. code-block:: http-request
 
-    POST /folder HTTP/1.1
+    POST /{{folder_name}} HTTP/1.1
```
tisto: @prakharjoshi I'm afraid this won't work since we moved to auto-generate the examples. We haven't been able to replace this everywhere yet though.

docs/source/content.rst
```diff
@@ -5,14 +5,14 @@ plone.restapi does not only expose content objects via a RESTful API. The API co
...
-POST    /folder                 Creates a new document within the folder
-GET     /folder/{document-id}   Request the current state of the document
-PATCH   /folder/{document-id}   Update the document details
-DELETE  /folder/{document-id}   Remove the document
-======= ======================= ==============================================
+======= =============================    ==============================================
+Verb    URL                              Action
+======= =============================    ==============================================
+POST    /{{folder_name}}                 Creates a new document within the folder
+GET     /{{folder_name}}/{document-id}   Request the current state of the document
```
tisto: @prakharjoshi we should either use single brackets or double brackets. Though, we should not mix both.

tisto:
Will close this pr. This section still needs an overhaul though.

PR #261, opened 2017-03-12
Tus upload

### LINKED ISSUE #79: TUS Endpoint
http://tus.io/

buchi:
Closes #79


tisto:
@buchi anything we can do to help here to get rid of the WIP? I might be able to allocate a few working hours early next week if you give us a hint where we could start.

csenger:
@buchi I'd add tests and docs for TUS to make the code ready to merge. Is there missing functionality before TUS can be merged, other changes you would like to be done or problems that block it from being merged?

csenger:
There are several more things to fix in this branch:

### errors/status codes

I corrected the status codes for some errors. I need to go through all the possible errors and add tests/correct status codes to match the spec.

### uploading text content does not work

The uploader queries the content type registry with the filename + mime type. For text files this is `Document` and the uploader can't handle that. I don't think we want to create Documents at all, but on the other hand we should not hard code `File`/`Image` as types to create.

I added a failing test for that.

### location
The spec requires the POST to respond with a location header to return the location of the generated resource:

> The Server MUST acknowledge a successful upload creation with the 201 Created status. The Server MUST set the Location header to the URL of the created resource. This URL MAY be absolute or relative.

The uploader is implemented in a way to store the unfinished data on disk in a temporary directory. We return a link to the upload endpoint, e.g.`http://localhost:55001/plone/testfolder/@upload/ab7aeaf37d2c40ffb5818aec52ab8150`. But this is only a temporary url to look up the disc cache. We create the content object when the upload is complete. This is a violation of the spec.

To fulfill the spec we need to create the file content during the initial POST request and return the final url of the file/image as the location. The consequence would be that 

* we have several transactions to the ZODB. At least 2 (the initial POST and the PATCH to upload the file data). Potentially more
* we might end up with empty or half uploaded files. There is a termination extension for TUS to cleanup uploads.

I see no way to circumvent that with a standard Plone site.

buchi:
@csenger @tisto I'll try to look at it this week and give some feedback. The implementation is not finished yet, mainly handling of edge cases and cleanup stuff is missing. So don't merge it yet! We're also missing the ability to replace files (not the Plone file object but the value of the file field in an existing file object). The TUS specification provides no information how to handle that.

@csenger regarding your remark about the location I do not agree. It's a question of interpretation though, but IMO we're fulfilling the specs. The created resource by POST is a temporary upload object and we're returning it's location. I don't see any indications that this has to be the final resource. My second point is that we have to make an implementation that fits with Plone. Creating unfinished content objects would not work well with Plone. E.g. Plone would try to index the object and the object would be visible to other users...

tisto:
@buchi did you had time to look into this yet? We really need the TUS upload in our project and I would like to get away without a custom solution on our side. :)

buchi:
@tisto I've now implemented the TUS expiration extension which is required to cleanup files of unfinished uploads lying around. With that we have a basic implementation that could be merged regarding features. However test coverage and documentation still need some work. I won't have much time to continue here in the next days. We also need support for overwriting existing files, but that can be implemented in a separate PR.

@csenger I've fixed upload of text documents by handling `IRichtText` fields. This means that in a default Plone site if somebody uploads a text file, it will create a Document and not a file. To create a file one can either provide the `@type` property in the `Upload-Metadata` header or change the configuration in the content type registry.

csenger:
@buchi I see the same problems with partial writes in Plone and it's far from ideal. 

The spec has no explicit mechanism for temporary upload URLs. For the upload itself it is no problem. Though we can not cover common use cases like
* showing a link to the uploaded file
* display the file/image after the upload finished
* link to the file we created to the content it belongs to

So say your app is an issue tracker. You create a ticket and upload attachments while creating a ticket. The app won't be able to link the attachments to the ticket as it can't get the final URL of the resource.

As a pragmatic solution we could return the new location with the final 204. Reading through the tus.io JS client an app using that library can't get the location header. And as the spec does not cover the change of the resource location when the upload is finished we can only cover these use cases within the spec if we return the final location on creation.

So with all the problems it may have on the Plone side I think we need to handle those.
The best approach I can think of is storing the upload information as non content objects in the target folder and replace them with content if the upload finishes. This way we won't interfere with content space and reserve the id.

src/plone/restapi/services/content/configure.zcml
```diff
@@ -70,6 +70,45 @@
     permission="zope2.View"
     />
 
+  <plone:service
+    method="OPTIONS"
+    name="@upload"
```
tisto: @buchi didn't we decide to use a separate "tus-upload" endpoint in addition to a regular "upload" endpoint in Bonn?

src/plone/restapi/services/content/configure.zcml
```diff
@@ -70,6 +70,45 @@
...
+
+  <plone:service
+    method="PATCH"
+    name="@upload"
+    for="Products.CMFCore.interfaces.IFolderish"
+    factory=".upload.UploadPatch"
+    permission="cmf.AddPortalContent"
+    />
 
+<!--   <plone:CORSPolicy
```
tisto: @buchi I guess this can be removed, correct?

buchi:
@csenger I get your point but I still think that an implementation that returns the final location on the initial POST request would be much more complex and not worth the trouble it causes.

I highly favor returning the location at the final PATCH request. I've tested it with the official tus-js-client and got it working without problems. I used the provided demo and the only thing I had to change is a line in the `onSuccess` handler. I replaced `anchor.href = upload.url;` with `anchor.href = upload._xhr.getResponseHeader('Location');`.  As this applies to integration code only and doesn't touch the library itself, I see no problem doing that. Do you have any objections against this approach?

src/plone/restapi/services/content/configure.zcml
buchi: @tisto Hmm... I remember that we were talking about two endpoint names but I don't see any need for two differently named endpoints in the TUS implementation.

src/plone/restapi/services/content/configure.zcml
buchi: Yes, but I would like to have it in tests and maybe in the docs. I'll work on it a little bit during the Easter weekend.

src/plone/restapi/services/content/configure.zcml
tisto: @buchi I was under the assumption that we have an existing @upload endpoint for regular (smaller) file uploads and a separate @tus-upload endpoint for the tus-upload. I think this is how the Guillotina API is implemented. @bloodbare @vangheem can you confirm that?

tisto:
@buchi can we remove WIP/in progress for this PR? I would like to make some final reviews and then merge asap.

src/plone/restapi/services/content/configure.zcml
buchi: Doesn't make sense for me to implement an additional upload endpoint for smaller files. TUS can handle those very well.

We now have the following endpoints:
 - `OPTIONS /myfolder/@upload`: Get info about configuration and supported options
 - `POST /myfolder/@upload`: Request new upload resource for creating a new content object
 - `PATCH /myfolder/@upload/mytusuploaduid`: Upload a file or part
 - `HEAD /myfolder/@upload/mytusuploaduid`: Get current offset of the upload

What we are missing yet are endpoints to replace the file (field) of an existing content object. Something like `POST /myfolder/myfile/@upload/field` and `PATCH /myfolder/myfile/@upload/field`. But that's out of scope for this PR.

buchi:
@tisto Implementation and tests should be in a good shape. Docs are still missing. I'll try to finish it this week.

src/plone/restapi/services/content/configure.zcml
tisto: @buchi here are my notes from our discussion at Beethoven sprint:

https://github.com/plone/plone.restapi/issues/79#issuecomment-287565245

I agree that it doesn't make sense to have two upload endpoints if TUS is all we have. Though we have to make sure we keep our API in sync with Guillotina.

tisto: @bloodbare @vangheem FYI: we really need your opinion here if we want to keep the Guillotina API and plone.restapi in sync.

tisto:
> What we are missing yet are endpoints to replace the file (field) of an existing content object. Something like POST /myfolder/myfile/@upload/field and PATCH /myfolder/myfile/@upload/field. But that's out of scope for this PR.

@buchi I think we need a way to replace an uploaded file as well if we want people to use the endpoint. I could ask one of our devs to look into it this week. Any pointers or anything we need to take into account when looking into this?

tisto:
@buchi any update on this? I would really like to merge this asap. We can provide help with docs or even implementation if necessary.

src/plone/restapi/services/content/configure.zcml
jaroel: Guillatina uses `tusupload`, see https://github.com/plone/guillotina/blob/master/guillotina/api/files.py#L79

buchi:
@tisto This one is now ready for final review and merging.

I've added the replace functionality by creating an additional endpoint named `@upload-replace`. Unfortunately the tus.io specification does not define a way to replace an existing file, e.g. by using the PUT method. Thus I've implement an additional POST endpoint that works similar to the `@upload` POST endpoint but can be used to replace an existing file. This way it's possible to use existing tus.io libraries to replace existing files.

tisto:
@buchi wohoo! Awesome! Thank you!!!

tisto:
@buchi I just released plone.restapi 1.0a18. This is scheduled for 1.0a19 :)

tisto:
@buchi we discussed the "tus-upload" vs "upload" issue in our team (@sneridagh @csenger @jaroel). Our main concern is that by using just "upload" we force any client library to support TUS and we won't be able to support a "simple" upload mechanism in the future. In our current project we have a simple upload for instance, that we might want to push into plone.restapi in the future. I am not saying that that we won't get away with a tus-only upload in plone.restapi. I just would like to keep our options open, because TUS is still the new and advanced thing here. Would you be opposed to just rename "upload" to "tus-upload" and "upload-replace" to "tus-replace"?

When we sorted that out we could do a quick final review and push a "tus"-release today or in the next days. I would be also available for a quick call if that helps. :)

buchi:
@tisto Absolutely no objections. In fact I like "tus-upload" and "tus-replace" more. I've renamed the endpoints now.

tisto:
@buchi awesome! Again: thank you for your fantastic work! I just updated the TUS docs. Hope you are ok with that. If not, let me know.

@jaroel could please check the docs and add examples for the tus-replace endpoint tomorrow? Maybe you can also have a final look at the PR and let us know what you think.

Tomorrow is a holiday in parts of Germany and I will be off during the day. I can merge and do a release tomorrow evening though.

docs/source/tusupload.rst
```diff
@@ -1,88 +1,99 @@
 TUS resumable upload
 ====================
 
-Plone REST API supports the TUS Open Protocol for Resumable File Uploads.
-See http://tus.io/ for more information on TUS.
+plone.restapi supports the `TUS Open Protocol <http://tus.io>`_ for resumable file uploads.
+The two TUS endpoints to upload '@tus-upload' a file and to replace an existing file '@tus-replace' are available on all 'folderish resources' in Plone (content types that implement the IFolderish interface, e.g. a 'Folder').
```
buchi: @tisto '@tus-replace' is not limited to folderish resources, it's available on all `IContentish` resources.

docs/source/tusupload.rst
```diff
@@ -1,88 +1,99 @@
...
+.. literalinclude:: _json/tusupload_patch_finalized.req
+   :language: http
 
-.. literalinclude:: _json/tusupload_post.resp
+When the upload request contains of just one single file, the server will respond with a `204: No Content` response and the final file URL in the HTTP location header:
+
+.. literalinclude:: _json/tusupload_patch_finalized.resp
    :language: http
 
+The actual content item is creade afert the file upload has been finalized..
```
buchi: typos: "creade afert" -> "created after", second dot at the end of the sentence

tisto: fixed. thx!

docs/source/tusupload.rst
tisto: @buchi fixed in the docs. Though, I do not fully understand why one would want to use tus-replace without tus-upload on a non-folderish object. My understanding of TUS is limited though. :)

buchi: @tisto In fact it's more complicated. `POST @tus-upload` is limited to folderish resources, but `PATCH @tus-upload` works on IContentish resources. That's because of the replace functionality. Replace works on any content object while creating new files works only on folderish objects only.
While looking again at the ZCML I found some minor issues. `HEAD @tus-upload` should be registered for IContentish and `PATCH @tus-upload` should have a conditional permission check (cmf.AddPortalContent if creating a new file, cmf.ModifyPortalContent if replacing an existing file). I can fix that tomorrow.

tisto: @buchi any update on this? I'd like to merge and make a release now. :)

docs/source/tusupload.rst
```diff
@@ -0,0 +1,110 @@
...
+
+When the last partial file has been uploaded, the server will contain the final file URL in the 'location' header.
+
+
+Replacing Existing Files
+------------------------
+
+TUS can also be used to replace an existing file by sending a POST request  to the `@tus-replace` endpoint instead.
+
+TODO: Provide examples for the request/response.
```
jaroel: Would be awesome if we can get this!

buchi:
I've added a fix for the permissions of the `@tus-upload` endpoint.
Can provide the missing example for `@tus-replace` tomorrow.

buchi:
@tisto ready

tisto:
@buchi WOHOOOO!!!

lukasgraf:
🎉 💚 🎆

PR #267, opened 2017-03-13
Add a @translations endpoint to handle translated content

erral:
This addresses #225


erral:
To test this endpoint in Plone 4.x, I created a new entry in .travis.yml and added an additional buildout file that includes LinguaPlone for testing. I did that because of version incompatibilities of LinguaPlone in Plone 5. So far so good.

But now with the `@vocabularies` endpoint merged the full list of vocabularies that is shown in the documentation differs when the tests are run for Plone 4.3 with LinguaPlone and for Plone 4.3 or Plone 5. How can I handle this case in tests?

Any hint will be appreciated

PR #269, opened 2017-03-14
Fix @types endpoint non addable types for member users and local roles

### LINKED ISSUE #268: The @types endpoint is not honoring local roles.
The for non manager users is always returning "false", even if the user has permission to add content in the target folder.

sneridagh:
This fixes #268 .

### DIFF
--- a/src/plone/restapi/tests/test_services_types.py
+++ b/src/plone/restapi/tests/test_services_types.py
@@ -121,10 +121,21 @@ class TestServicesTypes(unittest.TestCase):
             type='Folder',
             title=u'folder',)
 
+        folder_cant_add = api.content.create(
+            container=self.portal,
+            id="folder_cant_add",
+            type='Folder',
+            title=u'folder_cant_add',)
+
         api.user.grant_roles(
             user=user,
             obj=folder,
             roles=['Contributor', ])
+
+        api.user.grant_roles(
+            user=user,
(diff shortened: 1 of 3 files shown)


src/plone/restapi/tests/test_services_types.py
```diff
@@ -105,3 +107,58 @@ def test_file_type(self):
...
+        self.api_session.auth = ('noam', '1234')
+        # In the folder, the user should be able to add types since we granted
+        # Contributor role on it
+        response = self.api_session.get('/folder/@types')
+        response = response.json()
+
+        self.assertIn(
+            'Document', [a['title'] for a in response if a['addable']])
+
+        # In the folder where the user only have Reader role, no types are
```
tisto: @sneridagh I prefer to not have assert statements in a test and then do another action. When this happens I try to split up the tests, so that each tests only has assert statements at the end (precondition, action, assert). This is just a matter of preferences though.

sneridagh: @tisto but in this case, it's not all related to the same state? (with the GET types actions to several objects, we are not changing that state).

tisto: @sneridagh yes, it is. You can still have multiple tests that are related to a single issue. As said, matter of taste.

PR #270, opened 2017-03-14
Add available_roles attribute to the GET @sharing endpoint.

sneridagh:
It's required to build an UI on the top of the sharing feature.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -36,6 +36,9 @@ New Features:
   sharing UI widget or other user/groups search widgets.
   [sneridagh]
 
+- Add ``available_roles`` attribute to the GET @sharing endpoint.
+  [sneridagh]
+
 Bugfixes:
 
 - Return correct @id for folderish objects created via POST.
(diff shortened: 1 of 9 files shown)


CHANGES.rst
```diff
@@ -36,6 +36,9 @@ New Features:
   sharing UI widget or other user/groups search widgets.
   [sneridagh]
 
+- Add ``available_roles`` attribute to the GET @sharing endpoint.
```
tisto: @sneridagh I wonder if we want to include this in our changelog. We add the sharing endpoint in the same release. :)

tisto:
@sneridagh don't we have the sharing output somewhere in our docs? I'm just wondering why there is no update on the _json folder.

CHANGES.rst
sneridagh: True! 😛 Should I then not to add any changelog? Github will complain, but I guess it's bypassable.

sneridagh:
@tisto didn't check it out the docs, I asumed they were ok and with the new format. I will take a look.

tisto:
@sneridagh you would get a red build if the docs would be outdated thanks to @lukasgraf test script. :)

sneridagh:
@tisto the sharing docs has its own format :) that's why it passed the test script! I will fix it.

Issue #283, opened 2017-03-21
The @move/@copy endpoints don't return correct 401/403 codes.


sneridagh:

tisto:
@sneridagh this issue has been solved by https://github.com/plone/plone.restapi/pull/284, correct? Will close this issue now.

PR #284, opened 2017-03-21
Fix for #283 providing 401/403 correct status codes.

sneridagh:


src/plone/restapi/services/copymove/copymove.py
```diff
@@ -1,12 +1,17 @@
 # -*- coding: utf-8 -*-
 from Acquisition import aq_parent
-from Products.CMFCore.utils import getToolByName
+from plone import api
```
tisto: @sneridagh we haven't made a final decision regarding plone.api usage in plone.restapi. If it is not too much work I would remove the dependency here. Keep them in tests is fine though. The alternative would be to get a quick feedback from the @plone/framework-team about this. Though, I guess it would take some time do discuss the details.

src/plone/restapi/tests/test_copymove.py
```diff
@@ -101,8 +62,114 @@ def test_get_object_by_uid(self):
...
+        transaction.commit()
+
+    def test_copy_single_object(self):
+        response = self.api_session.post(
+            '/@copy',
+            json={
+                "source": self.doc1.absolute_url()
+            }
+        )
+        transaction.commit()
```
tisto: @sneridagh sorry for being picky. Just one very minor comment. I usually add a blank line before the assert statements to separate the preconditions/actions from the actual test. I leave it up to you if you want to adopt this...

sneridagh: Added!

Issue #234, opened 2017-03-04
Use plone.api in plone.restapi


tisto:
plone.api is shipped with core but not allowed to use in core packages to prevent circular dependencies and possible infinite loops.

Though, plone.restapi, even if we ship it with core in the future, will always be just a consumer of plone.api. It is highly unlikely and not desirable that plone.api depends in any way on plone.restapi (though, one could come up with examples if you really want).

Therefore we think it is not only save, but also desirable to use plone.api in the core. For one thing this will avoid re-implementing functionality in plone.restapi that already exists in plone.api. Though, this might also lead to a situation in the future where a Python developer might be able to implement plone.restapi endpoints with just mastering plone.api, without the need to understand the underlying stack.

We discussed this matter at the Beethoven Sprint in Bonn today with @jaroel @witsch @sneridagh @bloodbare @vangheem and others and reached an agreement.

CC @plone/framework-team: if you have any objections, please let us know asap. Even though, we do not have a PLIP to include plone.restapi in core yet, it would not be desirable to have to remove the plone.api dependency we are about to introduce.

jensens:
Core is not allowed to do, because we have no clear boundaries in our dependency graph.

So, I would really like to define the line below and above plone.api in Plone core. So plone.restapi is the first consumer of plone.api and this makes perfect sense. I am also open discussing to allow usage of plone.api for other packages in future. 

I think important is to document the boundary and also great would be to have a test ensuring it, if this is possible.

witsch:
@jensens, I completely agree & ftr, what you said was also a significant part of the discussions we had at the sprint.

Meanwhile, I've started to work on a [small tool](https://github.com/witsch/apitizer) to help refactor code to use `plone.api`.  It uses the [rope refactoring library](https://github.com/python-rope/rope) to hopefully automate most of the changes we want.  The first version can rewrite all uses of `getToolByName` (to `plone.api.portal.get_tool`) while keeping the (`plone.restapi`) tests happy.  I'm planning to set up a (throw-away) pull request to reflect any progress there…

tisto:
@jensens I fully agree with you and share your concerns. The idea would be that we make plone.restapi an exception to the no-plone.api rule. I think we can safely assume that no Plone core package ever will rely on plone.restapi. plone.restapi is supposed to be the same facade to HTTP that plone.api is on the Python-level. If this is the case, the only thing we have to make sure is that plone.api never uses any plone.restapi code (something which is highly unlikely).

[...]

PR #289, opened 2017-03-22
Add five.intid, plone.api and z3c.relationfield as dependencies

### LINKED ISSUE #288: plone.restapi 1.0a10 fails in plain Plone 4.3.x
Zope refuses to start using latest Plone 4.3.11 because there is no `zope.intid` package in there. `zope.intid` is used in the `copymove` endpoint (https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/tests/test_copymove.py#L45).

A new requirement should be added to add `five.initd` and/or `zope.intid` in setup.py to fix that.

The test setup doesn't catch this, because the dependency is pulled from some other package.

erral:
Fixes #288

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -42,6 +42,9 @@ setup(name='plone.restapi',
           'setuptools',
           'plone.rest >= 1.0a6',  # json renderer moved to plone.restapi
           'PyJWT',
+          'five.intid',
+          'plone.api',
+          'z3c.relationfield', 
       ],
       extras_require={'test': [
           'Products.Archetypes',
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,7 @@ setup(name='plone.restapi',
           'PyJWT',
           'five.intid',
           'plone.api',
-          'z3c.relationfield', 
+          'z3c.relationfield',
       ],
       extras_require={'test': [
           'Products.Archetypes',
--- a/plone-5.0.x.cfg
+++ b/plone-5.0.x.cfg
@@ -3,3 +3,11 @@ extends =
     base.cfg
     http://dist.plone.org/release/5.0.6/versions.cfg
     versions.cfg
+
+[versions]
+# We need five.intid >= 1.1.2 in Plone 5 because otherwise a NotYet
+# error is raised during the test setup, and that NotYet error is fixed
+# in five.intid 1.1.2 but it's not in the Plone 5.0.6 KGS we are using
+# for testing
+# See https://github.com/plone/five.intid/pull/4 for reference
+five.intid = 1.1.2


tisto:
@erral thanks! Will try to make a new release asap!

setup.py
```diff
@@ -42,6 +42,9 @@
           'setuptools',
           'plone.rest >= 1.0a6',  # json renderer moved to plone.restapi
           'PyJWT',
+          'five.intid >= 1.1.2',
```
tisto: @erral could you please elaborate why we need at least 1.1.2 (and possibly add a note to setup.py). Those requirements usually lead to problems at some point, so I try to avoid them whenever possible.

tisto:
@erral do you have any idea why we did not catch this in our test suite?

erral:
I am running some more tests right know, will report ASAP

erral:
It seems that the requirements are satisfied by the test dependencies, and that's why our test setup did not caught them

- plone.api -> plone.app.linkintegrity -> plone.app.intid -> five.intid
- plone.app.contenttypes -> plone.app.dexterity[relations] -> z3c.relationfield

erral:
@tisto I added the five.intid >= 1.1.2 requirement information on the plone-5.0.x.cfg file

If we update the testing setup to test the latest Plone 5.0.7 (#290), that would be unneeded.

setup.py
```diff
@@ -42,6 +42,9 @@
           'setuptools',
           'plone.rest >= 1.0a6',  # json renderer moved to plone.restapi
           'PyJWT',
+          'five.intid',
+          'plone.api',
```
tisto: @erral the plone.api dependency is a problem. We did not finish our discussion about plone.api: https://github.com/plone/plone.restapi/issues/234

Personally I don't have a problem with introducing that dependency now. Though, we have to think about the future implications...

buchi:
@erral @tisto There should be no dependency on five.intid and z3c.relationfield.
Support for z3c.relationfield should be conditional as it's not part of a vanilla Plone 4.3 site. https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/serializer/configure.zcml#L53
The same for five.intid. It's only required in Plone 5 and should be included automatically by it's dependencies. We do not use five.intid directly in code anywhere afaik and thus we don't have to declare it as a dependency.

tisto:
@buchi good point! Will try to find time to look into that...

setup.py
erral: Then, until the discussion is closed, we should change the `@principals` endpoint not to use plone.api: https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/services/principals/get.py#L45

erral:
@buchi it looks like we have another z3c.relationfield dependency at `@types` https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/types/configure.zcml#L18

Regarding the `five.intid` one, I agree. We don't use `five.intid` directly, but I added it because it is the Zope 2 integration library of `zope.intid` and we use it in the `@copymove` endpoint: https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/services/copymove/copymove.py#L10

erral:
I just sent another PR to work on the dependencies:

https://github.com/plone/plone.restapi/pull/292

buchi:
@erral you're right. Didn't have the intid thing in @copymove in mind.

I'm favoring the approach in #292

tisto:
Closed in favor of https://github.com/plone/plone.restapi/pull/292

PR #290, opened 2017-03-22
Update testing environment to latest Plone 4 and 5 releases

erral:
Plone 4.3.x setup can't be updated to 4.3.12 because some errors on the generated KGS, so I updated it to Plone 4.3.11 (see https://community.plone.org/t/plone-4-3-12-soft-released/3529/14)


erral:
I also had to fix the documentation regarding the vocabularies endpoint, due to changes in new versions of Plone. 

It is not a change of the implementation, just a change in the output.

tisto:
@erral thanks!

PR #292, opened 2017-03-22
Fix plone.api, z3c.relationfield and zope.intid dependencies.

### LINKED ISSUE #288: plone.restapi 1.0a10 fails in plain Plone 4.3.x

erral:
Fixes #288.

This PR removes the use of plone.api and zope.intid, and makes the use of z3c.relationfield conditional on the installation of it (present in Plone 5, not in Plone 4.3.x)

See #289 for further discussion.

Merge only this one or #289, but not both.

### DIFF
--- a/src/plone/restapi/services/copymove/copymove.py
+++ b/src/plone/restapi/services/copymove/copymove.py
@@ -5,9 +5,7 @@ from plone.restapi.services import Service
 from Products.CMFCore.utils import getToolByName
 from zExceptions import BadRequest
 from zope.component import getMultiAdapter
-from zope.component import queryUtility
 from zope.interface import alsoProvides
-from zope.intid.interfaces import IIntIds
 from zope.security import checkPermission
 
 import plone
@@ -25,12 +23,8 @@ class BaseCopyMove(Service):
         self.catalog = getToolByName(self.context, 'portal_catalog')
 
     def get_object(self, key):
-        """Get an object by intid, url, path or UID."""
-        if isinstance(key, int):
-            # Resolve by intid
-            intids = queryUtility(IIntIds)
-            return intids.queryObject(key)
-        elif isinstance(key, basestring):
+        """Get an object by url, path or UID."""
+        if isinstance(key, basestring):
             if key.startswith(self.portal_url):
                 # Resolve by URL
                 return self.portal.restrictedTraverse(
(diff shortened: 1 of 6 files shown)


src/plone/restapi/types/adapters.py
```diff
@@ -32,6 +31,15 @@
 from plone.restapi.types.utils import get_fieldsets
 from plone.restapi.types.utils import get_jsonschema_properties
 
+import pkg_resources
```
buchi: You could move the `ChoiceslessRelationListSchemaProvider` into e separate module. Then you could avoid the conditional import stuff by just using a ZCML condition.

src/plone/restapi/services/copymove/copymove.py
```diff
@@ -25,12 +23,8 @@ def __init__(self, context, request):
         self.catalog = getToolByName(self.context, 'portal_catalog')
 
     def get_object(self, key):
-        """Get an object by intid, url, path or UID."""
```
buchi: We could do a conditional import here, to still have intid support if it's available. But I'm also fine with just dropping it.

src/plone/restapi/services/copymove/copymove.py
```diff
@@ -25,12 +23,8 @@ def __init__(self, context, request):
...
-        if isinstance(key, int):
```
tisto: Removing the lookup by intid is fine with me. @csenger @sneridagh I think one of you guys added that, would you be ok with removing it, or do you have a strong opinion on keeping it? We can always add it properly at a later point if needed.

src/plone/restapi/types/adapters.py
erral: Yes, it is cleaner. I just updated it.

tisto:
LGTM. I'm ok with merging. @buchi are you happy with it as well?

buchi:
@tisto @erral 👍

tisto:
@erral thanks for your fix and pushing this!

PR #294, opened 2017-03-23
Make import of LocalrolesModifiedEvent conditional

lukasgraf:
I propose to make the import of `LocalrolesModifiedEvent` from `plone.app.workflow` conditional, so `plone.restapi` doesn't prevent Plone 4.3 deployments < 4.3.4 from booting.

---

A deployment with `plone.restapi == 1.0a10` and Plone < 4.3.4 will currently fail to boot because of the import of `LocalrolesModifiedEvent` from `plone.app.workflow`. This event has only been [added in `p.a.workflow` 2.2.1](https://github.com/plone/plone.app.workflow/pull/5), resp. backported to 2.1.8, which is available in Plone 4.3.4 or higher.

This currently bites us, preventing us from using the most recent `plone.restapi`, because we still have a few deployments that we can't yet bump to the most recent 4.3.x version.

I'm definitely not advocating here that we should officially claim support for older bugfix versions of Plone 4.3.x, and/or run 4.3 tests against anything else than a recent version. Or that actual *functionality* is supposed to be guaranteed to work on those versions.

But if we can make imports that get triggered during boot time a bit more defensive and through that achieve that `plone.restapi` doesn't completely prevent the instance from even *booting* on those older versions, we'd currently benefit from that.

So, @tisto @sneridagh @csenger would you object to this change?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,11 @@ Changelog
 1.0a11 (unreleased)
 -------------------
 
-- Nothing changed yet.
+Bugfixes:
+
+- Make import of LocalrolesModifiedEvent conditional, so plone.restapi
+  doesn't prevent Plone 4.3 deployments < 4.3.4 from booting.
+  [lgraf]
 
 
 1.0a10 (2017-03-22)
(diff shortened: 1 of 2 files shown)


tisto:
@lukasgraf LGTM. Merged. Expect a release today or during the weekend.

lukasgraf:
@tisto awesome, thank you!

PR #295, opened 2017-03-24
Add Products.PasswortResetTool to dependencies.

tisto:
This dependency is gone in Plone 5.1.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,10 @@ Changelog
 
 Bugfixes:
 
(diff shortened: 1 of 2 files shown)


tomgross:
Hi @tisto I was very unclear in my issue. The dependency to PasswordResetTool is undefined. It's not gone in Plone 5.1 but moved to CMFPlone:
https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/PasswordResetTool.py
The correct fix would be a conditional input.

PR #299, opened 2017-03-24
Remove unittest2 imports.

tisto:


tisto:
@lukasgraf @buchi would that be ok with you? Are you running Plone 4.3 on Python 2.6?

lukasgraf:
@tisto nope, we're on Python 2.7 everywhere where it matters, ok for us 👍

PR #301, opened 2017-03-29
Dockerfile for plone.restapi

tisto:


site.cfg
```diff
@@ -0,0 +1,7 @@
+[buildout]
+extends = buildout.cfg
+eggs += plone.restapi
+
+[versions]
+plone.stringinterp = 1.0.14
```
avoinea: I don't think you need this pin ;)

Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM plone:4
+MAINTAINER "Timo Stollenwerk" <tisto@plone.org>
+
+USER root
+RUN apt-get update \
+ && apt-get install -y --no-install-recommends  build-essential \
```
avoinea: You may not need to install `build-essential` if you don't compile C modules

svx: If you use apt install, you also may want to use apt-get clean

.travis.yml
```diff
@@ -24,3 +24,4 @@ after_success:
   - bin/createcoverage
   - pip install coverage==3.7.1 coveralls
   - coveralls
+  - docker.io build --tag plone-dev:207.03.29 .
```
avoinea: Typo in `207.03.29` ;)

tisto:
@avoinea @svx I have a docker image with plone.restapi that I can push to dockerhub now. To automate the process I guess it would make sense to hook this up with the plone user on dockerhub. Do you agree with that?

gforcada:
@tisto a changelog entry please 😄

avoinea:
@tisto yes ;)

tisto:
@gforcada a changelog entry does not make much sense here in my opinion. This is not a change on the package itself.

tisto:
@avoinea do you know who can help me with the plone credentials for dockerhub? We need to add them to the plone.restapi repo, to be able to use it to make travis automatically upload docker images to dockerhub.

.travis.yml
```diff
@@ -24,3 +24,5 @@ after_success:
   - bin/createcoverage
   - pip install coverage==3.7.1 coveralls
   - coveralls
+  - docker.io build --tag plone.restapi:$(git rev-parse HEAD) .
+#  - docker.io push kitconcept/plone.restapi
```
svx: Reminder: Do not forget to change that to plone, later on, otherwise you will push it to kitconcept on dockr hub and not to plone :)

tisto: Absolutely. I need the plone credentials on the plone.restapi first though. :)

svx: Oh and make sure to check: https://docs.travis-ci.com/user/docker/

You need:

```
services:
- docker
```

and change

```
docker.io
```

to ```docker```

tisto: @svx I don't want to run docker on travis. Just automatically build the latest version and push it to dockerhub. Other projects should use the plone.restapi docker service on travis then in order to run tests against it (with their custom front-ends for instance).

tisto: I am a bit confused. On Ubuntu I need to user docker.io. Maybe travis is using a symlink to be able to use docker instead when you use a docker service. Anyways, will test this as soon as I have the plone dockerhub credentials. :)

svx: Ok, but if you want travis to push you need docker installed, right ?

Yeah, this is when you install docker from the ubuntu repositories, but if you for example install docker from the docker repos it is called docker.

I do not know how travis uses/installs docker there are different ways.

tisto: @svx you are correct. It just turned out my commands weren't executed at all. :(

Issue #298, opened 2017-03-24
Plone 5.1


tisto:
Make sure plone.restapi works with Plone 5.1. Make sure all tests and the travis build passes. cc @tomgross

tisto:
Travis build is in place: https://travis-ci.org/plone/plone.restapi/jobs/214588779 Buildout fails though:

```$ bin/buildout -N -t 3 -c travis.cfg
Setting socket time out to 3 seconds.
Getting distribution for 'mr.developer==1.35'.
Got mr.developer 1.35.
Getting distribution for 'appdirs==1.4.2'.
zip_safe flag not set; analyzing archive contents...
While:
  Installing.
  Loading extensions.
  Getting distribution for 'appdirs==1.4.2'.

An internal error occurred due to a bug in either zc.buildout or in a
recipe being used:
Traceback (most recent call last):
  File "/home/travis/build/plone/plone.restapi/local/lib/python2.7/site-packages/zc/buildout/buildout.py", line 1982, in main
    getattr(buildout, command)(args)
  File "/home/travis/build/plone/plone.restapi/local/lib/python2.7/site-packages/zc/buildout/buildout.py", line 509, in install
    self._load_extensions()
  File "/home/travis/build/plone/plone.restapi/local/lib/python2.7/site-packages/zc/buildout/buildout.py", line 1035, in _load_extensions
    newest=self.newest, allow_hosts=self._allow_hosts)
  File "/home/travis/build/plone/plone.restapi/local/lib/python2.7/site-packages/zc/buildout/easy_install.py", line 913, in install
    return installer.install(specs, working_set)
  File "/home/travis/build/plone/plone.restapi/local/lib/python2.7/site-packages/zc/buildout/easy_install.py", line 714, in install
    for dist in self._get_dist(req, ws):
  File "/home/travis/build/plone/plone.restapi/local/lib/python2.7/site-packages/zc/buildout/easy_install.py", line 563, in _get_dist
    dists = [_move_to_eggs_dir_and_compile(dist, self._dest)]
  File "/home/travis/build/plone/plone.restapi/local/lib/python2.7/site-packages/zc/buildout/easy_install.py", line 1730, in _move_to_eggs_dir_and_compile
    assert newdist is not None  # newloc above is missing our dist?!
AssertionError

travis_time:end:03df3462:start=1490349970611469701,finish=1490349973945631307,duration=3334161606
[0K
[...]

PR #310, opened 2017-04-10
Don't fetch unnecessary PasswordResetTool in Plone 5.1

tomgross:


tisto:
@tomgross thanks! Please note that we currently do not officially support Plone 5.1 due to https://github.com/plone/plone.restapi/issues/298

PR #312, opened 2017-04-17
Add navigation and breadcrumbs as top-level services.

tisto:


tisto:
@jensens I would like to add deprecation messages when someone accesses the components which we are going to remove in future versions. I recall you looked into this a while ago. Would you mind providing me with some pointers what is considered best practice these days?

PR #315, opened 2017-04-18
Add missing endpoints to menu

jaroel:
@groups and @principals endpoints


jaroel:
nevermind, the prinicals endpoint is nowhere near sufficient to implement the sharing tab.

PR #316, opened 2017-04-18
Add search to @sharing

jaroel:


tisto:
@jaroel so the idea would be that you can query @sharing on the portal root instead of the @principals endpoint, correct?

FYI: Even though we are still in alpha, I would prefer to deprecate a method first, before finally removing it at a later version. People are already using plone.restapi in production. Unlike Google/Angular, we take breaking changes seriously. ;)

docs/source/sharing.rst
```diff
@@ -35,6 +35,9 @@ The sharing information of a content object can also be directly accessed by app
 .. literalinclude:: _json/sharing_folder_get.resp
    :language: http
 
+.. note::
+    Searching for users and/or groups who do not yet have a sharing entry can be done by appending the argument `search_term` to the query string. ie search_term=admin
```
tisto: @jaroel I would prefer to use "search" instead of "search_term" to be consistent with the other endpoints.

tisto:
@jaroel is it possible that you forgot to add the "@principals" search implementation? Or would that be another pr?

jaroel:
@tisto you'll use @sharing on context, so the results will include the correct acquired and inherited flags. @principals never supported this and never will, unless it is called on context.
Yeah, @principals should be deprecated first.
I'll make use we can use "search" instead of "search_term".

tisto:
@sneridagh if you have time I would like to hear your opinion on that pull request.

docs/source/sharing.rst
```diff
@@ -35,6 +35,9 @@ The sharing information of a content object can also be directly accessed by app
 .. literalinclude:: _json/sharing_folder_get.resp
    :language: http
 
+.. note::
+    Users and/or groups without a sharing entry can be found by appending the argument `search` to the query string. ie search=admin. It will also flag or global roles.
```
tisto: @jaroel what does "It will also flag or global roles" mean?

src/plone/restapi/serializer/local_roles.py
```diff
@@ -17,10 +17,11 @@ def __init__(self, context, request):
         self.context = context
         self.request = request
 
-    def __call__(self):
+    def __call__(self, search=None):
+        self.request.form['search_term'] = search
         sharing_view = getMultiAdapter((self.context, self.request),
                                        name='sharing')
-        local_roles = sharing_view.existing_role_settings()
+        local_roles = sharing_view.role_settings()
```
tisto: @jaroel this changes the structure of what we return on the sharing view, correct? If this is the case, this is a breaking change.

src/plone/restapi/tests/test_content_local_roles.py
```diff
@@ -38,6 +38,25 @@ def _get_ac_local_roles_block(self, obj):
...
+
+        response = requests.get(
+            self.portal.folder1.absolute_url() + '/@sharing?search=admin',
+            headers={'Accept': 'application/json'},
+            auth=(SITE_OWNER_NAME, SITE_OWNER_PASSWORD),
+        )
+        search_entries = response.json()['entries']
+
+        # Did we find anything?
+        self.assertNotEqual(len(non_search_entries), len(search_entries))
```
tisto: @jaroel could we add some more tests here to make it visible what the sharing search actually returns?

src/plone/restapi/serializer/local_roles.py
jaroel: Nope, it's just that the search results are also in 'entries'

src/plone/restapi/tests/test_content_local_roles.py
jaroel: Sure, no problem!

docs/source/sharing.rst
jaroel: Right, that's just stupid.

src/plone/restapi/tests/test_content_local_roles.py
jaroel: done

sneridagh: @jaroel Can we have a test that verifies that a POST to @sharing can be fed using the result from a previous GET to the endpoint?

docs/source/sharing.rst
```diff
@@ -35,6 +35,10 @@ The sharing information of a content object can also be directly accessed by app
 .. literalinclude:: _json/sharing_folder_get.resp
    :language: http
 
+.. note::
+    Users and/or groups without a sharing entry can be found by appending the argument `search` to the query string. ie search=admin.
```
sneridagh: @jaroel it would be nice to include a documentation example of the usage of the search term to test_documentation.py . I think it will be common enough use case to have it's own example.

src/plone/restapi/tests/test_content_local_roles.py
jaroel: That would be the same a normal post, wouldn't it?

src/plone/restapi/serializer/local_roles.py
tisto: Ok. Fine with me then.

src/plone/restapi/tests/test_content_local_roles.py
tisto: Yes it would. The idea is just to make sure that you can GET first, slightly manipulate the result and POST it back.

docs/source/sharing.rst
tisto: +1

tisto:
Apart from the minor issues @sneridagh and I raised this PR looks good to me and can be merged.

src/plone/restapi/tests/test_content_local_roles.py
jaroel: @tisto / @sneridagh I've added a test for this

docs/source/sharing.rst
jaroel: @tisto / @sneridagh I've added a test for this

tisto:
@sneridagh could you please have a second look at this pr and let me know if you are happy with it now?

PR #318, opened 2017-04-19
Remove the disabled flag from @sharing entries

jaroel:
This is a UI flag for Plone templates and does not actually specify that the entry is disabled.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,8 +4,10 @@ Changelog
 1.0a14 (unreleased)
 -------------------
 
-- Nothing changed yet.
-
+Bugfixes:
+- Remove 'disabled' from @sharing entries, as it is a Plone template hint.
+  It doesn't actually mean that the entry is disabled.
+  [jaroel]
 
 1.0a13 (2017-04-18)
 -------------------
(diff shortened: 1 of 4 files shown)


jaroel:
Maybe some client already depends on this?

tisto:
@jaroel I get that the naming is misleading. Could you elaborate what it actually means? I am wondering why we don't display certain entries in the existing (old-school) sharing view.

jaroel:
@tisto We do display the entries in the existing UI, but the inputs are disabled.
See https://github.com/plone/plone.app.workflow/blob/master/plone/app/workflow/browser/sharing.py#L266

tisto:
@jaroel thanks for the clarification. Though, don't we need this information to render the users/roles table in JS then? One could argue that we should fix the naming. Though, I think we still need this kind of information in the front-end.

tisto:
@jaroel reminder: please reply to my last comment. :)

jaroel:
@tisto right, in plone-react we do disable the UI, but the logic for that is in the UI.
I've removed this field from the response so we only deal with actual state.
I can add it again, though

jaroel:
Lets leave it in; might be easy for UI purposes.

PR #322, opened 2017-04-19
Content reordering support

jaroel:
Allow reordering of sub resources by sending a PATCH "ordering".

ie
`"ordering": {"obj_id": [str:object_id], "delta": ["top"|"bottom"|int, "subset_ids": [str:object_id, [str:object_id..]]}`


docs/source/_json/types.resp
```diff
@@ -3,63 +3,68 @@ Content-Type: application/json
 
 [
   {
-    "@id": "http://localhost:55001/plone/@types/Collection",
```
tisto: @jaroel you really have to stop opening the generated _json files with your editor. :) This makes it hard to review your pull requests.

jaroel: @tisto It was an .editorconfig in the plone-react root, as I develop for that currently.
I've added an .editorconfig here that prevents this behaviour for .resp files.

docs/source/_json/types.resp
```diff
@@ -18,6 +18,11 @@ Content-Type: application/json
     "title": "DXTestDocument"
   }, 
   {
+    "@id": "http://localhost:55001/plone/@types/DXTestFolder",
```
tisto: @jaroel can you elaborate why you need the DXTestFolder type here? I know we added this for AT. Though, I somehow dislike test profiles because they obfuscate the test setup (you need to look into many different external files to understand your test set up). A test type can be easily set up within Python code and this improves the readability of the test. 

Therefore I would like to get rid of the ATTestFolder in the profile instead of adding the DX equivalent.

jaroel: @tisto I don't need DXTestFolder at all. Standard Folder and Document are enough.
I'll remove it and cleanup a bit.

jaroel:
@tisto the DXTestFolder is gone. This will have to do.

Issue #325, opened 2017-04-24
The @move endpoint fails to return 403 when the user don't have proper delete permissions over the parent.


sneridagh:
The endpoint should check for delete permissions on the parent too.

csenger:
Close. #326 was merged.

PR #326, opened 2017-04-24
Fix the @move endpoint fails to return 403 when the user don't have p…

sneridagh:
…roper delete permissions over the parent folder


sneridagh:
This fixes #325

src/plone/restapi/tests/test_copymove.py
```diff
@@ -201,6 +201,7 @@ def test_move_single_object_no_auth_raises_401(self):
     def test_move_single_object_no_permission_delete_source_raises_403(self):
         api.user.grant_roles(
             username='memberuser', obj=self.folder1, roles=['Manager', ])
+        api.content.transition(obj=self.doc1, transition='publish')
```
tisto: @sneridagh why do we need to publish here?

sneridagh: otherwise, the user can't see the admin's doc, and the service fails not for the 403 but because it's unable to see the source doc.

PR #336, opened 2017-04-28
history and versioning

### LINKED ISSUE #281: Endpoint to manage history
The workflow endpoint does contain the history of all workflow transition but doesn't include all revisions. The new endpoint should return all the revisions and it should also be possible to get the content of a specific revision and to restore a specific revision.

jaroel:
`@history` endpoint GET list workflow and editing history
`@content` endpoint support getting historical versions by adding `version_id` querystring parameter

Todo:
- [x] consolidate timestamps in `@history`. Some are floats, some are datetimes.
- [x] checkPermission when trying to view older versions. (fix by endpoint permission)
- [x] add revert to version endpoint
- [x] add link to content endpoint at specific version_id

Fixes #281


docs/source/_json/history_get.resp
```diff
@@ -0,0 +1,36 @@
...
+  {
+    "action": "Edited", 
+    "actor": {
+      "@id": "http://localhost:55001/plone/@users/test-user", 
+      "fullname": "test-user", 
+      "id": "test-user", 
+      "username": null
+    }, 
+    "comments": null, 
+    "content_url": "http://localhost:55001/plone/front-page/?version_id=0",
```
tisto: @jaroel I think the content_url attribute should be a hypermedia link with the "@id" attribute. This is the link to the revision of that specific document, correct?

docs/source/_json/history_get.resp
```diff
@@ -0,0 +1,36 @@
+HTTP/1.1 200 OK
+Content-Type: application/json
+
+[
+  {
+    "action": null,
```
tisto: @jaroel I guess the null value for content creations comes from CMFEditions. This should be "Created" in my opinion. plone.restapi might be the wrong level to fix this though.

docs/source/history.rst
```diff
@@ -0,0 +1,43 @@
...
+- time: when this action occured in ISO format.
+- transition_title: the workflow transition's title, for 'Edited' for versioning.
+- type: 'workflow' for workflow changes, 'versioning' for editing, or null for content creation.
+- version_id: identifier for this specific version of the resource.
+
+
+PATCH revert historical versions
+--------------------------------
+
+Reverting to older versions of a resource can be done by issueing a PATCH request with a version_id.
```
tisto: @jaroel I did not come to a final conclusion yet. Though, I am not convinced yet that PATCH is the correct HTTP verb. PATCH is a pretty specific HTTP verb to indicate a change on a document by providing a DIFF of the original document. 

One could argue (and this is the reason for your choice I guess), that a version change is just a minor change to the original document. Though, a PUT is usually used to completely replace the document. A PATCH with just the version number will lead to much more on the server than just changing the version_id on the document.

If we would use PUT, we would have to send the entire document though (or at least the link to that document).

There are people that argue the body of a PATCH should include a real diff of the original document only:

http://williamdurand.fr/2014/02/14/please-do-not-patch-like-an-idiot/

In any case, I think we need to make sure that however we use PATCH, we use it in a consistent way. I am just thinking out loud here and I am open for suggestions. cc @buchi @lukasgraf @sneridagh @bloodbare

docs/source/_json/history_get.resp
tisto: Excuse my ignorance about how CMFEditions work. Why does the first entry does not provide a "content_url" and it does not start with version 0? I guess this exposes how Plone works internally (or used to work with AT). The object is always created empty first and then "edited" for the first time when the user submits the add form? If this is the case one could argue that this is an implementation detail that an API should hide.

docs/source/history.rst
jaroel: @tisto yep, I ran into this too.

Maybe if we issue a PATCH on the actual resource with version_id?
In the end it is up to the server to handle a request the way it wants too. So, if we get a PATCH {'version_id': 3} on a resource, the server may -as a side effect- revert to the specified version.

A `POST @history_revert {'version_id': x}` might be an option.

docs/source/_json/history_get.resp
jaroel: @tisto I've changed this.

docs/source/_json/history_get.resp
jaroel: @tisto I'm not sure if we can change this anymore. I can imaging that there are existing application that do something with the None/null value.

Maybe we should add this to plone.api (see https://github.com/plone/plone.api/issues/305) and use this method in here.

docs/source/_json/history_get.resp
jaroel: @tisto the 'created' bit is a type workflow change (initial state in this case). Only versioning (editing) events create a new version.

See https://github.com/plone/plone.app.layout/blob/master/plone/app/layout/viewlets/content.py#L354

Maybe this should be split up in two endpoints, versioning_history and workflow_history.

docs/source/_json/history_get.resp
tisto: I agree. If we change things, plone.api might be the right level to do it.

docs/source/_json/collection.resp
```diff
@@ -71,5 +71,6 @@ Content-Type: application/json
   "sort_reversed": null, 
   "subjects": [], 
   "text": null, 
-  "title": "My Collection"
+  "title": "My Collection", 
+  "version_id": "current"
```
tisto: @jaroel is there a specific reason why you choosed "version_id" instead of just "version"?

docs/source/_json/content_get_versioned.req
```diff
@@ -0,0 +1,3 @@
+GET /plone/folder/my-document?version_id=0 HTTP/1.1
```
tisto: @jaroel it might be more consistent with our current API design if we would retrieve a specific version with "my-document/@history/<VERSION_ID>". The hypermedia links would also be a bit nicer because we would provide a collection-like API response on the /@history endpoint:
```

[
    {
        "@id": "http://localhost:55001/plone/front-page/@history/0",
        "action": null,
    },
    {
        "@id": "http://localhost:55001/plone/front-page/@history/1",
        "action": "Edited",
    }
]

```

Though, this might lead to a problem with our initial revision that does not have a version id.

jaroel: @tisto yes, that makes more sense!

docs/source/_json/collection.resp
jaroel: @tisto nope, just to have it the same as CMFEditions

docs/source/content.rst
```diff
@@ -111,6 +111,15 @@ After a successful POST, we can access the resource by sending a GET request to
 ..  http:example:: curl httpie python-requests
     :request: _json/content_get.req
 
+.. _content_get_version:
```
tisto: @jaroel I think this needs to be removed, right?

jaroel: yep

jaroel:
@tisto please review

docs/source/history.rst
```diff
@@ -0,0 +1,51 @@
...
+- version: identifier for this specific version of the resource.
+
+
+Get a Historical Version
+------------------------
+
+Older versions of a resource can be retrieved by appending `version` to the @history endpoint url.
+
+..  http:example:: curl httpie python-requests
+    :request: _json/history_get_versioned.req
```
lukasgraf: No example response for retrieving a version?

PR #339, opened 2017-04-30
Controlpanels Endpoint

tisto:


jaroel:
Refs #342

src/plone/restapi/tests/test_documentation.py
```diff
@@ -837,6 +844,20 @@ def test_history_revert(self):
         response = self.api_session.patch(url, json={'version': 0})
         save_request_and_response_for_docs('history_revert', response)
 
+    @unittest.skipIf(not PLONE5, 'Just Plone 5 currently.')
```
tisto: @jaroel how can we retrieve the controlpanel request/response if we just run the tests on Plone 4?

docs/source/_json/controlpanels_get.resp
```diff
@@ -0,0 +1,15 @@
+HTTP/1.1 200 OK
+Content-Type: application/json
+
+[
+  {
```
tisto: @jaroel how do you make it possible that you just show the mail and editing control panel and not all available control panels?

jaroel: The available controlpanels in the rest API are define in the service/controlpanels/configure.zcml, as most -but not all- controlpanels are registry based.

tisto: @jaroel that's why I am asking, we have more than two controlpanels listed there and just two show up in the response.

tisto:
LGTM

PR #353, opened 2017-05-04
Add batched listing to `@registry` endpoint

### LINKED ISSUE #347: No way to list all registry entries.
We do not provide a @registry endpoint that lists all the existing registry entries. Problem is that we will need batching on that which will lead to a situation where some of our endpoints return batched results and some don't.

jaroel:
Fixes #347

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,10 @@ Changelog
 1.0a15 (unreleased)
 -------------------
 
-- Nothing changed yet.
+New Features:
+
+- Add batched listing of registry entries to @registry endpoint.
+  [jaroel]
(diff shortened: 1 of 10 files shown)


tisto:
Crap. The registry entries considerably differ between Plone 4.3 and 5.0. Therefore our _json output check fails for Plone 4.3. :(

jaroel:
Yep. Maybe we should split the examples (req/resp) up in Plone 4 and Plone 5. Something like:
```
Example output for Plone 5:
...
Example output for Plone 4:
...

```

tisto:
Thanks @jaroel!

PR #357, opened 2017-05-16
@comments endpoint

jaroel:


src/plone/restapi/tests/test_documentation.py
```diff
@@ -809,10 +809,6 @@ def setUp(self):
         self.portal = self.layer['portal']
         self.portal_url = self.portal.absolute_url()
 
-        # Register custom UUID generator to produce stable UUIDs during tests
-        pushGlobalRegistry(getSite())
```
lukasgraf: @jaroel The fact that `register_static_uuid_utility()` is called multiple times isn't the issue here - the issue is that after you push a new global component registry on the stacked registry, you'll also need to **pop** it from the stack in `tearDown()` (see the [existing test case](https://github.com/plone/plone.restapi/blob/2d37bc77265cac5ea73fe9e151c4a6481dcb0b35/src/plone/restapi/tests/test_documentation.py#L167-L169)).

Also, because you don't stop `self.time_freezer` in `tearDown()`, the time freezing bleeds through to the testrunner, and that's why you get stuff like `Ran 216 tests with 0 failures and 0 errors in -17858596.730 seconds.`.

jaroel: Nice, thanks!

src/plone/restapi/services/discussion/conversation.py
```diff
@@ -0,0 +1,203 @@
...
+
+        action = form.actions['comment']
+        data, errors = form.extractData()
+        if errors:
+            raise BadRequest({'errors': [err.error for err in errors]})
+
+        form.handleComment(form=form, action=action)
+
+        self.request.response.setStatus(204)
+        if 'location' in self.request.response.headers:
```
tisto: @jaroel the server should set the location header to the URL of the newly created comment to allow the client to display the new comment.

src/plone/restapi/services/discussion/conversation.py
```diff
@@ -0,0 +1,203 @@
...
+        action = form.actions['comment']
+        data, errors = form.extractData()
+        if errors:
+            raise BadRequest({'errors': [err.error for err in errors]})
+
+        comment.modification_date = datetime.utcnow()
+        form.handleComment(form=form, action=action)
+
+        self.request.response.setStatus(204)
+        if 'location' in self.request.response.headers:
```
tisto: @jaroel same here. the location header should be sent.

src/plone/restapi/services/discussion/conversation.py
```diff
@@ -0,0 +1,203 @@
...
+        permission.
+        """
+        if PLONE4:
+            return self.can_review()
+        return getSecurityManager().checkPermission(
+            'Delete comments', aq_inner(reply)
+        )
+
+    def delete_own_comment_allowed(self):
+        if PLONE4:
```
tisto: @jaroel could you elaborate this please? And then add a comment to the code so others with the same question understand why Plone 4 and 5 are different here?

src/plone/restapi/services/discussion/conversation.py
```diff
@@ -0,0 +1,203 @@
...
+            return False
+        # Check if delete own comments is allowed in the registry
+        registry = queryUtility(IRegistry)
+        settings = registry.forInterface(IDiscussionSettings, check=False)
+        return settings.delete_own_comment_enabled
+
+    def could_delete_own(self, comment):
```
tisto: @jaroel why not can_delete_own to be consistent?

jaroel:
@tisto all issues are resolved

PR #361, opened 2017-05-17
Users global roles

tisto:


tisto:
@jaroel LGTM. You need to add a changelog entry.

jaroel:
@tisto thanks, I've added it.

PR #363, opened 2017-05-18
Add schema to registry listing

jaroel:
This allows frontend developers to implement a full registry interface

### DIFF
--- a/docs/source/_json/registry_get_list.resp
+++ b/docs/source/_json/registry_get_list.resp
@@ -6,32 +6,76 @@ Content-Type: application/json
   "batching": {
     "@id": "http://localhost:55001/plone/@registry", 
     "first": "http://localhost:55001/plone/@registry?b_start=0", 
-    "last": "http://localhost:55001/plone/@registry?b_start=275", 
+    "last": "http://localhost:55001/plone/@registry?b_start=1500", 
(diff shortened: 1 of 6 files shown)


docs/source/_json/registry_get_list.resp
```diff
@@ -12,26 +12,70 @@ Content-Type: application/json
   "items": [
     {
       "name": "Products.CMFPlone.i18nl10n.override_dateformat.Enabled", 
+      "schema": {
```
tisto: @jaroel @sneridagh sorry for being too late on this. I was wondering if we want to wrap those attributes into another schema attribute. We are not using that pattern anywhere else and this is not a schema (=JSON schema). I haven't made up my mind but what do you guys think about flattening that and move description,title, and type to the top level to keep it as simple as possible?

jaroel: I get what your saying. I've added this to 'schema' because it uses a JsonSchemaProvider and produces a lot more information than just title description and type.
See for example https://github.com/plone/plone.restapi/pull/363/files#diff-1cd823c9eb155c78b8fd8be2b6fdc818R180

jaroel: @tisto how 'bout https://github.com/plone/plone.restapi/compare/registry-typing?expand=1 ?

PR #364, opened 2017-05-18
Add/update user global roles

jaroel:


lukasgraf:
Is the `@user` endpoint really the right place for this? Global roles can not only be assigned to users, but to principals in general (users as well as groups). Also, how would one *list* the available global role assignments without iterating over all users/groups?

jaroel:
@lukasgraf listing all assignments is not something I would implement because of exactly that. I've added a `@roles` endpoint for getting the role definitions in #365

jaroel:
@lukasgraf the remark about principals is correct and we're looking into that. Thank you!

tisto:
@lukasgraf I think no matter what we decide, changing roles of a user is a feature that we want. I will merge this pr now.

Issue #331, opened 2017-04-25
Specialise ISerializeToJson


jaroel:
We have multiple ISerializeToJson implementations with different signatures for `__call__`.
For content objects we require a `version_id` in `__call__`.
`BrainSerializer` needs `metadata_fields`, as does `LazyCatalogResultSerializer`.
`SerializeVocabularyToJson` and `SerializeTermToJson` need `vocabulary_id`.

tisto:
@jaroel could you please provide a list with pointers to the different implementations?

jaroel:
Content serializers:
https://github.com/plone/plone.restapi/blob/history/src/plone/restapi/serializer/atcollection.py#L17
https://github.com/plone/plone.restapi/blob/history/src/plone/restapi/serializer/atcontent.py#L33
https://github.com/plone/plone.restapi/blob/history/src/plone/restapi/serializer/atcontent.py#L83
https://github.com/plone/plone.restapi/blob/history/src/plone/restapi/serializer/collection.py#L17
https://github.com/plone/plone.restapi/blob/history/src/plone/restapi/serializer/dxcontent.py#L43
https://github.com/plone/plone.restapi/blob/history/src/plone/restapi/serializer/dxcontent.py#L114

Others:
BrainSerializer: https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/serializer/catalog.py#L39
LazyCatalogResultSerializer: https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/serializer/catalog.py#L74
SerializeVocabularyToJson: https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/serializer/vocabularies.py#L20
SerializeTermToJson: https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/serializer/vocabularies.py#L42

The interface https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/interfaces.py#L21 specifies nothing. Technically the implementations are correct.

We could define `__call__` with **args and **kwargs. This would allow the interface.
The downside is that content serializers might not implement the actual signature used in the services.

jensens:
Closing as outdated (8 years old). About Archetypes serializer inconsistencies. Archetypes removed in Plone 6.
If you think this issue is still relevant please reopen it. Thank you for the original report!

davisagli:
This was not primarily about Archetypes. There are still some inconsistencies we could review.

jensens:
[...]

PR #370, opened 2017-06-03
Search full objects retrieval

### LINKED ISSUE #369: Retrieving full objects with @search
We could add a `&fullobjects` parameter to indicate we want the results as full serialized objects instead of just catalog metadata.

I know it will imply to wake up objects so it can be a performance issue. But in some uses cases it would be very useful:

- on our home page, we want to display a carousel showing the 5 last news items: it would be nice to ba able to load the 3 news with one unique request retrieving texts, images, etc.
- the site manager has a folder where he/she manages useful links (using the Link content type): in order to display them in block on a given page, we need to retrieve the remoteUrl for each of those links.

At the moment, for those kind of use cases we need to perform as many GET requests as there are contents to retrieve; So that's also kind of weak regarding performances.

ebrehault:
Fix #369

### DIFF
--- a/docs/source/searching.rst
+++ b/docs/source/searching.rst
@@ -99,3 +99,20 @@ The metadata from those columns then will be included in the results.
 In order to specify multiple columns, simply repeat the query string parameter once for every column name (the ``requests`` module will do this automatically for you if you pass it a list of values for a query string parameter).
 
 In order to retrieve all metadata columns that the catalog knows about, use ``metadata_fields=_all``.
+
+
+Retrieving full objects
+-----------------------
+
+If the data provided as metadata is not enough, you can retrieve search results as full serialized objects equivalent to what the resource GET request would produce.
+
+You do so by specifying the ``fullobjects`` parameter:
+
+.. code-block:: http
+
+    GET /plone/@search?fullobjects&SearchableText=lorem HTTP/1.1
+    Accept: application/json
+
+.. warning::
+
+    Be aware that this might induce performance issues when retrieving a lot of resources. Normally the search just serializes catalog brains, but with full objects we wake up all the returned objects.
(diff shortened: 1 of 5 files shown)


ebrehault:
@tisto @lukasgraf @jaroel tests are passing with Plone 5 but with Plone 4.3 it fails saying:
"ERROR: There are modified files in docs/source/_json/ after running test_documentation.py!"
And the changes I see seems very unrelated with my PR (see https://travis-ci.org/plone/plone.restapi/jobs/239063105).

Is there anything special I should do to generated the _json files properly? (I just ran `bin/test` on a Plone 5 install)

sneridagh:
You should run the tests in Plone 4, still using it as "canonical".

ebrehault:
Fixed. Thanks @sneridagh :)

docs/source/searching.rst
```diff
@@ -99,3 +99,20 @@ The metadata from those columns then will be included in the results.
...
+You do so by specifying the ``fullobjects`` parameter:
+
+.. code-block:: http
+
+    GET /plone/@search?fullobjects&SearchableText=lorem HTTP/1.1
+    Accept: application/json
+
+.. warning::
+
+    Be aware it might induces performance issues when retrieving a lot of resources. Regular search just uses lazy objects, but with full objects retrieving we wake up all the returned objects.
```
lukasgraf: => `Be aware that this might induce performance issues when retrieving a lot of resources. Normally the search just serializes catalog brains, but with full objects we wake up all the returned objects.`

(*Resultsets are still lazy, in the sense that only the objects in the returned batch(es) are woken up*)

lukasgraf:
Just a small suggestion for the docs, otherwise this looks good to me! 👍 

Although this does reinforce the issue described by @jaroel in #331, where the `__call__` signatures of `ISerializeToJson` are different everywhere. But that issue already existed for the `@search` endpoint, so I think we'll have to look at #331 separately.

ebrehault:
@lukasgraf fixed, thanks for the suggestion

jaroel:
How about `full=true` instead of `fullobjects=true`?

PR #386, opened 2017-07-26
Translate titles in @workflow

csenger:


tisto:
@csenger I find it strange that we do not have to amend any test for this. Not your fault of course. Will look into writing a test for this...

sneridagh:
LGTM, but +1 to try to add a test for this

PR #387, opened 2017-07-26
Locking

buchi:
Provides endpoints for locking/unlocking.

### DIFF
--- a/docs/source/_json/lock.resp
+++ b/docs/source/_json/lock.resp
@@ -1,0 +1,12 @@
+HTTP/1.1 200 OK
+Content-Type: application/json
+
+{
+  "creator": "admin", 
+  "locked": true, 
+  "name": "plone.locking.stealable", 
+  "stealable": true, 
+  "time": 1477076400.0, 
+  "timeout": 600, 
+  "token": "0.684672730996-0.25195226375-00105A989226:1477076400.000"
+}
--- a/docs/source/locking.rst
+++ b/docs/source/locking.rst
@@ -46,7 +46,7 @@ To unlock an object send a POST request to the ``/@unlock`` endpoint.
 ..  http:example:: curl httpie python-requests
     :request: _json/unlock.req
 
-The server responds with status 200 (OK) and returns the lock information.
+The server responds with status :term:`200 OK` and returns the lock information.
 
 .. literalinclude:: _json/unlock.resp
    :language: http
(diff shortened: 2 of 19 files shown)


docs/source/_json/lock.resp
```diff
@@ -0,0 +1,12 @@
+HTTP/1.1 200 OK
+Content-Type: application/json
+
+{
+  "creator": "admin", 
+  "locked": true, 
+  "name": "plone.locking.stealable", 
+  "stealable": true, 
+  "time": 1477076400.0,
```
tisto: @buchi I am wondering if we have other "time" attributes in plone.restapi. We have to be sure, that we are consistent in what we expose.

docs/source/locking.rst
```diff
@@ -0,0 +1,80 @@
+Locking
+=======
+
```
tisto: @buchi we should add a short introduction what locking is. I would like to write the documentation in a way that a developer without deep knowledge of zope/plone can understand the basics.

docs/source/locking.rst
```diff
@@ -0,0 +1,80 @@
...
+Locking an object
+-----------------
+
+To lock an object send a POST request to the ``/@lock`` endpoint.
```
tisto: @buchi I'd add the information that the lock endpoint can be called on any object in Plone. Otherwise the reader might think this is a root-level endpoint.

docs/source/_json/lock.resp
buchi: @tisto I'm not super happy with the term "time" but I'm using the same term as in plone.locking for consistency reasons (see https://github.com/plone/plone.locking/blob/master/plone/locking/lockable.py#L150). I'm also ok to rename it to "modified" or "modification-time".

tisto: @buchi yeah. I got that you just mapped over the attributes. This is what we did everywhere so far. On one hand an API should not expose internal data structures, on the other hand re-inventing new names might confuse devs even more. I am ok with leaving this as it is. I just wanted to bring it up.

buchi:
Just realized that locks in Plone are only checked in the UI. You can still modify a locked object without providing the lock token...
For plone.restapi we need to make sure that locked objects can only be modified if the request contains the lock token. I will add that check.

docs/source/locking.rst
tisto: @buchi fyi: I added an introduction, please check if you are happy with it.

tisto: done

buchi: 👍

PR #388, opened 2017-07-26
Show expired content when GET on a folderish object, include a way to…

sneridagh:
… display it on @search via the show_inactive parameter


sneridagh:
Related to #388

sneridagh:
@ebrehault didn't tackled yet the exclude_from_nav since it will require more thought and the addition of the index...

buchi:
This exposes more information than the default Plone interface.
Plone's search view never shows inactive items for users without "Access inactive portal content". See: https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/browser/search.py#L126
https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/CatalogTool.py#L458

Also `folder_listing` doesn't show inactive content without "Access inactive portal content" permission.

We should already have the default behavior, meaning if you have "Access inactive portal content" permission you will automatically get expired content in @search and items listing.

@sneridagh @ebrehault What's your use case for this? Doesn't the default behavior work out for you?

jensens:

PR #390, opened 2017-07-28
Add documentation about updating locked objects

buchi:
@tisto I've added an example how to update locked objects.


tisto:
@buchi thanks!

PR #392, opened 2017-08-03
Fixes #377

### LINKED ISSUE #377: TimeZone conflicts while deserializing
there should be check in data type if two value being compared are coming from different functions
[codeline this bug refers to](https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/deserializer/dxcontent.py#L81)

Timezone to DateTime value only be assigned dynamically if tzinfo not present in value.
[codeline this issue refer to](https://github.com/plone/plone.restapi/blame/master/src/plone/restapi/deserializer/dxfields.py#L49)

Also, tzinfo must not be always set to localZone() but to UTC, as sometimes server machine DateTime is not correctly set.

[Error it cause in plone.importexport](https://github.com/Shriyanshagro/plone.importexport/issues/4)

Shriyanshagro:


loechel:
This fix approach was part of GSoC project plone.importexport. It won't solve the actual problem. I will submit a new fix that solves #396 correctly.

Therefor I dismiss and close this pull request.

PR #394, opened 2017-08-11
Change the serialization of the datetime objects, and store them as o…

sneridagh:
…ffset-aware UTC-based objects

### DIFF
--- a/docs/source/deserialization.rst
+++ b/docs/source/deserialization.rst
@@ -6,13 +6,13 @@ It's worth to note an special case when deserializing Datetimes objects, and how
 Although not supported by Plone itself yet, plone.restapi will store all the Datetimes that will be handling along with its timezone converted to UTC.
 This will provide a common ground for all the datetimes operations.
 
-There is an special case when using datetimes objects in p.a.event, and its behavior is different due to implementation differences for versions 1.x (Plone 4) and 2.x and above (Plone 5).
+There is a special case when using datetimes objects in p.a.event, and its behavior is different due to implementation differences for versions 1.x (Plone 4) and 2.x and above (Plone 5).
 
 .. warning::
   In case of using zope.schema date validators you should also use a datetime object that also contains offset-aware object as the validator value.
 
 .. note::
-  This does not applies in case that you are using Plone 4 with no Dexterity support at all or not p.a.event installed.
+  This does not apply in case that you are using Plone 4 with no Dexterity support at all or not p.a.event installed.
 
 p.a.event 1.x in Plone 4
 ------------------------
--- a/docs/source/deserialization.rst
+++ b/docs/source/deserialization.rst
@@ -49,7 +49,7 @@ The final stored datetime takes this field into account and adds the correct off
 
 and builds the `start` and `end` fields with the proper timezone, depending on the `timezone` field. It also returns the datetime object with the proper timezone offset.
 
-If using Plone 4 and p.a.event 1.x you should construct the Event type using this approach, otherwise the Event object will be created with a wrong timezones.
+If using Plone 4 and p.a.event 1.x you should construct the Event type using this approach, otherwise the Event object will be created with a wrong timezone.
 
 This approach was counterintuitive, and thus, it was changed it Plone 5 version of p.a.event.
 
--- a/docs/source/deserialization.rst
+++ b/docs/source/deserialization.rst
@@ -1,0 +1,87 @@
+Deserialization
+===============
+
+It's worth to note an special case when deserializing Datetimes objects, and how plone.restapi will handle them.
+
+Although not supported by Plone itself yet, plone.restapi will store all the Datetimes that will be handling along with its timezone converted to UTC.
+This will provide a common ground for all the datetimes operations.
+
+There is an special case when using datetimes objects in p.a.event, and its behavior is different due to implementation differences for versions 1.x (Plone 4) and 2.x and above (Plone 5).
+
+.. warning::
+  In case of using zope.schema date validators you should also use a datetime object that also contains offset-aware object as the validator value.
+
+.. note::
+  This does not applies in case that you are using Plone 4 with no Dexterity support at all or not p.a.event installed.
+
+p.a.event 1.x in Plone 4
+------------------------
+
+The implementation of p.a.event in 1.x requires to provide a `timezone` schema property along with the Event type information, otherwise the creation fails, because it's a required field, like this::
+
+.. code-block:: json
+
+  {
+    "@type": "Event",
+    "id": "myevent2",
+    "title": "My Event",
+    "start": "2018-01-01T10:00:00",
+    "end": "2018-01-01T12:00:00",
+    "timezone": "Asia/Saigon"
+  }
+
+The final stored datetime takes this field into account and adds the correct offset to the content type (abbreviated JSON response)::
+
+.. code-block:: json
+
+  {
+    "@id": "http://localhost:55001/plone/folder1/myevent2",
+    "@type": "Event",
+    "UID": "bcfc3914ea174cc1aa8042147edfa33a",
+    "creators": ["admin"],
+    "description": "",
+    "end": "2018-01-01T12:00:00+07:00",
+    "id": "myevent2",
+    "start": "2018-01-01T10:00:00+07:00",
+    "timezone": "Asia/Saigon",
+    "title": "My Event",
+    }
+
+and builds the `start` and `end` fields with the proper timezone, depending on the `timezone` field. It also returns the datetime object with the proper timezone offset.
+
+If using Plone 4 and p.a.event 1.x you should construct the Event type using this approach, otherwise the Event object will be created with a wrong timezones.
+
+This approach was counterintuitive, and thus, it was changed it Plone 5 version of p.a.event.
+
+p.a.event 2.x in Plone 5
+------------------------
+
+The implementation of p.a.event in 2.x no longer requires to provide a `timezone` schema property, because the timezone is computed taking the timezone already existent in dates supplied::
+
+.. code-block:: json
+
+  {
+    "@type": "Event",
+    "id": "myevent2",
+    "title": "My Event",
+    "start": "2018-01-01T10:00:00+07:00",
+    "end": "2018-01-01T12:00:00+07:00",
+  }
+
+You should pass the timezone information in the ISO8601 format, otherwise the system will fallback to UTC. The response given is also computed given the timezone information supplied, but this time it's UTC based:
+
+.. code-block:: json
+
+  {
+    "@id": "http://localhost:55001/plone/folder1/myevent2",
+    "@type": "Event",
+    "UID": "4c031960718246db86c97685f83047ee",
+    "creators": ["admin"],
+    "description": "",
+    "end": "2018-01-01T05:00:00+00:00",
+    "id": "myevent2",
+    "start": "2018-01-01T03:00:00+00:00",
+    "title": "My Event",
+  }
+
+This approach is better because all Javascript serializers/deserializers works with UTC based dates (e.g. .toJSON Javascript API).
--- a/docs/source/deserialization.rst
+++ b/docs/source/deserialization.rst
@@ -1,9 +1,9 @@
 Deserialization
 ===============
 
-It's worth to note an special case when deserializing Datetimes objects, and how plone.restapi will handle them.
+It's worth to note a special case when deserializing datetime objects, and how plone.restapi will handle them.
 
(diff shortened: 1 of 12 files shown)


sneridagh:
This fixes #253

csenger:
Looks good to me.

src/plone/restapi/deserializer/dxfields.py
```diff
@@ -46,8 +46,7 @@ def __call__(self, value):
         try:
             # Parse ISO 8601 string with Zope's DateTime module
             # and convert to a timezone naive datetime in local time
-            value = DateTime(value).toZone(DateTime().localZone()).asdatetime(
-            ).replace(tzinfo=None)
+            value = DateTime(value).toZone('UTC').asdatetime()
```
sneridagh: @tisto @buchi @lukasgraf 

I just tried a new approach and change the code to store the datetime objects as datetime offset-aware and referenced always to UTC. It felt bad to convert it to the local time of the server each time, and it caused a bad offset when the API waas used by people in other timezones than the server's one.

docs/source/serialization.rst
```diff
@@ -24,6 +24,10 @@ Python                                  JSON
 ``DateTime('2015/11/23 19:45:55')``     ``'2015-11-23T19:45:55'``
 ======================================= ======================================
 
+.. warning::
```
sneridagh: The only thing to keep in mind by using this approach is when used in combination with zope.schema date validators. If that's the case, you should also use a datetime object that also contains offset-aware object as the validator value.

tisto:
@thet any chance you have time to have a look at this?

src/plone/restapi/deserializer/dxfields.py
loechel: That is perfect, It should always be in an isoformat with UTC timezone.

src/plone/restapi/deserializer/dxcontent.py
```diff
@@ -78,7 +78,14 @@ def __call__(self, validate_all=False, data=None):  # noqa: ignore=C901
                             'message': e.doc(), 'field': name, 'error': e})
                     else:
                         field_data[name] = value
-                        if value != dm.get():
+                        try:
+                            if value != dm.get():
+                                dm.set(value)
+                                modified = True
+                        except TypeError:
```
loechel: I am sorry, but I did not like the usage of an Error Handler to apply a value

I would suggest something like:

    dm_value = dm.get()
    if value.tzinfo is None:
        value = value.replace(tzinfo=pytz.UTC)
    else:
         value = value.asdatetime(pytz.UTC)
     if dm_value.tzinfo is None:
        dm_value = dm_value.replace(tzinfo=pytz.UTC)
    else:
         dm_value = dm_value.asdatetime(pytz.UTC)

    if value != dm_value:
        dm.set(value)
        modified = True
 

instead of the try except clause.

sneridagh: Ok, but you have to detect first if dm.get() returns a datetime right? if not, value.tzinfo will fail. I wanted to prevent to infer more conditional clauses here, which in turn, increases the complexity ratio of the method (and then plone.recipe.codeanalysis complains about it).

loechel: yes you are right, there needs to be an ``if isinstance(value, datetime) and isinstance(dm_value, datetime):``  that wraps both if else clause.

loechel: Well the complexity of an aditional if level is better than an try except. It could also be an idea to split this test aus into a separate function.

docs/source/deserialization.rst
```diff
@@ -0,0 +1,87 @@
+Deserialization
+===============
+
+It's worth to note an special case when deserializing Datetimes objects, and how plone.restapi will handle them.
```
tisto: @sneridagh type "an" -> "a"

docs/source/deserialization.rst
```diff
@@ -0,0 +1,87 @@
...
+Although not supported by Plone itself yet, plone.restapi will store all the Datetimes that will be handling along with its timezone converted to UTC.
+This will provide a common ground for all the datetimes operations.
+
+There is an special case when using datetimes objects in p.a.event, and its behavior is different due to implementation differences for versions 1.x (Plone 4) and 2.x and above (Plone 5).
+
+.. warning::
+  In case of using zope.schema date validators you should also use a datetime object that also contains offset-aware object as the validator value.
+
+.. note::
+  This does not applies in case that you are using Plone 4 with no Dexterity support at all or not p.a.event installed.
```
tisto: @sneridagh type "applies" -> "apply" / better -> "or if you don't have p.a.event installed"

src/plone/restapi/deserializer/dxcontent.py
tisto: @sneridagh @loechel For me the main objective of this code snippet is readability. I don't mind a try/except if it makes the code more readable. Though, without a few more comments I don't really understand what goes on here anyways. :)

docs/source/deserialization.rst
```diff
@@ -0,0 +1,87 @@
...
+    "end": "2018-01-01T12:00:00+07:00",
+    "id": "myevent2",
+    "start": "2018-01-01T10:00:00+07:00",
+    "timezone": "Asia/Saigon",
+    "title": "My Event",
+    }
+
+and builds the `start` and `end` fields with the proper timezone, depending on the `timezone` field. It also returns the datetime object with the proper timezone offset.
+
+If using Plone 4 and p.a.event 1.x you should construct the Event type using this approach, otherwise the Event object will be created with a wrong timezones.
```
thet: typo: -> "with a wrong timezone."

sneridagh:
@loechel Please check again the current implementation. It takes into account your comments.

src/plone/restapi/deserializer/dxcontent.py
```diff
@@ -78,7 +81,20 @@ def __call__(self, validate_all=False, data=None):  # noqa: ignore=C901
...
+                        # This is required in case that we can compare
+                        # offset-naive and offset aware objects. We convert all
+                        # offset-naive datetimes to UTC timezone first and then
+                        # compare them
+                        if isinstance(value, datetime) and \
+                           isinstance(dm_value, datetime):
+                            if value.tzinfo is None:
+                                value = value.replace(tzinfo=pytz.UTC)
+                            if dm_value.tzinfo is None:
+                                dm_value = dm_value.replace(tzinfo=pytz.UTC)
```
thet: Wouldn't it be better to localize the datetime values to the portal's timezone?

Something like:
```
from plone.app.event.base import default_timezone
dtz = default_timezone(self.context, as_tzinfo=True)
dtz.localize(value)
```

Or if you have to avoid a dependency on plone.app.event here:
```
value = DateTime(
                value.year(),
                value.month(),
                value.day(),
                value.hour(),
                value.minute()
).asdatetime()
```
That uses the default timezone of DateTime, which is the servers timezone (which should actually be set to UTC but often is set in the local zone).

src/plone/restapi/deserializer/dxfields.py
```diff
@@ -45,9 +45,8 @@ class DatetimeFieldDeserializer(DefaultFieldDeserializer):
     def __call__(self, value):
         try:
             # Parse ISO 8601 string with Zope's DateTime module
-            # and convert to a timezone naive datetime in local time
-            value = DateTime(value).toZone(DateTime().localZone()).asdatetime(
-            ).replace(tzinfo=None)
+            # and convert to a UTC based timezone as datetime
+            value = DateTime(value).toZone('UTC').asdatetime()
```
thet: Again like above, I wouldn't use 'UTC' but the portal's default timezone.

sneridagh: I think that the concept of having a "portal" timezone is wrong. In which case is p.a.event using it? It's ok if you always have users in the same timezone, but this is hardly the case in most user cases.

All datetimes should be referenced using a common baseline. Then the UI convert it to the user's timezone. If not, a user in India will see always the dates offsetted with the portal timezone, which is pretty useless for him. This is what happened before and that's the reason of this change.

sneridagh: @thet I thought you agree on that too when you said: https://github.com/plone/plone.restapi/issues/253#issuecomment-324383766

docs/source/deserialization.rst
sneridagh: Fixed.

src/plone/restapi/deserializer/dxfields.py
thet: Yes, true. I was suggesting that because I thought you do not have a timezone information in the JSON object available - besides the offset.
But if you have timezone-naive datetimes, like in the code above - and you convert those to UTC, then you make a decision which could be wrong. Unless you always exchange dates in UTC.

Fact is, plone.app.event 2.x does not convert it's datetimes to UTC but stores them in the default timezone - which is:
1) The event's timezone (not implemented yet)
2) The user's timezone (probably most often not set)
3) The portal's timezone (You have to set this on portal creation time)

I agree that users from anywhere want to have the option to show the datetimes in their local zone. But that's a UI thing and should be done via JS with the option to show the datetimes in the original zone.

So: converting to UTC and storing in UTC is OK, if you have no timezone identifier (offsets are a bit useless) and exchange in UTC anyways.
IMO it's not OK if the datetimes are "naive" - without any timezone information.

thet: @sneridagh If you want, we can have a short hangout if you think I can help here...

sneridagh: @thet sure, whenever you are available, we can have a talk about it. We should find a common agreement on this.

Agreed that the implementation makes an assumption that p.a.event is not (yet) doing but IMHO should do and we should fix it upstream in Plone. That is, save the internal dates with timezone information under a common frame of reference (say UTC).

Clients (and existing UI) should provide always datetimes depending on the user's timezone (we are already doing that in a client's project all Angular UI. Current libraries always transform to UTC on serializing and it's easy to be user's timezone aware.

Setting a portal default timezone (or even a user timezone) it's useless and error prone. The current implementation now it's broken, as you can see, the provided tests are not passing using the current version of plone.restapi. 

The more sensible solution (as agreed in #253 seems to store always the timezone information) but then, use a common reference for that (using server's timezone will also be error prone and dates could not be the same if you migrate one server from timezone).

@tisto @lukasgraf @buchi any insights on this?

loechel: I would say, if the convention in plone.restapi is to always export dates as UTC that is an convention. If other modules for other export formats uses another convetion that might not be consistent, but possible.

I would prefer to always export datetime objects as timezone aware objects, and that transformed to UTC, but also to always import all native datetime input as new timezone aware datetime objects. Presentation in the UI should be bound to user settings.

src/plone/restapi/deserializer/dxcontent.py
loechel: good catch that the second case in my proposal is unnessesary, two datetime objects that have the same time in utc are equal even if the contain different timezone infos.

loechel:
@tisto could you please make a last check and merge, and also close #396 

Thanks

buchi:
This doesn't work for me. I'v done the following test with Plone 5.0 with p.a.event 2.0:

Create an event TTW with:
```
start: 2017-10-01T10:00
end: 2017-10-31T10:00
effective: 2017-10-02T10:00
expires: 2017-10-30T10:00
```

Event information shown in event view:
```
01.10.2017 10:00 to 31.10.2017 10:00 (Europe/Berlin / UTC200)
Effective: 2 Oktober, 2017 10:00
Expires: 30 Oktober, 2017, 10:00
```

Create the "same" event with REST API:
```json
{
    "@type": "Event",
    "title": "A REST API event",
    "effective": "2017-10-02T10:00:00",
    "end": "2017-10-31T10:00:00+01:00",
    "expires": "2017-10-30T10:00:00",
    "start": "2017-10-01T10:00:00+02:00"
}
```

Event information shown in event view:
```
01.10.2017 08:00 to 31.10.2017 09:00 (UTC / UTC0)
Effective: 2 Oktober, 2017 10:00
Expires: 30 Oktober, 2017, 10:00
```

Both events represent the same dates but timezone information is wrong when the event is created using the REST API.

Editing the REST API event TTW by adding e description results in:
```
01.10.2017 08:00 to 31.10.2017 09:00 (Europe/Berlin / UTC200)
Effective: 2 Oktober, 2017 10:00
Expires: 30 Oktober, 2017, 10:00
```

Now the timezone information gets added, but no timezone conversion occurs. Thus the dates are now wrong. Even worse, they get modified though the user didn't change them.

sneridagh:
This is superseeded by #415 , closing.

Issue #150, opened 2016-10-22
types endpoint failing with AttributeError: lookupSchema


lukasgraf:
```
GET http://localhost:8080/Plone/@types/Document
```

```
2016-10-22 12:06:19 ERROR Zope.SiteErrorLog 1477152379.920.0579750084626 http://localhost:8080/Plone/@types/Document
Traceback (innermost last):
  Module ZPublisher.Publish, line 138, in publish
  Module ZPublisher.mapply, line 77, in mapply
  Module ZPublisher.Publish, line 48, in call_object
  Module plone.rest.service, line 20, in __call__
  Module plone.restapi.services, line 15, in render
  Module plone.restapi.services.types.get, line 50, in reply
  Module plone.restapi.types.utils, line 134, in get_jsonschema_for_portal_type
  Module plone.restapi.types.utils, line 108, in get_jsonschema_for_fti
  Module plone.restapi.types.utils, line 37, in get_ordered_fields
AttributeError: lookupSchema
```

(On both Plone 4.3.x and Plone 5.0.4)

lukasgraf:
Was caused by not having `plone.app.contenttypes` installed, so I had `DynamicViewTypeInformation`s instead of `DexterityFTI`s.

Apparently the `@types` endpoint only supports DX FTIs so far.

jaroel:
This is also an issue for the site root:
2017-04-28 15:55:37 ERROR Zope.SiteErrorLog 1493387737.640.088023899385 http://localhost:8080/Plone/@types/Plone%20Site
Traceback (innermost last):
  Module ZPublisher.Publish, line 138, in publish
  Module ZPublisher.mapply, line 77, in mapply
  Module ZPublisher.Publish, line 48, in call_object
  Module plone.rest.service, line 20, in __call__
  Module plone.restapi.services, line 19, in render
  Module plone.restapi.services.types.get, line 52, in reply
  Module plone.restapi.types.utils, line 169, in get_jsonschema_for_portal_type
  Module plone.restapi.types.utils, line 132, in get_jsonschema_for_fti
AttributeError: lookupSchema

PR #410, opened 2017-10-21
Make @types/[type] work for non-dx types

jaroel:
See also #150

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,9 @@ Changelog
 - Disallow None and u'' when TextLine is required. Refs #351.
   [jaroel]
 
+- Make getting '/@types/{type_id}' work for non-DX types, ie "Plone Site".
+  [jaroel]
+
 
 1.0a21 (2017-09-23)
 -------------------
--- a/src/plone/restapi/tests/test_types.py
+++ b/src/plone/restapi/tests/test_types.py
@@ -111,13 +111,8 @@ class TestJsonSchemaUtils(TestCase):
     def test_get_jsonschema_for_fti_non_dx(self):
         """Make sure FTIs without lookupSchema are supported.
         """
-
-        # We use this fake FTI because we need an instance with Title.
-        class FakeFTI(object):
-            def Title(self):
-                return 'Fake Type'
-
-        fti = FakeFTI()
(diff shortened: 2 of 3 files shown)


src/plone/restapi/types/utils.py
```diff
@@ -129,10 +129,16 @@ def get_jsonschema_for_fti(fti, context, request, excluded_fields=None):
...
         excluded_fields = []
 
-    schema = fti.lookupSchema()
-    additional_schemata = tuple(getAdditionalSchemata(portal_type=fti.id))
-
-    fieldsets = get_fieldsets(context, request, schema, additional_schemata)
+    try:
+        schema = fti.lookupSchema()
+    except AttributeError:
+        schema = None
```
davisagli: Maybe add a comment here to remind us when we expect this to happen.

jaroel:
Fair enough 👍

src/plone/restapi/types/utils.py
tisto: @jaroel +1 ^^^

CHANGES.rst
```diff
@@ -10,6 +10,9 @@ Changelog
 - Disallow None and u'' when TextLine is required. Refs #351.
   [jaroel]
 
+- Make getting '/@types/Plone Site' work.
```
tisto: @jaroel maybe "Make @types endpoint work on portal root"?

jaroel:
I've added some comments and a test for this.
Should be good?

src/plone/restapi/tests/test_types.py
```diff
@@ -108,6 +108,20 @@ def test_get_jsonschema_for_fti(self):
         )
         self.assertNotIn('title', jsonschema['properties'].keys())
 
+    def test_get_jsonschema_for_fti_non_dx(self):
```
tisto: @jaroel any reason why you didn't add a test here that just calls @types/Plone%20Site (I guess it kind of sucks that we use blanks in our type ids)

jaroel: No reason. Using the Plone Site FTI is a beter idea. I've changed the test to reflect that.

tisto:
Ready to merge if Travis is green.

Issue #127, opened 2016-07-15
Admin logged in with jwt token is not able to do all workflow transitions


pabo3000:
Steps to reproduce:
1. Create Plone site ( I used 4.3.9)
2. Install plone.restapi
3. Ask for possible transitions on front-page or any other doc using the rest api

```
GET front-page/@workflow HTTP/1.1
Accept: application/json
with Basic Auth   admin:admin
```

response:

```
{
  "history": [
    {
      "action": null,
      "actor": "admin",
      "comments": "",
      "review_state": "private",
      "time": "2016-07-15T12:52:31+02:00"
    },
    {
      "action": "publish",
      "actor": "admin",
      "comments": "",
      "review_state": "published",
      "time": "2016-07-15T12:52:31+02:00"
    }
  ],
  "transitions": [
    {
      "@id": "http://localhost:8080/test/front-page/@workflow/reject",
      "title": "Send back"
    },
    {
      "@id": "http://localhost:8080/test/front-page/@workflow/retract",
      "title": "Retract"
    }
  ]
}
```

Now ask the same but use the JWT. (Ask for it @login with Basic Auth admin:admin)
f.e.

```
Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJmdWxsbmFtZSI5bnVsbCwic3ViIjoiYWRtaW4iLCJleHAiOjE0Njg2MjMyNDN9.2iW5nJzHh0HgcY8CjRC5v0nnkSkTHAL-RUsL21fzdBY
```

the response is:

```
{
  "history": [
    {
      "action": null,
      "actor": "admin",
      "comments": "",
      "review_state": "private",
      "time": "2016-07-15T12:52:31+02:00"
    },
    {
      "action": "publish",
      "actor": "admin",
      "comments": "",
      "review_state": "published",
      "time": "2016-07-15T12:52:31+02:00"
    }
  ],
  "transitions": [
    {
      "@id": "http://localhost:8080/test/front-page/@workflow/retract",
      "title": "Retract"
    }
  ]
}
```

As you see you cannot reject despite you are admin.

tisto:
@buchi any idea how this is related to JWT?

buchi:
I guess this is because the JWT plugin is installed in the `acl_users` folder of the Plone site but the admin user is defined in the root `acl_users` folder. The user only gets roles from the `acl_users` folder he got authenticated in.
[...]

Issue #168, opened 2016-12-05
Permission inconsistency between authenticating with (Zope) admin using basic auth and token


sneridagh:
I guess it's not important, but maybe this hides something else that is. So here it goes.

I discovered that the use case when you auth with the admin **Zope** user via token, the user have a different set of permissions (not roles) than when it's authenticated via basic auth.

This make any sense for you? I haven't been able to reproduce it with a test. I stumbled into this when trying to change workflow of an item with the admin user via token in an app. It didn't allowed me, because in that context the user lack the Review content permission!

@buchi @lukasgraf @timo any insights?

buchi:
Seems to be the same issue as in #127.

The problem is that the admin user is defined in the root acl_users folder but the JWT plugin is installed in the acl_users folder of the Plone site. The user only gets roles from the acl_users folder he got authenticated in. When authenticated by the JWT plugin, the admin user doesn't get any roles from the root acl_users folder.

I think we should remove the support for authenticating users that are not defined in the same acl_users folder as the JWT plugin because it doesn't work properly. This also means that you will no longer be able to authenticate as admin user with a JWT token. A workaround would be to install the JWT plugin in the root acl_users folder. However the current implementation relies on plone.keyring which doesn't work outside of a Plone site.

tomgross:
I have seen a security issue in collective.solr which is probably related. https://github.com/collective/collective.solr/issues/150

tisto:
Will close this issue for https://github.com/plone/plone.restapi/issues/178

PR #412, opened 2017-10-21
Fix handling of Zope root users

### LINKED ISSUE #178: Dual authentication (from cookie and token) issues in production environments
Given an app completelly independent from Plone (and not served by Plone), that shares the same server name, but served from another path. The app is using token authentication from the plone.restapi PAS plugin.

We are experiencing issues since the cookies are leaking from Plone (if you are authenticated) to the other app. This causes authentication inconsistencies when querying plone.restapi endpoints, for example, when logging out the app (you are still logged in Plone), or if the token expires (most likely before the cookie does).

Could we find a way to mark how the authentication has been granted? Something like marking the response with a header (that later, could be checked by the app), let's say: X-Auth-Via: cookie/token ? Or just other approach to prevent this happen?

/cc @bloodbare @tisto @buchi @lukasgraf

buchi:
Another approach to fix #178, #127, #168 
Obsoletes #378 

- The `@login` endpoints now checks for an installed JWT plugin in the acl_users folder where the logging in user has been found.
- The JWT plugin is also installed in the Zope root acl_users folder.
- Each JWT plugin uses a unique signing secret to prevent authenticating tokens issued by other JWT plugins.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,11 @@ Changelog
 1.0a23 (unreleased)
 -------------------
 
-- Nothing changed yet.
+Bugfixes:
+
+- Fix JWT authentication for users defined in the Zope root user folder.
+  This fixes https://github.com/plone/plone.restapi/issues/168 and
+  https://github.com/plone/plone.restapi/issues/127.
 
 
 1.0a22 (2017-11-04)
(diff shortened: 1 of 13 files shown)


buchi:
@tisto @sneridagh This one is ready for a final review. Can you check if it fixes your issues. I'm pretty sure it fixes #127 and #168, but I'm not sure I understand #178 correctly.

tisto:
@buchi thank you! We will look into it during this week.

Issue #242, opened 2017-03-05
Add interfaces in content GET response


ebrehault:
We need an `interfaces` attribute which will contain a list of strings corresponding to the interfaces implemented by the content object.

It is needed for view matching in client-side traversal.

Note: It might not be by default (a expand could work).

erral:
This is also needed as we saw on the Angular training at PloneConf 2017, to get the global navigation correctly in contexts where a navigation root interface is added (subsites, ...)

erral:
I have started a branch implementing an `@interfaces` endpoint which is expandable. See the pull-request for more information.

PR #414, opened 2017-10-22
Issue 242: add an @interfaces expandable endpoint

erral:
See #242

### DIFF
--- a/docs/source/interfaces.rst
+++ b/docs/source/interfaces.rst
@@ -1,0 +1,22 @@
+Interfaces
+==========
+
+*Interfaces are objects that specify the external behavior of objects
+that "provide" them.* (from `zope.interface documentation`_)
+
+All objects in Plone have one or more interface which they provide and in some cases knowing those interfaces
+is interesting to act accordingly. For instance many views are registered to be available just for objects
+providing certain interfaces, or navigation and searches are restricted to the contents of objects providing
+the *INavigationRoot* interface.
+
+To get the interfaces provided by an object context we need to issue a ``GET`` request:
+
+...  http:example:: curl httpie python-requests
+    :request: _json/interfaces.req
+
+The response will include all provided interfaces:
+
+... literalinclude:: _json/interfaces.resp
+   :language: http
+
+.. _`zope.interface documentation`: https://docs.zope.org/zope.interface/README.html
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,8 @@ Changelog
 3.4.6 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Add an @interfaces endpoint to return the interfaces provided by the context object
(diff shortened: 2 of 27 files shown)


buchi:
👍 LGTM

I wonder if exposing all those interface names publicly may lead to unwanted information disclosure.

erral:
Yeah... during the training we wondered if we could restrict somehow the list of exposed interfaces, or at least set a list of known-good-set of interfaces, and expose just them...

docs/source/interfaces.rst
```diff
@@ -0,0 +1,12 @@
+Interfaces
+==========
+
+Getting the interfaces provided by the current context:
```
tisto: @erral we usually try to add a short introduction to non-Plone devs what this documentation section is about. E.g. a short introduction what interfaces are, maybe with a link to the Plone docs.

tisto:
@erral LGTM as well. Though, as @buchi said, the interfaces are currently not visible by anonymous users. Therefore we might have to think about this...

erral:
I share your concerns about the public visibility of the interfaces. Nevertheless they are needed, at least to know the navigation contexts when the front-end wants to build the global navigation for instance.

@ebrehault any comment on this?

tkimnguyen:
What would you recommend as a way of restricting who has access to this?

jensens:
So, what is the problem with exposing? 

Does exposing information about interfaces - and so about installed add-ons and other internals - give attackers hints about possible vulnerabilities outside of the core? But also FTI info and other endpoint expose that information. 

I doubt it a problem, but I would love to hear opposing opinions.

mauritsvanrees:
In PloneHotfix20161129 we removed the docstrings of lots of zope.interface methods and attributes to avoid 'publishing' them, that is making them available for browsers to call. See https://github.com/zopefoundation/Zope/pull/81. In some cases they could really be used for potential harm (setting values), in others it was: better safe than sorry.

Just giving the name of interfaces *should* be fine. But it *does* give you info about an object that is useful for a UI but also for potential hackers.
So a way to whitelist some interfaces, or to completely disable this (return an empty list) for the security conscious would be good.

jensens:
What is the state here? Do we want it or not? If not and hence there was no activity for a long time, I propose to close this PR within next two weeks. If you do not feel OK with this, please speak up and provide us a rough schedule.

tisto:
@jensens it is something we definitely want. Though, there are lots of things to consider and to discuss. I'd prefer to keep it open. Maybe something that can be discussed at one of the next sprints this year...

jensens:
This one got forgotten. Is it still valid or meanwhile superseded?

erral:
I am closing this because the main reason for this, the navigation root thing, has been fixed using another endpoint. If we need to expose the interfaces in the future we can use this code as an inspiration, but we do not need this right now.

PR #415, opened 2017-10-22
Fix datetime deserialization for timezone aware fields

### LINKED ISSUE #253: p.a.event: Cannot compare offset-naive and offset-aware datetimes

buchi:
Fixes #253

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,11 @@ Changelog
 1.0a23 (unreleased)
 -------------------
 
-- Nothing changed yet.
+Bugfixes:
+
+- Fix datetime deserialization for timezone aware fields.
+  This fixes https://github.com/plone/plone.restapi/issues/253
+  [buchi]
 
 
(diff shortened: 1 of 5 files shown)


buchi:
@sneridagh I've implemented another approach to fix datetime deserialization. Can you check if this works for your use case?

If the field already has a value, the new value will be converted to a timezone aware datetime object with the same timezone. This works also for newly created events as the start and end field both have default values.
If the field doesn't have a value, the given value is stored as datetime naive object (same behavior as before).

tisto:
@sneridagh could you please review this pr? I would like to make a new release soon and include this one if possible.

Issue #293, opened 2017-03-23
Undeclared dependency on Products.PasswordResetTool


tomgross:
There is a undeclared dependency on Products.PasswordResetTool in plone.restapi 1.0a10.

Products.PasswordResetTool is gone in Plone 5.1

see plone.restapi-1.0a10-py2.7.egg/plone/restapi/services/users/add.py
from Products.PasswordResetTool.PasswordResetTool import ExpiredRequestError

erral:
We are removing some other undeclared dependencies for Plone 4.3 at #292 we can handle this there too.

tisto:
@tomgross thanks! I am wondering if there is a tool that could detect undeclared dependencies like that...

lukasgraf:
@tisto [`z3c.dependencychecker`](https://github.com/reinout/z3c.dependencychecker) will report undeclared dependencies like that. It can result in some false positives though, so it's not necessarily suited to be wire up in a strict CI check, but I've found it to be very helpful indeed to run manually on occasion, e.g. before doing a release.

tisto:
Fixes: 

Undeclared dep: https://github.com/plone/plone.restapi/pull/295
Dep check: https://github.com/plone/plone.restapi/pull/296

tisto:
Plone 5.1 buildout and travis conf: https://github.com/plone/plone.restapi/pull/297

tisto:
Fixed in https://github.com/plone/plone.restapi/pull/295

PR #416, opened 2017-10-23
PasswordResetTool is optional

tomgross:


tomgross:
There seems to be an issue with travis-CI.

tisto:
Restarted Travis: https://travis-ci.org/plone/plone.restapi/jobs/291436360

tisto:
@tomgross if we make P.Pwresettool a soft-dependency, integrators will run into issues at run time (because the feature does not work) instead of at build time. Not sure if a hard-dependency isn't the better choice here. 

@sneridagh @csenger @jaroel @buchi @lukasgraf: opinions?

erral:
FWIW on #293 we had to add this dependency because it wasn't included on Plone 5.1, has the situation changed?

jaroel:
This is just fine. Products.PasswordResetTool is a dependency for Products.CMFPlone < 5.1, so no runtime errors for integrators.

See https://github.com/plone/Products.CMFPlone/commit/fbbc911e9e8566e33d7bc17ef1a0b4bb6e6a4aad#diff-2eeaed663bd0d25b7e608891384b7298

tomgross:
The only place Products.PasswordResetTool is used is here: https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/services/users/add.py#L25
And this works for all supported Plone versions.

 * 4.3 (PasswordResetTool is a hard dependency of CMFPlone itself)
 * 5.0 (PasswordResetTool is a hard dependency of CMFPlone itself)
 * 5.1 (PasswordResetTool moved to CMFPlone and needed componets are imported from there)

Integrators will not run into any troubles with this change.

PR #418, opened 2017-10-28
Include title and description in site serialization.

buchi:


tisto:
@buchi FYI: I removed the site description in Plone 5 since it was not used anywhere in the code: 

https://github.com/plone/Products.CMFPlone/issues/139

I am not opposed merging this. Though, it might be misleading for users on Plone 5.

BTW: Travis failure is something we ran into with our other Plone projects already. Will look into it...

tisto:
@buchi buildout is fixed in master. Rebase and Travis should be green.

buchi:
@tisto Thanks for your feedback. I wasn't aware that the site description isn't used anywhere. Thus I will remove it here. We just need the title.

PR #428, opened 2017-11-08
Pass context to richtext.output_relative_to

### LINKED ISSUE #424: IRichTextValue serializer uses value.output instead of value.output_relative_to
See https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/serializer/converters.py#L139

Also see https://github.com/plone/plone.app.textfield/issues/24

jaroel:
Fixes #424
This changes the serialisers to pass in the correct context.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,9 @@ Changelog
 1.0b1 (unreleased)
 ------------------
 
-- Nothing changed yet.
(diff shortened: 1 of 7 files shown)


sunew:
Another option could perhaps be to create a serializer in dxfields.py - just like ImageFieldSerializer

jaroel:
Yeah, this should be more like it!

PR #429, opened 2017-11-09
Add 'is_editable' and 'is_deletable' to the serialization of comments…

sneridagh:
… objects. Also refactored the comments endpoint to DRY.

### DIFF
--- a/docs/source/_json/comments_get.resp
+++ b/docs/source/_json/comments_get.resp
@@ -13,6 +13,8 @@ Content-Type: application/json
       "comment_id": "1477076400000000", 
       "creation_date": "2016-10-21T19:00:00", 
       "in_reply_to": null, 
+      "is_deletable": true, 
+      "is_editable": true, 
       "modification_date": "2016-10-21T19:00:00", 
       "text": {
         "data": "Comment 1", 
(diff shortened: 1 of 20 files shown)


docs/source/_json/comments_get.resp
```diff
@@ -13,6 +13,8 @@ Content-Type: application/json
       "comment_id": "1477076400000000", 
       "creation_date": "2016-10-21T19:00:00", 
       "in_reply_to": null, 
+      "is_deletable": true,
```
tisto: @sneridagh we decided to just "map over" existing attributes instead of cleaning them up. Are those attributes present in the same way on the discussion object?

src/plone/restapi/serializer/discussion.py
```diff
@@ -1,12 +1,18 @@
...
-from plone.app.discussion.interfaces import IConversation
 from plone.app.discussion.interfaces import IComment
+from plone.app.discussion.interfaces import IConversation
 from plone.restapi.batching import HypermediaBatch
-from zope.publisher.interfaces import IRequest
-from plone.restapi.interfaces import ISerializeToJson
 from plone.restapi.interfaces import IJsonCompatible
-from zope.component import adapter, getMultiAdapter
+from plone.restapi.interfaces import ISerializeToJson
+from plone.restapi.services.discussion.utils import can_delete
```
tisto: @sneridagh I guess this is a matter of taste but I usually prefer to keep the stuff in one file. No other class is using the methods in utils, so we do not need to factor them out. Therefore the dev has to open two files instead of just one.

docs/source/_json/comments_get.resp
tisto: Had a discussion with @sneridagh on that and he argued we are mapping the attribute of the discussion class to plone.restapi, not an attribute of the object. Therefore we do not have to be too strict about our mapping-over-policy here. That somehow convinced me.

PR #442, opened 2017-11-22
Fix fullobjects content get

sneridagh:


src/plone/restapi/serializer/catalog.py
```diff
@@ -84,7 +84,8 @@ def __call__(self, metadata_fields=(), fullobjects=False):
         for brain in batch:
             if fullobjects:
                 result = getMultiAdapter(
-                    (brain.getObject(), self.request), ISerializeToJson)()
+                    (brain.getObject(), self.request), ISerializeToJson)(
+                        include_items=False)
```
tisto: @sneridagh can we rename "include_items" to "return_items_as_fullobjects"?

PR #444, opened 2017-11-23
Ensure the userid returned by `authenticateCredentials` is a byte string.

### LINKED ISSUE #443: User not in their groups when doing a request (Plone 5.1rc1)
Hi,

I started a new project recently with p.restapi and Plone 5.1rc1. But I have a really bad bug right now.

I authenticate my user with JWT. Everything works fine except for the groups of the user. The groups I get for the user is empty (except for `AuthenticatedUser`).

Does anyone have any idea what could be going on?

Thanks.

Gagaro:
fix #443

### DIFF
--- a/src/plone/restapi/pas/plugin.py
+++ b/src/plone/restapi/pas/plugin.py
@@ -117,10 +117,7 @@ class JWTAuthenticationPlugin(BasePlugin):
         if 'sub' not in payload:
             return None
 
-        userid = payload['sub']
-
-        if isinstance(userid, unicode):
-            userid = userid.encode('utf8')
+        userid = payload['sub'].encode('utf8')
 
         if self.store_tokens:
             if userid not in self._tokens:
--- a/src/plone/restapi/pas/plugin.py
+++ b/src/plone/restapi/pas/plugin.py
@@ -119,6 +119,9 @@ class JWTAuthenticationPlugin(BasePlugin):
 
         userid = payload['sub']
 
+        if isinstance(userid, unicode):
+            userid = userid.encode('utf8')
+
         if self.store_tokens:
             if userid not in self._tokens:
                 return None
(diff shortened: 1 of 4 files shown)


tisto:
@Gagaro thanks for the fix! Could you imagine trying to write a unit test to get those lines covered?

Gagaro:
Test added

src/plone/restapi/pas/plugin.py
```diff
@@ -119,6 +119,9 @@ def authenticateCredentials(self, credentials):
 
         userid = payload['sub']
 
+        if isinstance(userid, unicode):
```
buchi: I would omit that check for unicode as the decoded token always contains unicode strings. Or can you provide a test for the code path where the if statement evaluates to false?

tisto: @Gagaro ^^^ :)

Gagaro: I answered but it was stuck on pending, not sure what github changed this time..

`payload` comes from a method which could be overwritten. I wanted to be sure that I'm not trying to encode an already encoded string.

buchi: @Gagaro `payload` comes from a private (underscore prefixed) method `_decode_token()`. It's pretty safe to assume that this method doesn't get overwritten. Even if the method changes in the future and for some reason returns byte strings then tests will fail. I prefer to have those kind of checks in tests instead of having an unreachable code path. Just my two cents.

Gagaro: All right then, I removed the check.

tisto:
@buchi pls have a second look and merge if you are ok with it.

buchi:
👍 LGTM

Gagaro:
Thanks!

PR #447, opened 2017-11-27
Allow properties in user add

jaroel:


csenger:
Looks good to me.

tisto:
@jaroel please check if applying roles is guarded, so an anonymous user can not just set any roles she would want.

jaroel:
@tisto Users cannot set their roles unless they have 'plone.app.controlpanel.UsersAndGroups'.
I've added a test to show this for anonymous users.

PR #449, opened 2017-11-28
Return the patched object in a patch response.

Gagaro:
To avoid having to do a GET request after a PATCH one.


tisto:
@Gagaro maybe I am mistaken. Though, this is not a breaking change in my opinion. We are adding something that should not break existing client code. Removing or amending existing calls would.

Gagaro:
I put that in the breaking change section because the behavior change. Some client might actually test the return code which is now `200` instead of `204`.

I just wanted to be sure. If you think this is all right though, that's fine for me too.

buchi:
I would like this to be optional. There are use cases where the object is not needed in the response and serializing the object on every PATCH request makes the API slower.

I would suggest to use the `Prefer` header for this: https://www.greenbytes.de/tech/webdav/rfc7240.html#return
The default should be no serialization.

Gagaro:
How do you generate the `.resp` for the documentation? I added a `.req` for the patch request with the representation but it didn't generate the resp.

lukasgraf:
@Gagaro you have `plone.restapi` write the `.req` and `.resp` for you, by creating/modifying a test in `test_documentation.py`, and using `save_request_and_response_for_docs()`. You should probably look into whether this test needs to be updated/extended: https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/tests/test_documentation.py#L233-L239

I would also prefer this behavior to be optional. The `Prefer` header sounds like a perfect fit for this.

Gagaro:
I agree, I was just waiting for your answer on how to update the doc to commit that.

Thanks.

docs/source/_json/content_patch_representation.req
```diff
@@ -0,0 +1,8 @@
+PATCH /plone/folder/my-document HTTP/1.1
+Accept: application/json
+Authorization: Basic YWRtaW46c2VjcmV0
+Content-Type: application/json
```
buchi: The `Prefer` header is missing in the example. You need to add it to the list of request header keys https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/tests/test_documentation.py#L56

Gagaro: I missed that, thanks for the review.

tisto:
I would like to hear @ebrehault and @robgietema opinion on this issue. I would especially like to hear if the current way of not returning the actual object makes it hard to implement the standard behavior of Plone (show the item view after creation).

tisto:
The prefer header seems to be an elegant solution. Will look into it and get back to you...

ebrehault:
In the standard scenario we meet in a CMS context, we usually want to get the resulting object after a PATCH. Nevertheless I agree it is probably better to keep it optional.

tisto:
@Gagaro Travis build is failing due to outdated docs.

src/plone/restapi/tests/test_content_patch.py
```diff
@@ -44,6 +44,18 @@ def test_patch_document(self):
         transaction.begin()
         self.assertEqual("Patched Document", self.portal.doc1.Title())
 
+    def test_patch_document_with_representation(self):
```
tisto: @Gagaro would it be possible to add a few more tests here? What happens if you use "return=minimal", "return=", "", etc. Just to make sure we have all options covered and documented.

Gagaro: Sorry, I was on holiday so I didn't see your message. I will be next week too, I don't know when I will be able to do that.

PR #454, opened 2017-11-30
Fix NamedFieldDeserializer to take null correctly and validate required correctly

jaroel:


tisto:
@sneridagh could you please double check if removing this test is fine with you? https://github.com/plone/plone.restapi/pull/454/commits/ca39148ece50edcc5665854ac05d7f597476db53

PR #455, opened 2017-11-30
Fix workflow translations with unicode characters

Gagaro:
I added a workflow in the testing profile (`test_workflow`) which is the `simple_publication_workflow` but with a unicode character in every title. I set it as the workflow for the custom testing types.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,9 @@ Changelog
 - Build documentation on Plone 5.0.x (before: Plone 4.3.x).
   [timo]
 
+- Fix workflow translations with unicode characters.
+  [Gagaro]
+
 
 1.0b1 (2018-01-05)
 ------------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,9 @@ Changelog
 - Fix workflow translations with unicode characters.
   [Gagaro]
 
+- Fix workflow encoding in transition endpoint
+  [Gagaro]
+
 
 1.0b1 (2018-01-05)
 ------------------
--- a/docs/source/_json/workflow_post.resp
+++ b/docs/source/_json/workflow_post.resp
@@ -7,5 +7,5 @@ Content-Type: application/json
   "comments": "", 
   "review_state": "published", 
   "time": "2016-10-21T19:05:00+00:00", 
-  "title": "Published"
+  "title": "Published with accent \u00e9"
 }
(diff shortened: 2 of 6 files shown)


tisto:
@Gagaro would it be possible to get away without adding a full custom workflow to the testing profile? For instance just amending an existing workflow via Python code?

I really dislike testing profiles (not your fault of course) because they represent a global state in the test that is hard to maintain and easy to break. I wanted to get rid of the testing profile all together for quite some time and merging this PR would make that a lot harder...

Gagaro:
The thing I like by adding the workflow in the profile is that the unicode character will be tested everywhere instead of just one test.

I'll see how I can change a specific workflow to have that in a single test if you prefer though.

tisto:
@Gagaro right, I forgot about the change in the HTTP status. Then it's actually a breaking change.

tisto:
@Gagaro true. Though, don't you think we are testing special characters on the workflow on the wrong level then? Unit/integration tests in plone.restapi should only test plone.restapi functionality, not Plone functionality.

Gagaro:
This bug is in plone.restapi. It could be somewhere else than `WorkflowInfo` and we can't be sure unless we test it everywhere.

tisto:
@Gagaro what about just amending the default workflow used in the test fixture to use a special character?  This would allow us to reduce the full testing profile with two or three lines of Python code which would make the test fixture a lot more readable and maintainable. What do you think?

Gagaro:
I did that and rebased on master.

tisto:
@Gagaro awesome! Thank you!

tdesvenain:
Hi i also have to include a commit that fixes the @transition endpoint https://github.com/plone/plone.restapi/commit/b1c035e8ff39592a6c4b61f1992ae635b86b6ade

tisto:
@tdesvenain now I am confused. Shall I wait with merging this one?

tdesvenain:
@tisto yes, sorry for the confusion
and this was this commit : https://github.com/makinacorpus/plone.restapi/commit/88ca28bf2c467f0f4167ef45c1e1958415dd910c

Gagaro:
Yes, I just added the fix + the updated test.

tisto:
Ok. Cool. Then you might want to update the changelog.

CHANGES.rst
```diff
@@ -7,6 +7,10 @@ Changelog
 - Build documentation on Plone 5.0.x (before: Plone 4.3.x).
   [timo]
 
+- Fix workflow translations with unicode characters.
```
tisto: @Gagaro this one is missing an author.

Gagaro: It's the same for both. I added it.

tisto:
@Gagaro Travis build fails. Please fix code violations.

tisto:
@Gagaro ping :)

Gagaro:
Sorry for the delay, it's all green now.

tisto:
Thanks a lot @Gagaro! It took a while but I think the result was worth the effort. :)

PR #456, opened 2017-11-30
Batching: Preserve list-like query string params when canonicalizing URLs

### LINKED ISSUE #450: Batching with multiple metadata fields generates an uncomplete result
If I want to query **multiple metadata_fields**, for example _Title_ and _Description_ while also batching the results, the _batching.first_ that is returned (same for _last_ and _next_) will contain only one of the _metadata fields_ present in the query (the last one, in this case _Description_ ).

To be more specific, a query such as:

```
curl 'http://localhost:8080/Plone/@search?b_size=1&b_start=0&metadata_fields=Title&metadata_fields=Description' -H 'Accept: application/json' --user admin:admin
```

will return for _batching_ a result such as:

```
"batching": {
    "@id": "http://localhost:8080/Plone/@search?b_size=1&b_start=0&metadata_fields=Title&metadata_fields=Description",
    "first": "http://localhost:8080/Plone/@search?b_size=1&b_start=0&metadata_fields=Description",
    "last": "http://localhost:8080/Plone/@search?b_size=1&b_start=16419&metadata_fields=Description",
    "next": "http://localhost:8080/Plone/@search?b_size=1&b_start=1&metadata_fields=Description"
  }
```

In the result listed above, it can be seen that the batching **first, last and next** contain only the _Description_ metadata field.

lukasgraf:
Query string parameters can be **repeated with the same key**, but different values, which is a de-facto standard for passing lists of values in query strings.

These got mangled when creating canonical batch URLs (and other URLs derived from the canonical batch URL, like `next` and `first` links) because we turned the result of `parse_qsl` into a dict.

This change makes sure these list-like query string parameters are preserved.

As a side effect, I also had to modify the order the batching parameters are added to the URLs slightly, but because they're now ordered that should make the URLs more stable overall and tests less brittle for the future.

Fixes #450

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@ Changelog
 1.0b1 (unreleased)
 ------------------
 
+- Batching: Preserve list-like query string params when canonicalizing URLs.
+  [lgraf]
+
 - Allow properties when adding a user.
   This allows setting the fullname by anonymous users.
   [jaroel]
--- a/docs/source/_json/batching.resp
+++ b/docs/source/_json/batching.resp
@@ -5,9 +5,9 @@ Content-Type: application/json
   "@id": "http://localhost:55001/plone/folder/@search", 
   "batching": {
     "@id": "http://localhost:55001/plone/folder/@search?b_size=5&sort_on=path", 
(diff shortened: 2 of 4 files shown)


tisto:
Thank you @lukasgraf for the fix!

Issue #470, opened 2017-12-30
ZODB.POSException.ReadConflictError: '\x0e\x15\x8a\t\x83\xaf\x1e\xb5'


tisto:
Upgrading the plone.restapi buildout from Plone 5.1b3 to 5.1b4 leads to a ZODB.POSException when running the tests:

```
$ bin/test
Running plone.restapi.testing.PloneRestApiATLayer:Functional tests:
  Set up plone.testing.zca.LayerCleanup in 0.000 seconds.
  Set up plone.testing.z2.Startup in 0.422 seconds.
  Set up plone.app.testing.layers.PloneFixture in 20.745 seconds.
  Set up plone.restapi.testing.PloneRestApiATLayer in 2.362 seconds.
  Set up plone.testing.z2.ZServer in 0.502 seconds.
  Set up plone.restapi.testing.PloneRestApiATLayer:Functional in 0.000 seconds.
  Running:

  Ran 8 tests with 0 failures and 0 errors in 6.787 seconds.
Running plone.restapi.testing.PloneRestApiATLayer:Integration tests:
  Tear down plone.restapi.testing.PloneRestApiATLayer:Functional in 0.000 seconds.
  Tear down plone.testing.z2.ZServer in 5.176 seconds.
  Set up plone.restapi.testing.PloneRestApiATLayer:Integration in 0.000 seconds.
  Running:

  Ran 88 tests with 0 failures and 0 errors in 30.821 seconds.
Running plone.restapi.testing.PloneRestApiDXLayer:Functional tests:
  Tear down plone.restapi.testing.PloneRestApiATLayer:Integration in 0.000 seconds.
  Tear down plone.restapi.testing.PloneRestApiATLayer in 0.018 seconds.
  Set up plone.app.event.testing.PAEventLayer in 2.349 seconds.
  Set up plone.app.contenttypes.testing.PloneAppContenttypes in 2.918 seconds.
  Set up plone.restapi.testing.PloneRestApiDXLayer in 4.478 seconds.
  Set up plone.testing.z2.ZServer in 0.507 seconds.
  Set up plone.restapi.testing.PloneRestApiDXLayer:Functional in 0.000 seconds.
  Running:
    76/316 (24.1%)Traceback (most recent call last):
  File "bin/test", line 292, in <module>
    '--test-path', '/Users/timo/workspace/plone/plone.restapi/src',
[...]

PR #471, opened 2017-12-30
5.1b4

tisto:
Do not run test_documentation for Plone > 5.0. This is a workaround for https://github.com/plone/plone.restapi/issues/470

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,9 +24,9 @@ before_install:
   - bin/pip install -r requirements.txt
 install:
   - sed -ie "s#plone-x.x.x.cfg#plone-$PLONE_VERSION.cfg#" travis.cfg
+  - bin/python --version
   - bin/buildout -t 10 -c travis.cfg
 script:
-  - bin/python --version
   - bin/code-analysis
   - bin/sphinxbuilder
   - bin/test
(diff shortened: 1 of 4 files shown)


tisto:
@lukasgraf it seems freezegun does not work well with the latest ZODB related upgrades we did for Plone 5.1b4: 

https://github.com/plone/buildout.coredev/commit/b58ef87a14d20fb400e33e86c61c96a6fac97aab

(I wasn't really able to prove this yet though)

Not running test_documentation for Plone > 5.0 is just a workaround.

lukasgraf:
@tisto I only had time to take a cursory glance at this, but it wasn't obvious to me from the couple travis failures I looked at why `freezegun` would be responsible for them. Any way I can help?

If `freezegun` turns out to be a blocker, we could always incorporate [our approach from `ftw.testing` (freezer)](https://github.com/4teamwork/ftw.testing#freezing-datetime-now). That uses [`forbiddenfruit`](https://github.com/clarete/forbiddenfruit) though, which we require to patch builtin classes written in C - never had any issues with it, but I wouldn't feel comfortable including `forbiddenfruit` as anything other than a test-only dependency.

lukasgraf:
@tisto FYI: I just did a test on current `master` locally with removing the

```
        if PLONE_VERSION.base_version >= '5.1':
            self.skipTest('Do not run documentation tests for Plone 5')
```

skips from `test_documentation.py` - tests passed for me on Plone 5.x without a problem. I think the `freezegun` stuff might have been a fluke, and we can unskip those "tests".

tisto:
@lukasgraf create a pr and let give it a try.

PR #476, opened 2018-01-09
Allow each user to get his own information

### LINKED ISSUE #420: User can't access its own metadata
Currently `cmf.ManagePortal` is required on the `@users` endpoint to get all users (plain GET to `/@users`) but also to get the information of a given user (GET to `/@users/userid`). I think that each user should be allowed to access its own properties.

Opinions?

erral:
This fixes #420

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,11 @@ Changelog
 - Build documentation on Plone 5.0.x (before: Plone 4.3.x).
   [timo]
 
+New Features:
+
+- Allow users to get their own user information.
+  [erral]
+
 
 1.0b1 (2018-01-05)
 ------------------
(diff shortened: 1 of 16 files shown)


tisto:
@erral awesome! Thank you! Will try to find time to look into this. Other reviews are also welcome. This needs close examination because we potentially expose sensitive information when removing the manage permission.

erral:
Yes. That's why I tried to add the test for cases in which the endpoints are used by unauthorized users (either anonymous or logged-in users without proper permissions).

Perhaps @tdesvenain can have a look at it he was implementing something that depended on this on plone.restapi-angular (https://github.com/plone/plone.restapi-angular/issues/53)

erral:
I've just fixed the typos. Thx!

PR #482, opened 2018-01-24
Fix the use of fullobjects in Archetypes based sites @search

### LINKED ISSUE #481: @search endpoints fails when searching with fullobjects=True parameter in a site with Archetypes content
the "include_items" parameter was added to Dexterity object serializer, both for plain items and for folderish ones:

https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/serializer/dxcontent.py#L45
https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/serializer/dxcontent.py#L113

This was not the case in Archetypes, where only was added to folderish items:

https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/serializer/atcontent.py#L35
https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/serializer/atcontent.py#L89

This breaks ~Archetypes Collection~ search results serializer when calling with `?fullobjects`, because it calls the serializer with the include_items parameter and it fails with an exception:

https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/serializer/catalog.py#L88

```2018-01-24T08:57:21 ERROR Zope.SiteErrorLog 1516780641.880.499275435646 https://www.eibar.eus/eu/@search
Traceback (innermost last):
  Module ZPublisher.Publish, line 138, in publish
  Module ZPublisher.mapply, line 77, in mapply
  Module ZPublisher.Publish, line 48, in call_object
  Module plone.rest.service, line 23, in __call__
  Module plone.restapi.services.search.get, line 12, in reply
  Module plone.restapi.search.handler, line 65, in search
  Module plone.restapi.serializer.catalog, line 88, in __call__
TypeError: __call__() got an unexpected keyword argument 'include_items'
```

I will try to provide a failing test and a fix in a pull-request.

erral:
This fixes #481

### DIFF
--- a/src/plone/restapi/tests/test_serializer_catalog.py
+++ b/src/plone/restapi/tests/test_serializer_catalog.py
@@ -57,15 +57,44 @@ class TestCatalogSerializers(unittest.TestCase):
 
     def test_lazy_map_serialization_with_fullobjects(self):
         # Test serialization of a Products.ZCatalog.Lazy.LazyMap
-        lazy_map = self.catalog()
+        lazy_map = self.catalog(path='/plone/my-folder/my-document')
         results = getMultiAdapter(
             (lazy_map, self.request), ISerializeToJson)(fullobjects=True)
 
         self.assertDictContainsSubset({'@id': 'http://nohost'}, results)
-        self.assertDictContainsSubset({'items_total': 2}, results)
-        self.assertEqual(2, len(results['items']))
+        self.assertDictContainsSubset({'items_total': 1}, results)
+        self.assertEqual(1, len(results['items']))
         result_item = results['items'][0]
-        self.assertDictContainsSubset({'layout': 'listing_view'}, result_item)
+
+        self.assertDictContainsSubset({
+            '@id': 'http://nohost/plone/my-folder/my-document',
+            '@type': 'Document',
+            u'changeNote': u'',
+            u'contributors': [],
+            u'creators': [u'test_user_1_'],
+            u'description': u'',
+            u'effective': None,
+            u'exclude_from_nav': False,
+            u'expires': None,
+            'id': u'my-document',
+            'is_folderish': False,
+            u'language': u'',
+            'layout': 'document_view',
+            'parent': {
+                u'@id': u'http://nohost/plone/my-folder',
+                u'@type': u'Folder',
+                u'description': u'',
+                u'review_state': u'private',
+                u'title': u'My Folder'},
+            u'relatedItems': [],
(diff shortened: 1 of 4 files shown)


src/plone/restapi/tests/test_serializer_catalog.py
```diff
@@ -31,7 +31,7 @@ def setUp(self):
         # /plone/my-folder/my-document
         self.doc = createContentInContainer(
             self.folder, u'Document',
-            created=DateTime(2015, 12, 31, 23, 45),
+            creation_date=DateTime(2015, 12, 31, 23, 45),
```
erral: This requires an explanation:

In Dexterity the attribute name that saves the creation date is not 'created' but 'creation_date'. https://github.com/plone/plone.dexterity/blob/2.2.x/plone/dexterity/content.py#L508

If we leave the document creation code as is (with created), when the object is serialized to JSON in the test that uses fullobjects, we get an error saying that 'DateTime object is not callable.

```Error in test test_lazy_map_serialization_with_fullobjects (plone.restapi.tests.test_serializer_catalog.TestCatalogSerializers)
  File "/home/erral/buildout.python/parts/opt/lib/python2.7/unittest/case.py", line 329, in run
    testMethod()
  File "/home/erral/dev/plone.restapi/src/plone/restapi/tests/test_serializer_catalog.py", line 64, in test_lazy_map_serialization_with_fullobjects
    results = getMultiAdapter((lazy_map, self.request), ISerializeToJson)(fullobjects=True)
  File "/home/erral/dev/plone.restapi/src/plone/restapi/serializer/catalog.py", line 88, in __call__
    include_items=False)
  File "/home/erral/dev/plone.restapi/src/plone/restapi/serializer/dxcontent.py", line 58, in __call__
    'created': json_compatible(obj.created()),
TypeError: 'DateTime' object is not callable
```

erral: The datetime field story in Dexterity objects is quite a nightmare: there is an [effective_date and effective method on the base class](https://github.com/plone/plone.dexterity/blob/2.2.x/plone/dexterity/content.py#L516) but there is also an [effective field in the IPublication behavior](https://github.com/plone/plone.app.dexterity/blob/2.3.x/plone/app/dexterity/behaviors/metadata.py#L146), similar things happen with expiration_date and expires.

tisto:
@buchi @lukasgraf since you guys use Archetypes in plone.restapi I would like to hear your opinion on that.

Issue #285, opened 2017-03-22
Discrepancy in documentation for Authentication sometimes root used as plone sometimes Plone


envy7:
In http://plonerestapi.readthedocs.io/en/latest/authentication.html
 the first example is using **Plone** as root for Get request
![image](https://cloud.githubusercontent.com/assets/12188993/24186004/25654f1e-0efc-11e7-9369-1af5eda57ec8.png)
 and on the same page the example of authentication with token **plone** is used as root for Get request
![image](https://cloud.githubusercontent.com/assets/12188993/24186040/55337d92-0efc-11e7-9b42-0e409847ff36.png)

since it is mentioned here http://plonerestapi.readthedocs.io/en/latest/conventions.html#upper-vs-lowercase that we should use lowecase in URIs the first one should be changed to from Plone to plone

tisto:
@envy7 thanks for reporting! The difference between "Plone" and "plone" is actually that the default name for a Plone site is "Plone". The default Plone site name when running tests (and auto-generating documentation) is "plone". Not sure why plone.app.testing introduced this inconsistency. Though, we should be consistent. I will check if it is possible to make plone.app.testing use "Plone" uppercase as well. Otherwise I'd go with using the lower-case version "plone" everywhere.

lukasgraf:
This also bugs me. For the same reason, we do have some inconsistencies regarding `localhost:8080` vs. `localhost:55001` in some places.

In order to achieve consistency, one option would be to do some post-processing to requests/responses in  [`save_request_and_response_for_docs`](https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/tests/test_documentation.py#L54-L92) before writing them to disk.

ajayns:
Any updates on this issue on how to move forward in fixing it? @tisto @lukasgraf

tisto:
@ajayns as @lukasgraf wrote, we would need to post-process the responses to fix this. Look into the save_request_and_response_for_docs function.

shivaansh0610-LUFFY:
@tisto @stevepiercy 

I checked the current documentation in the repository (docs/source),
including authentication and usage examples, and I couldn’t find any
occurrence of `/Plone` being used as the API root.

All examples seem to consistently use lowercase paths now.

This issue might already be resolved and could likely be closed.

Thanks

stevepiercy:
[...]

PR #483, opened 2018-01-25
remove inconsistency in documentation #285

kakshay21:
see #285

### DIFF
--- a/docs/source/authentication.rst
+++ b/docs/source/authentication.rst
@@ -7,8 +7,8 @@ That means that any authentication method supported by an installed PAS Plugin s
 
 For example, to authenticate using HTTP basic auth, you'd set an ``Authorization`` header:
 
-.. literalinclude:: _json/siteroot.req
-   :language: http
+..  http:example:: curl httpie python-requests
+    :request: _json/siteroot.req
 
 HTTP client libraries usually contain helper functions to produce a proper ``Authorization`` header for you based on given credentials.
 
--- a/docs/source/authentication.rst
+++ b/docs/source/authentication.rst
@@ -7,11 +7,8 @@ That means that any authentication method supported by an installed PAS Plugin s
 
 For example, to authenticate using HTTP basic auth, you'd set an ``Authorization`` header:
 
-.. code-block:: http
-
-  GET /Plone HTTP/1.1
-  Authorization: Basic Zm9vYmFyOmZvb2Jhcgo=
-  Accept: application/json
+.. literalinclude:: _json/siteroot.req
+   :language: http
 
 HTTP client libraries usually contain helper functions to produce a proper ``Authorization`` header for you based on given credentials.
 
--- a/docs/source/authentication.rst
+++ b/docs/source/authentication.rst
@@ -12,24 +12,6 @@ For example, to authenticate using HTTP basic auth, you'd set an ``Authorization
 
 HTTP client libraries usually contain helper functions to produce a proper ``Authorization`` header for you based on given credentials.
 
-Using the ``requests`` library, you'd set up a session with basic auth like this:
-
-.. code-block:: python
-
-    import requests
-
-    session = requests.Session()
-    session.auth = ('username', 'password')
-    session.headers.update({'Accept': 'application/json'})
-
-    response = session.get(url)
-
-Or the same example using ``curl``:
-
-.. code-block:: bash
-
-    curl -u username:password -H 'Accept:application/json' $URL
-
 
 JSON Web Tokens (JWT)
 ---------------------
--- a/src/plone/restapi/tests/test_documentation.py
+++ b/src/plone/restapi/tests/test_documentation.py
@@ -133,7 +133,7 @@ def save_request_and_response_for_docs(name, response):
                             reqs[i] = reqs[i].replace('plone', 'Plone')
                     json_body['source'] = reqs
                 elif 'id' in json_body:
-                    reqs = json_body['source']
+                    reqs = json_body['id']
                     if isinstance(reqs, str) or isinstance(reqs, unicode):
                         reqs = reqs.replace('55001', '8080')
                         reqs = reqs.replace('plone', 'Plone')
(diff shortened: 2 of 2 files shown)


erral:
Here instead of using the `..literalinclude` I would use the `.. http:example`` one, to show how you authenticate using several tools (requests, httpie, curl, ...)

docs/source/authentication.rst
```diff
@@ -7,11 +7,8 @@ That means that any authentication method supported by an installed PAS Plugin s
 
 For example, to authenticate using HTTP basic auth, you'd set an ``Authorization`` header:
 
-.. code-block:: http
-
-  GET /Plone HTTP/1.1
-  Authorization: Basic Zm9vYmFyOmZvb2Jhcgo=
-  Accept: application/json
+.. literalinclude:: _json/siteroot.req
+   :language: http
```
erral: Here instead of using the `..literalinclude` I would use a `.. http:example` to show how you authenticate using several tools (requests, httpie, curl, ...)

kakshay21:
@erral is that right?

docs/source/authentication.rst
```diff
@@ -7,8 +7,8 @@ That means that any authentication method supported by an installed PAS Plugin s
 
 For example, to authenticate using HTTP basic auth, you'd set an ``Authorization`` header:
 
-.. literalinclude:: _json/siteroot.req
-   :language: http
+..  http:example:: curl httpie python-requests
+    :request: _json/siteroot.req
 
 HTTP client libraries usually contain helper functions to produce a proper ``Authorization`` header for you based on given credentials.
```
erral: I would remove also the following paragraphs until the JSON Web Tokens (JWT) section, because the explanation is already covered by the example.

erral:
And what do you think about other inconsistencies such as the ones mentioned on the original ticket? The ones about /Plone vs /plone or the use of 8080 vs 55001

docs/source/authentication.rst
kakshay21: @erral  I don't think that would harm anyone. besides, it's not even wrong information. I guess we should leave it like that... 🙂

kakshay21:
@erral  I'm checking for anymore occurrence of `/Plone`. If I find any I'll change it to `/plone`.
I've no idea about dealing with the use of 8080 vs 55001. @tisto

docs/source/authentication.rst
erral: I said that because when using the `http:example` tag and you build the documentation a nice tabbed interface is shown with all the examples with python requests, curl, httpie and plain http.

That's why we are using it all through the documentation: to write one HTTP request and render it using multiple renderers (python-requests, httpie, curl). 

If we keep the following paragraphs we are repeating the information already given by the `http-example`.

You can check how the built documentation looks like running `./bin/sphinxbuilder` in your buildout and then looking at the documents created on `docs/html`

erral:
I would say that the `localhost:8080/Plone` bits are parts that were manually written in the documentation. We need to find a way to create them like other parts (the ones with `localhost:55001/plone`) and render them using `http:example` and `literalinclude`.

With the default configuration Plone starts to listen on port 8080, and when you create a Plone site in the browser, the default id is `Plone`, so when plone developers write `localhost:8080/Plone` they are using the usual defaults for a Plone installation. That's the reason to have some `localhost:8080/Plone`. 

In any case (`Plone` or `plone`) what that means is the location of the root of the Plone site. Usually Plone is deployed behind a web server so the end user does not see the `Plone` or `plone` stuff on the URL. For instance [this is a Plone site](https://www.eibar.eus) but there is no `Plone` on the URL. Perhaps we can add some documentation as an introductory note explaining why in our examples we use `Plone` or `plone` and explain that that refers to the root of the Plone site. The same happens with the port, in the examples we use the development port 8080 or 55001, but in production they are almost never used, because you talk with the API directly. This bits are clear from a Plone developer's perspective but could be confusing for non-plone developers, and they are a definitive interesting audience for plone.restapi.

tisto:
I think we should use "localhost:8080/Plone" everywhere. There are two options to accomplish this:

1) Post-process the auto-generated results
2) Run the tests on port 8080 and with "Plone" as site id (here is how you can change the site id in plone.app.testing: https://stackoverflow.com/questions/13055860/changing-the-site-root-in-plone-app-testing-plone-4/44969256#44969256, adding "export ZSERVER_PORT=8080")

Though, with the second approach we risk collisions on local dev machines...

tisto:
I created a branch that runs the test on port 8080 with "Plone" as a site id:

https://github.com/plone/plone.restapi/pull/484

Though, now most tests fail because of the change. Post processing might be simpler with less side effects...

tisto:
I guess we could factor out the test_documentation tests and run them with a different setting (basically monkey patch port and plone id in the test setup method).

docs/source/authentication.rst
kakshay21: Thanks! @erral

kakshay21:
@tisto  I'll try with the approach 1 to avoid these circumstances, and see if that works. And will add installation guide in the README.rst to install it with buildouts as mentioned by @jaroel.
Later if time permits(mid-term exams 😓), I'll add Postman tutorial to avoid those beginners confusions.
sounds good?
 @tisto @erral @jaroel

kakshay21:
@tisto, @erral  tests are working fine, but why is it not changing the documentation?
Am I missing something?

kakshay21:
@tisto @erral  can you please check this?

docs/source/authentication.rst
```diff
@@ -7,32 +7,11 @@ That means that any authentication method supported by an installed PAS Plugin s
...
-
-  GET /Plone HTTP/1.1
-  Authorization: Basic Zm9vYmFyOmZvb2Jhcgo=
-  Accept: application/json
+..  http:example:: curl httpie python-requests
+    :request: _json/siteroot.req
 
 HTTP client libraries usually contain helper functions to produce a proper ``Authorization`` header for you based on given credentials.
 
-Using the ``requests`` library, you'd set up a session with basic auth like this:
```
tisto: @kakshay21 may I ask why you removed this section? (Sorry for not coming back to this sooner).

kakshay21: It's already covered.
I build the documentation to show you.
screen shot 2018-03-20 at 5 53 15 pm

kakshay21: @tisto any update on this?

PR #490, opened 2018-02-07
Expandable translations

erral:
- Make `@translations` endpoint expandable
- Modify `@translations` endpoint to return results in an attribute called `items` instead of `translations` as previously to behave like the other expandable endpoints.
- Document the upgrade path from the previous version

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,20 @@ Changelog
 1.1.1 (unreleased)
 ------------------
 
+Breaking Changes:
+
+- Make `@translations` endpoint expandable
+  [erral]
+
+- Rename the results attribute in `@translations` endpoint to be 'items'
+  [erral]
+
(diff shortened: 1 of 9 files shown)


tisto:
@erral awesome! FYI: I created a new 2.0.0 milestone and assigned it to this PR since this is definitely a breaking change.

erral:
sorry, i clicked by mistake :)

tisto:
@erral bear with me a little longer. I plan to do another 1.x release in the next days. After that, we can do a 2.0.0 release with your pr.

erral:
No problem, there is no need to hurry

tisto:
@erral would you mind adding a more detailed description of the breaking changes to our upgrade guide? 

https://github.com/plone/plone.restapi/blob/master/docs/source/upgrade-guide.rst

erral:
I updated the docs with an example of the changes.

PR #493, opened 2018-02-11
convert date/times to UTC when exporting, use python-datetime for parsing date strings

thet:
Most of the test failures for Plone 5.2 were related to missing initialization of ZServer, which is fixed but not yet merged here: https://github.com/plone/plone.testing/pull/45

Other Plone 5.2 failures were related to date/timezone related offsets which did not match (I do not yet understand why this failures only were apparent for Plone 5.2).

Investigating this, I found out that the date/times were not consistently converted to UTC when exporting. I recall some discussions about this including some contradictory statements including from myself in https://github.com/plone/plone.restapi/issues/253

However, I think since ISO8601 lacks of a concept of timezone names which are necessary to do date math when daylight saving time boundaries are included (timezone offsets do not allow to conclude to DST dates), the most practical thing would be to convert all date/times to UTC when exporting.

This PR addresses this:

- Convert all datetime, DateTime and time instances to UTC before serializing.
- Use python-dateutil instead of DateTime to parse date strings when de-serializing.

please comment, if this approach is going in the right direction!

### DIFF
--- a/src/plone/restapi/deserializer/dxfields.py
+++ b/src/plone/restapi/deserializer/dxfields.py
@@ -1,6 +1,4 @@
 # -*- coding: utf-8 -*-
-from DateTime import DateTime
-from DateTime.interfaces import DateTimeError
 from datetime import timedelta
 from plone.app.textfield.interfaces import IRichText
 from plone.app.textfield.value import RichTextValue
@@ -24,6 +22,8 @@ from zope.schema.interfaces import IFromUnicode
 from zope.schema.interfaces import ITextLine
 from zope.schema.interfaces import ITime
 from zope.schema.interfaces import ITimedelta
+
+import dateutil
 
 
 @implementer(IFieldDeserializer)
@@ -86,11 +86,11 @@ class DatetimeFieldDeserializer(DefaultFieldDeserializer):
             self.field.validate(value)
             return
 
-        # Parse ISO 8601 string with Zope's DateTime module
+        # Parse ISO 8601 string with dateutil
         try:
-            dt = DateTime(value).asdatetime()
-        except (SyntaxError, DateTimeError) as e:
-            raise ValueError(e.message)
+            dt = dateutil.parser.parse(value)
+        except ValueError:
+            raise ValueError(u'Invalid date: {}'.format(value))
 
         # Convert to TZ aware in UTC
         if dt.tzinfo is not None:
@@ -169,9 +169,10 @@ class TimeFieldDeserializer(DefaultFieldDeserializer):
             # Create an ISO 8601 datetime string and parse it with Zope's
             # DateTime module and then convert it to a timezone naive time
             # in local time
-            value = DateTime(u'2000-01-01T' + value).toZone(DateTime(
-            ).localZone()).asdatetime().replace(tzinfo=None).time()
-        except (SyntaxError, DateTimeError):
+            # TODO: should really a timezone naive time be returned?
+            # using ``timetz()`` would be timezone aware.
+            value = dateutil.parser.parse(value).time()
+        except ValueError:
             raise ValueError(u'Invalid time: {}'.format(value))
 
         self.field.validate(value)
--- a/src/plone/restapi/serializer/converters.py
+++ b/src/plone/restapi/serializer/converters.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
-from datetime import date
 from DateTime import DateTime
+from datetime import date
 from datetime import datetime
 from datetime import time
 from datetime import timedelta
@@ -8,10 +8,11 @@ from persistent.list import PersistentList
 from persistent.mapping import PersistentMapping
 from plone.app.textfield.interfaces import IRichTextValue
 from plone.dexterity.interfaces import IDexterityContent
-from plone.restapi.interfaces import IJsonCompatible
 from plone.restapi.interfaces import IContextawareJsonCompatible
+from plone.restapi.interfaces import IJsonCompatible
 from Products.CMFPlone.utils import safe_unicode
-from zope.component import adapter, queryMultiAdapter
+from zope.component import adapter
+from zope.component import queryMultiAdapter
 from zope.globalrequest import getRequest
 from zope.i18n import translate
 from zope.i18nmessageid.message import Message
@@ -19,6 +20,30 @@ from zope.interface import implementer
 from zope.interface import Interface
 
 import Missing
+import pytz
+# import re
+
+
+def datetimelike_to_iso(value):
+    if isinstance(value, DateTime):
+        value = value.asdatetime()
+
+    if getattr(value, 'tzinfo', None):
+        # timezone aware date/time objects are converted to UTC first.
+        utc = pytz.timezone('UTC')
+        value = value.astimezone(utc)
+    value.replace(microsecond=0)  # Remove microsecond part, we don't need it.
+    iso = value.isoformat()
+    # if value.tzinfo:
+    #     # Use "Z" instead of a timezone offset of "+00:00" to indicate UTC.
+    #     regex = None
+    #     if isinstance(value, datetime):
+    #         regex = re.compile(r'\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}')
+    #     if isinstance(value, time):
+    #         regex = re.compile(r'\d{2}:\d{2}:\d{2}')
+    #     match = regex.match(iso)
+    #     iso = match.group(0) + 'Z'
+    return iso
 
 
 def json_compatible(value, context=None):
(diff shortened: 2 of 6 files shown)


thet:
ping @tisto

tisto:
@sneridagh can you please have a look. I won't be able to look into it this week...

tisto:
@thet I don't see any changes in the docs. Does that mean that we only changed internal and the API stays the same? I am trying to understand if this is an internal bugfix or a breaking change.

tisto:
@buchi would you mind having a quick look here? I would like to hear your opinion as well since you also worked on this...

tisto:
@thet ping :)

thet:
@tisto IMO this is a breaking change, because as the following test shows, all date-times are converted to UTC before serializing:
https://github.com/plone/plone.restapi/pull/493/files#diff-42b8e56bfe820a154bfca3683e40731fR146

Is this change important? Yes, if date arithmetics should be done. E.g. for recurring events or calculating the duration of two date/times and ONLY if a DST change is in between (which we have to assume that there is).

Will a conversion to UTC fix DST change issues? Actually no. It just delegates the issue to the consumer of the serialized value. In case of Plone there are some hints how to handle it, because we (partly) support timezones since 5.0.

I'm not 100% confident about the relevance of this PR again.
It's a no-brainer in Python but a problem due to lack of timezone support in ISO8601 in JavaScript.
I feel the need to consult others about that. @tisto, let's talk in Berlin about this again.
Does @regebro has a opinion on that?
I'll try to also contact https://stackoverflow.com/users/634824/matt-johnson - he seems to be very deep into this topic.

Some more resources about this topic:

https://codeofmatt.com/content/images/best-dates-ever.pdf
https://codeofmatt.com/2015/02/07/what-is-a-time-zone/
https://codeofmatt.com/2015/06/17/javascript-date-parsing-changes-in-es6/
http://www.cl.cam.ac.uk/~mgk25/iso-time.html

tisto:
@thet @sneridagh @buchi I would like to release plone.restapi 2.0.0 soon. Any chance we can sort this out? As said, this goes over my head since I haven't worked with date/times lately. The pull request looks ok to me. I'd like to have a second or third opinion and approval on this before merging and releasing.

buchi:
This is definitely a breaking change!

First, let's recapitulate what the ideal solution for handling of dates and times would be:
- Always store in UTC (timezone naive)
- Always do calculations in UTC
- Return (GET) in UTC
- Accept (POST/PATCH) any TZ
- Display in local TZ

Unfortunately Plone does not store all dates and times in UTC. plone.app.event and Archetypes store them timezone aware. This is something that can't be fixed in plone.restapi.

So returning dates and times in UTC seems to be the way to go. But there are a few caveats because of not storing them in UTC:
- If you do a GET and then a PATCH with the data from the GET, you potentially modify the dates, which is not what one would expect.
- How do you get the timezone of an event (p.a.event)? The best approach would be to store the timezone in a separate field. But that's not the case now.

Btw, whats the reason for switching to python-dateutil for parsing of ISO8601 datetime strings? Anything that Zope's DateTime doesn't right?

thet:
@buchi I don't agree with the ideal solution:
- IMO storing date/times timezone aware is the right thing, as this a a integral information of any timezone aware date/time and also what humans expect. We had long discussion preceding plone.app.event 2.0, where we switched from UTC to timezoned date/times - it allowed for a simpler API when working with datetime attributes and it was a prerequisite to get rid of the ``data_postprocessor`` in form handling for those who remember that.
- IMO the lack of the concept of timezones in ISO8601 which only knows about timezone offsets is the real problem which we have to work around.

When de/serializing from/to JSON it would probably be good to store the IANA/Olson timezone identifier for each UTC-converted date. 

The switch to python-dateuti (which is a dependency of plone.app.event) l for parsing date strings is mostly to get rid of DateTime in the long run. It's not needed anymore in Python and it gets the timezones wrong as it only handles timezone offsets.

buchi:
@thet Storing dates/times in UTC has the following benefits:
- No ambiguity (e.g. 2017-10-29 02:30 in TZ Europe/Zurich is ambiguous, could be 2017-10-29 01:30 or 2017-10-29 00:30 in UTC)
- No invalid dates (e.g. 2018-03-25 2:30 in TZ Europe/Zurich is not a valid date)
- Calculating durations is simple
- Sorting is simple

IMHO timezones or local time is something that's needed only for input and output (display).
In most cases you want to display an event in your local timezone, not in a timezone defined by the event.

Also have a look at the documentation of pytz regarding local time:
"The preferred way of dealing with times is to always work in UTC, converting to localtime only when generating output to be read by humans."
https://pypi.python.org/pypi/pytz/2018.3/#problems-with-localtime

But let's get back to your PR. I'm fine if we merge it. But there are still the two issues I mentioned earlier:
- Plone stores the timezone for the start date of an event, but there's no way to get it using the REST API. I would suggest to add it to the serialization, e.g. `{ "start.timezone": "Europe/Zurich" }`. On the other side I would not use the timezone information anyway as we display dates in local time. Thus we can keep it as-is.
- There's a discrepancy between GET and POST/PATCH. That's something we have to live with and for me it's not a problem. @tisto ?

Regarding Zope's DateTime I don't believe that we will get rid of it that soon. But if python-dateutil is a dependency of Plone anyway, I don't mind.

tisto:
"If you do a GET and then a PATCH with the data from the GET, you potentially modify the dates, which is not what one would expect." That is indeed a problem because it breaks the REST assumptions, caching, etc.

Do we really modify content, when reading a content object in Plone? That sounds really wrong...

buchi:
@tisto I think you misunderstood that. The GET will not modify anything but the PATCH could despite I'm providing the same value as I got with GET. But I was wrong I guess. Due to #415 TZ info should be readded during deserialization, if the previous date had a TZ.

tisto:
@buchi ok, that's a relief. :)

sneridagh:
Please, also allow me to weigh in, IMO, I'm also +1 with @buchi ideal solution...

Handling UTC is always desirable. Libraries also expect somehow that, then convert it to local TZ (e.g momentjs).

What about the differences of implementation on p.a.e. for Plone 4/5? When I tried to fix this, I always got hit by this wall... 

@buchi which detail of your solution is then out with the current PR?

tisto:
@sneridagh @buchi @thet I would like to make a 2.0.0 release in the next days. This is the only issue that holds me back from merging all the open PRs with breaking changes and do a release. It would be really cool if we could figure that out in the next days.

src/plone/restapi/deserializer/dxfields.py
```diff
@@ -169,9 +169,10 @@ def __call__(self, value):
             # Create an ISO 8601 datetime string and parse it with Zope's
             # DateTime module and then convert it to a timezone naive time
             # in local time
-            value = DateTime(u'2000-01-01T' + value).toZone(DateTime(
-            ).localZone()).asdatetime().replace(tzinfo=None).time()
-        except (SyntaxError, DateTimeError):
+            # TODO: should really a timezone naive time be returned?
+            # using ``timetz()`` would be timezone aware.
+            value = dateutil.parser.parse(value).time()
```
thet: @buchi @sneridagh @tisto I'm not sure, if time should always converted to timezone-naive. Pythons datetime.time and thus zope.schema Time field supports the tzinfo objects.

src/plone/restapi/serializer/converters.py
```diff
@@ -1,24 +1,51 @@
...
+    if getattr(value, 'tzinfo', None):
+        # timezone aware date/time objects are converted to UTC first.
+        utc = pytz.timezone('UTC')
+        value = value.astimezone(utc)
+    if getattr(value, 'microsecond', False):
+        # Microseconds are normally not used in Plone
+        value = value.replace(microsecond=0)
+    iso = value.isoformat()
+    # if value.tzinfo:
+    #     # Use "Z" instead of a timezone offset of "+00:00" to indicate UTC.
```
thet: @buchi @sneridagh @tisto I'm favoring the "Z" notation of indicating the UTC, but that's maybe a personal flavor and I don't have a strong preference here. I'll remove this code block, agreed?

thet:
I might begin to agree that dates stored in UTC are more practical. While discussing the issue with @Rotonen he brought up an argument that date calculations with two different versions of pytz can produce different results. Which is bad. Although the timezone rules are part of the tzinfo objects and thus not necessarily prone to errors with different versions of pytz ... UTC dates have advantages.

However the UTC vs timezoned isn't actually an issue as @buchi mentioned above:
timezone conversion to the field's target zone is handled here: https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/deserializer/dxfields.py#L77

Do we need to hande Archetypes content also?
I'd like to have a second opinion on the two inline code comments in this PR.
Then this PR is ready to merge from my point of view.

tisto:
@thet would you mind adding a few lines to the upgrade guide regarding this? http://plonerestapi.readthedocs.io/en/latest/upgrade-guide.html#

PR #502, opened 2018-02-20
Observe the allow_discussion allowance (global, fti, object) on objec…

sneridagh:
…t serialization

### DIFF
--- a/docs/source/_json/collection.resp
+++ b/docs/source/_json/collection.resp
@@ -16,7 +16,7 @@ Content-Type: application/json
   "@id": "http://localhost:55001/plone/collection", 
   "@type": "Collection", 
   "UID": "SomeUUID000000000000000000000002", 
-  "allow_discussion": null, 
+  "allow_discussion": false, 
   "contributors": [], 
   "created": "2016-01-21T08:14:48+00:00", 
   "creators": [
--- a/src/plone/restapi/serializer/dxcontent.py
+++ b/src/plone/restapi/serializer/dxcontent.py
@@ -2,11 +2,12 @@
 from AccessControl import getSecurityManager
 from Acquisition import aq_inner
 from Acquisition import aq_parent
-from Products.CMFCore.utils import getToolByName
+from plone.app.discussion.interfaces import IDiscussionSettings
 from plone.autoform.interfaces import READ_PERMISSIONS_KEY
 from plone.dexterity.interfaces import IDexterityContainer
 from plone.dexterity.interfaces import IDexterityContent
 from plone.dexterity.utils import iterSchemata
+from plone.registry.interfaces import IRegistry
 from plone.restapi.batching import HypermediaBatch
 from plone.restapi.deserializer import boolean_value
 from plone.restapi.interfaces import IFieldSerializer
@@ -15,12 +16,13 @@ from plone.restapi.interfaces import ISerializeToJsonSummary
 from plone.restapi.serializer.converters import json_compatible
 from plone.restapi.serializer.expansion import expandable_elements
 from plone.supermodel.utils import mergedTaggedValueDict
+from Products.CMFCore.utils import getToolByName
 from zope.component import adapter
 from zope.component import getMultiAdapter
 from zope.component import queryMultiAdapter
 from zope.component import queryUtility
-from zope.interface import Interface
 from zope.interface import implementer
+from zope.interface import Interface
 from zope.schema import getFields
 from zope.security.interfaces import IPermission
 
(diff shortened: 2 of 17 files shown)


src/plone/restapi/serializer/dxcontent.py
```diff
@@ -84,6 +86,24 @@ def __call__(self, version=None, include_items=True):
...
+        registry = queryUtility(IRegistry)
+        settings = registry.forInterface(IDiscussionSettings, check=False)
+
+        # Check if discussion is allowed globally, overwritting whatever that
+        # is set in the object
+        if not settings.globally_enabled:
+            result['allow_discussion'] = False
+        else:
+            if 'allow_discussion' not in result or \
+               result.get('allow_discussion', False) is None:
```
jaroel: `result.get('allow_discussion', None) is None` does the same thing as both lines above

docs/source/_json/collection.resp
```diff
@@ -16,7 +16,7 @@ Content-Type: application/json
   "@id": "http://localhost:55001/plone/collection", 
   "@type": "Collection", 
   "UID": "SomeUUID000000000000000000000002", 
-  "allow_discussion": null, 
+  "allow_discussion": false,
```
jaroel: Did something change in the types as well?

tisto: @sneridagh @jaroel please note that "null" is a different setting than "false". "Null" means "check parent object setting" and "false" means "do not allow discussion".

tisto: Ok, got it. https://docs.google.com/document/d/1vKmHXzw2vVWYEVEvFtIUr7gcpf_3Qh7dhYx0cx5rUGM/edit#heading=h.xr2un7csqeq2 sets global discussion to False, therefore all objects return "False" by default.

src/plone/restapi/serializer/dxcontent.py
```diff
@@ -84,6 +86,24 @@ def __call__(self, version=None, include_items=True):
...
+                # The object allow_discussion attribute is not set, then
+                # lookup for the default value on the fti
+                portal_types = getToolByName(self.context, 'portal_types')
+                document_fti = getattr(portal_types, self.context.portal_type)
+                result['allow_discussion'] = document_fti.getProperty(
+                    'allow_discussion')
+
```
jaroel:
```
if not settings.globally_enabled:
    result['allow_discussion'] = False
elif result.get('allow_discussion', None) is None:
    # The object allow_discussion attribute is not set, then
    # lookup for the default value on the fti
    fti = portal_type[self.context.portal_type]
    result['allow_discussion'] = fti.getProperty('allow_discussion')
```

src/plone/restapi/serializer/dxcontent.py
```diff
@@ -84,6 +86,24 @@ def __call__(self, version=None, include_items=True):
                 value = serializer()
                 result[json_compatible(name)] = value
 
+        # Fetch discussion registry
```
tisto: @sneridagh maybe we should add a comment here where to find the corresponding plone.app.discussion code.

docs/source/_json/collection.resp
jaroel: @tisto where is this set?

jaroel: @tisto Ah yeah, sorry, that makes sense. This looks clean imho.

src/plone/restapi/serializer/dxcontent.py
sneridagh: I've just improved the condition a bit and added an additional test for it.

src/plone/restapi/serializer/dxcontent.py
sneridagh: @jaroel Finally changed the condition with your suggestion.

src/plone/restapi/serializer/dxcontent.py
sneridagh: Done.

sneridagh:
@tisto @jaroel Done with the comments review!

src/plone/restapi/serializer/dxcontent.py
```diff
@@ -84,6 +86,24 @@ def __call__(self, version=None, include_items=True):
...
+        # Check if discussion is allowed on the object
+        # The original code and condition worflow is in
+        # https://github.com/plone/plone.app.discussion/blob/master/plone/app/discussion/browser/conversation.py#L111
+        registry = queryUtility(IRegistry)
+        settings = registry.forInterface(IDiscussionSettings, check=False)
+
+        # Check if discussion is allowed globally, overwritting whatever that
+        # is set in the object
+        if not settings.globally_enabled:
+            result['allow_discussion'] = False
```
jaroel: @sneridagh What happens if there is no registry data for IDiscussionSettings? Does settings.globally_enabled fail?

src/plone/restapi/serializer/dxcontent.py
```diff
@@ -84,6 +86,24 @@ def __call__(self, version=None, include_items=True):
...
+        elif result.get('allow_discussion', None) is None:
+            # The object allow_discussion attribute is not set, then
+            # lookup for the default value on the fti
+            portal_types = getToolByName(self.context, 'portal_types')
+            document_fti = getattr(portal_types, self.context.portal_type)
+            result['allow_discussion'] = document_fti.getProperty(
+                'allow_discussion')
+
```
jaroel: Can we do 'fti' instead of 'document_fti', so the code is a bit cleaner?

sneridagh: Done.

src/plone/restapi/serializer/dxcontent.py
sneridagh: I guess p.a.discussion will be around in the current form and is not going anywhere for a long time, right? :P

jaroel: I was just wondering if we should remove the 'check=False' bit just above it :)

tisto: Why do we use check=False in the first place? Wouldn't it be better to fail hard in case p.a.discussion is not around? The other option is to really make it optional by checking if "setting != None" before proceeding...

sneridagh: Removed check=False

PR #507, opened 2018-02-27
New @email-send endpoint to send arbitrary mails from Plone

sneridagh:
The endpoint is protected with the default Plone "Use Mailhost Services" permission" (defaulting only to Manager and Site administrators).

I think that the mail still should look like it's been sent from Plone, so the message should be wrapped with that information, and the subject should be also custom, probably. The drawback of this is the i18n, since we are adding a feature, Plone lacks of the translations required for this. Any suggestion?

/cc @tisto


tisto:
@sneridagh ok, so as long as we don't add this feature to the core we have to add a plone.restapi po file for those translations, correct?

sneridagh:
@tisto Please review the code, give me feedback about the mail and subject texts and complete the German translations.

Issue #186, opened 2017-01-19
@search endpoint path.query does not work with VHM


sneridagh:
Given a site is behind a VHM config using a root server name, so the instance name is hidden, when you request

`@search?path.query=/`

You don't get nothing, and if you use the name of the instance:

`@search?path.query=/Plone`

 it works fine.

For consistency I would match the VHM configuration, so the query root matches the root of the server. Thus, the users don't have to know the instance name behind a server deployment.

/cc @tisto

lukasgraf:
@sneridagh I agree with the desired behavior you described. Hopefully I'll get around to a fix for this soon.

sneridagh:
Cool! I remember to have a similar bug while implementing p.a.m. and the
test fixture for testing it was insane :(

Please tell me if I can help in sonething.


On Thu, 19 Jan 2017 at 19:57, Lukas Graf <notifications@github.com> wrote:

@sneridagh <https://github.com/sneridagh> I agree with the desired behavior
you described. Hopefully I'll get around to a fix for this soon.




—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/plone/plone.restapi/issues/186#issuecomment-273865259>,
or mute the thread

sunew:
@sneridagh and @lukasgraf this was fixed by csengers PR (and I have fixed restricting to multiple paths behind vhm in https://github.com/plone/plone.restapi/pull/615 )

I close this issue :)

PR #508, opened 2018-03-07
Add VHM support to @search

csenger:
Solves issue #186 

If called via VHM it will include the VirtualHostRoot path in a path search

### DIFF
--- a/src/plone/restapi/search/handler.py
+++ b/src/plone/restapi/search/handler.py
@@ -33,6 +33,20 @@ class SearchHandler(object):
         if 'path' not in query:
             query['path'] = {}
 
+        if isinstance(query['path'], str):
+            query['path'] = {'query': query['path']}
+
+        # If this is accessed through a VHM the client does not know
+        # the complete physical path of an object. But the path index
+        # indexes the complete physical path. Complete the path.
+        vhm_physical_path = self.request.get('VirtualRootPhysicalPath')
+        if vhm_physical_path:
+            path = query['path'].get('query')
+            if path:
+                path = path.lstrip('/')
+                full_path = '/'.join(vhm_physical_path + (path,))
+                query['path']['query'] = full_path
+
         if isinstance(query['path'], dict) and 'query' not in query['path']:
             # We either had no 'path' parameter at all, or an incomplete
             # 'path' query dict (with just ExtendedPathIndex options (like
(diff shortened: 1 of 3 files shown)


tisto:
@sneridagh what's the status here?

src/plone/restapi/search/handler.py
```diff
@@ -33,6 +33,20 @@ def _constrain_query_by_path(self, query):
...
+        if isinstance(query['path'], str):
+            query['path'] = {'query': query['path']}
+
+        # If this is accessed through a VHM the client does not know
+        # the complete physical path of an object. But the path index
+        # indexes the complete physical path. Complete the path.
+        vhm_physical_path = self.request.get('VirtualRootPhysicalPath')
+        if vhm_physical_path:
+            path = query['path'].get('query')
+            if path:
```
lukasgraf: Could we change this to something along the lines of `if vhm_physical_path and not path.startswith('/'.join(vhm_physical_path)):`?

(Only prepend VirtualRootPhysicalPath if client didn't already supply it).

That would keep backwards compatibility with clients that for some reason already did know the Plone site ID (maybe communicated out-of-band), and still assist clients that don't know it.

csenger: I thought about that. This will fail if you have a folder in the site root with the same id as the site. As it is not a dev who decides that it i did not implement it.

I don't think we should make this behavior optional and the old behavior the default. So if backwards compatibility is a blocker it should go into a 2.0alpha release.

lukasgraf: @csenger yeah, I thought about that case too. I was thinking that with Plone and Acquisition you're not going to have a good time naming an object the same as the Plone site ID - but you're right, it's possible, though probably rare.

> I don't think we should make this behavior optional and the old behavior the default. So if backwards compatibility is a blocker it should go into a 2.0alpha release.

Yeah, I think you're right - offering two code paths here is probably more confusing than it helps 👍

tisto: @lukasgraf @csenger 2.0.0 is around the corner anyways. I don't mind just including that one in a new major release as a breaking change. Though, if this is a breaking change, we have to document it in our upgrade guide: http://plonerestapi.readthedocs.io/en/latest/upgrade-guide.html

PR #513, opened 2018-03-12
Support null in content PATCH to delete a field value #187

### LINKED ISSUE #187: Null value representation. WAS: How to clear a datetime field with json?

csenger:
This fixed #187

### DIFF
--- a/docs/source/content.rst
+++ b/docs/source/content.rst
@@ -182,7 +182,7 @@ PATCH allows to provide just a subset of the resource
 If you send the value `null` for a field, the field's content will be
 deleted. It will instead set the `missing_value` defined for the field
 in the schema. Note that this is not possible if the field is `required`,
-and it only works for dexterity types, not archetypes:
+and it only works for Dexterity types, not Archetypes:
 
 ..  http:example:: curl httpie python-requests
     :request: _json/content_patch.req
--- a/docs/source/content.rst
+++ b/docs/source/content.rst
@@ -176,7 +176,13 @@ Updating a Resource with PATCH
 ------------------------------
 
 To update an existing resource we send a PATCH request to the server.
-PATCH allows to provide just a subset of the resource (the values you actually want to change):
+PATCH allows to provide just a subset of the resource
+(the values you actually want to change).
+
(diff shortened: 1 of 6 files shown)


src/plone/restapi/deserializer/dxcontent.py
```diff
@@ -62,6 +62,22 @@ def __call__(self, validate_all=False, data=None):  # noqa: ignore=C901
...
+                        else:
+                            errors.append({
+                                'field': field.__name__,
+                                'message': (
+                                    '{} is a required field.'.format(
+                                        field.__name__
+                                    ),
+                                    'Setting it to null is not allowed.'
+                                ),
+                                'error': None})
```
lukasgraf: Hmm, `'error': None` feels a bit clunky. I don't have a clear suggestion in mind though.

docs/source/content.rst
```diff
@@ -176,7 +176,13 @@ Updating a Resource with PATCH
 ------------------------------
 
 To update an existing resource we send a PATCH request to the server.
-PATCH allows to provide just a subset of the resource (the values you actually want to change):
+PATCH allows to provide just a subset of the resource
+(the values you actually want to change).
+
+If you send the value `null` for a field, the field's content will be
```
lukasgraf: In reStructuredText these need to be double backticks (``) for code literals, single backticks are just emphasis 😉

PR #516, opened 2018-03-13
Expose widgets tagged values via @types endpoint

jaroel:


tisto:
@jaroel there is a test failure on Plone 4: https://travis-ci.org/plone/plone.restapi/jobs/352907727

tisto:
@jaroel this is still WIP, correct?

tisto:
@jaroel just a quick check. This is not a breaking change, correct? We just add the tagged values to the response. I am just asking because I plan to do a 2.0.0 release soon and I am wondering if we have to include this in the release or not (if it is a breaking change, we have to).

jaroel:
@tisto This only adds stuff, so unless someone is using type definitions for the json, this is not a breaking change.

This renders the keywords vocabulary in the schema, alike choice fields. Rob and I think that shouldn't be that much of a problem and allows us to have an autocomplete widget for keywords.

jaroel:
I'll check the Plone 4.x test failures.

PR #517, opened 2018-03-13
WIP: Tiles

tisto:


tisto:
@robgietema @sneridagh @csenger @datakurre here is a draft for the tiles endpoint. Please review and give feedback.

datakurre:
@tisto @robgietema What does "listing tiles" or "adding a tile" through API mean in this context?

TL; DR; in the current Mosaic, the "easiest" way to figure out all the tiles on the current page is to render it and look for data-tile -attributes in HTML (before tiles are merged).

Background: Surprisingly, in Deco and Mosaic, there is not and never was any method for retrieving current tiles used on some context (early versions had "tile book keeping", but it was abandoned later). This is still the case. The reason is that adding "transient tile" is simply about adding a single tag with data-tile-attribute into content are HTML through Deco / Mosaic editor. In the current Mosaic, all tiles with any configuration have their configuration persisted, but completely anonymous tiles are still supported.

tisto:
@datakurre we will definitely need to add a simple storage mechanism (like the layout in mosaic/plone.app.blocks) to store a list of existing types. So adding/deleting will manipulate that internal list and expose it via the REST API. Does that answer your question?

tisto:
@csenger can you confirm again please that the available tiles are context specific? Otherwise, we have to adapt the redesign we just did again.

datakurre:
@tisto Yes. That is something that would not work easily on current Mosaic, but probably works for you when that API design guides the implementation :)

tisto:
@datakurre I am open to discussing how we could make this work with standard Mosaic. Though, this might become a lot more complex...

csenger:
@tisto @datakurre Yes, the plan is to provide an API to create/read plone.tiles. It wouldn't be an API for Mosaic or plone.app.blocks. Having that on top of Mosaic will be pretty complex both for the API and the implementation. 

The question is how much is needed to write it mosaic compatible.

datakurre:
It’s probably better the other way around. You implement tiles as you see best and then we see if we can refactor Mosaic to match that.

tisto:
Just listing the changes we discussed in Bonn:

```
@tiles list of all tiles (content)
@tile-schema list of all tile schemata
@tile-types list of all available tile types
```

sneridagh:
Closing this in favor of #547 and the newnewtiles branch.

PR #520, opened 2018-03-27
Allow users to update their own properties and password.

sneridagh:


src/plone/restapi/services/users/update.py
```diff
@@ -47,29 +55,78 @@ def reply(self):
...
+                    password=None,
+                    roles=target_roles,
+                    domains=user.getDomains(),
+                )
+        elif self._get_current_user == self._get_user_id:
+            for key, value in user_settings_to_update.items():
+                if key == 'password' and \
+                   security.enable_user_pwd_choice and \
+                   self.can_set_own_password:
+                    user.userSetPassword(user.getUserId(), value)
```
jaroel: whitespace

src/plone/restapi/tests/test_services_users.py
```diff
@@ -441,6 +461,28 @@ def test_update_user_password(self):
...
+            'fullname': 'Noam A. Chomsky',
+            'username': 'noam',
+            'email': 'avram.chomsky@plone.org'
+        }
+        self.api_session.auth = ('noam', 'password')
+        response = self.anon_api_session.patch('/@users/noam', json=payload)
+
+        self.assertEqual(response.status_code, 401)
+
+    def test_update_user_not_self(self):
```
jaroel: I'd explain this test a bit. Took me a while to realise that the intent wasn't to see if the user /can/ edit another user, but /cannot/ edit another user.

src/plone/restapi/services/users/update.py
jaroel: I'm not to fond of the line wrapping, but I don't see an easy way out.
We could split this up in separate functions outside the class and then call them?

src/plone/restapi/tests/test_services_users.py
```diff
@@ -441,6 +461,28 @@ def test_update_user_password(self):
             old_password_hashes['noam'], new_password_hashes['noam']
         )
 
+    def test_update_user_anon(self):
```
tisto: test_update_user_anon -> test_anonymous_user_can_not_update_existing_user

src/plone/restapi/tests/test_services_users.py
tisto: test_update_user_not_self -> test_user_can_not_edit_another_user

src/plone/restapi/tests/test_services_users.py
```diff
@@ -409,6 +409,26 @@ def test_update_user(self):
             noam.getProperty('email')
         )
 
+    def test_update_user_by_himself(self):
```
tisto: test_update_user_by_himself -> test_user_can_update_himself

src/plone/restapi/services/users/update.py
tisto: Following clean code best practices would mean to split that up into functions or methods. Though, I don't see how this would become significantly "cleaner". If somebody wants to put efforts into that, fine. If not I am also ok with merging. Working code is better than no code at all. ;)

PR #537, opened 2018-06-22
Update to latest plone releases (4.3.17, 5.1.2); remove old buildout …

sunew:
…bootstrap files (we use pip install -r requirements.txt); split versions overrides for each plone version


tisto:
@sunew LGTM. Though, I am not sure if moving the package pinning to the individual buildout files is a good idea. This way we have to maintain multiple version pinnings. So far we got away with a single versions.cfg file. Can you elaborate why you did this?

sunew:
@tisto there were some 'upgrades' for the RC that would now be 'downgrades' for 5.1.2.
While at the same time we still need those pinnings for plone 4 and 5.0.
So we need to split the versions for 5.1 and the older releases.
Of course some are the same, such as Sphinx and other tools - I could move them to a shared versions.cfg
I don't know - it's either the plague or cholera ;)

sunew:
@tisto fixed. Now we have a common versions.cfg, and a pre-51 versions for the upgrades for plone 5.0 and 4.

sunew:
@tisto I suggest going with this as it is now - the old pins are needed to upgrade to tools for 5.0 and 4, (to keep it equivalent to the old buildout and test runs). 
While we certainly don't want to downgrade 5.1 versions, which would happen if we keep the versions as is.

sunew:
@tisto changed to be more in more in line with your feedback:)

PR #241, opened 2017-03-05
Add support for named IJsonSchemaProvider adapter to target a single field in a schema

jaroel:
This allows us to prevent rending choices for relatedItems, which would dump the whole DB.

Fixes #199 by not rendering relatedItems' choices.

PR #539, opened 2018-06-22
[WIP] Add the 'widget' property to the @types endpoint for IRelationList fields.

thet:


thet:
I add the WIP / inprogress tag... adding fixes as we're working on the relateditems widget on plone-react.

src/plone/restapi/types/configure.zcml
```diff
@@ -17,8 +17,7 @@
     <adapter factory=".adapters.ListJsonSchemaProvider" />
     <adapter
         zcml:condition="installed z3c.relationfield"
-        factory=".z3crelationadapter.ChoiceslessRelationListSchemaProvider"
-        name="IRelatedItems.relatedItems" />
+        factory=".z3crelationadapter.ChoiceslessRelationListSchemaProvider" />
```
thet: Note: I removed the name of the adapter as I do not see yet why this was necessary. To use the adapter you'd have to explicitly call it by it's name - the adapter wasn't applied if called without the name.

thet: @jaroel as you added the ChoiceslessRelationListSchemaProvider, I have two questions:
- Why a named adapter?
- Why the class name "Choiceless"?

jaroel: @thet It's a named adapter so I can target one field while getting the jsonschema properties, see https://github.com/plone/plone.restapi/blob/e6875793867e261bbb99c3b4d8e8507fbf724076/src/plone/restapi/types/utils.py#L110

It's called `Choiceless` because it prevents rendering the choices. This is set in https://github.com/plone/plone.restapi/blob/1b88d9366ec560e53a6e13f083cdd9f4f8d6ecf4/src/plone/restapi/types/z3crelationadapter.py#L23 and checked in https://github.com/plone/plone.restapi/blob/4c5707973cdb404d280d9af41456dc0b0ab170cb/src/plone/restapi/types/adapters.py#L274

jaroel: Also see https://github.com/plone/plone.restapi/pull/241

jensens:

PR #540, opened 2018-06-22
Translate the fti title in @types endpoint

### LINKED ISSUE #337: i18n support
Most i18n should be handled on the frontend but fields like tiles of contenttypes etc should be returned correctly by the REST API. My suggestion is to use the "Accept-Language" header for this.

erral:
This fixes #337

### DIFF
--- a/src/plone/restapi/tests/test_services_types.py
+++ b/src/plone/restapi/tests/test_services_types.py
@@ -1,11 +1,12 @@
 # -*- coding: utf-8 -*-
 from plone import api
-from plone.restapi.testing import PLONE_RESTAPI_DX_FUNCTIONAL_TESTING
-from plone.restapi.testing import RelativeSession
 from plone.app.testing import setRoles
-from plone.app.testing import TEST_USER_ID
 from plone.app.testing import SITE_OWNER_NAME
 from plone.app.testing import SITE_OWNER_PASSWORD
+from plone.app.testing import TEST_USER_ID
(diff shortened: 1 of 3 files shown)


tisto:
@erral LGTM. Please add a changelog entry.

erral:
I have a test for it, although I don't know if it's necessary, as far as we are already doing this translate stuff in some other places (like in the `@types/Folder` call) without the test: https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/types/utils.py#L189

lukasgraf:
I've cleaned up setting the available and supported languages, as well as enabling request language negotiation for my work on the `@roles` titles in a way where we can apply it for all our tests, for Plone 4 as well as Plone 5.

So you should be able to easily base your test on that once I've pushed.

erral:
#545 needs to be merged before this one.

src/plone/restapi/tests/test_services_types.py
```diff
@@ -1,15 +1,19 @@
...
 from plone.app.testing import SITE_OWNER_PASSWORD
+from plone.app.testing import TEST_USER_ID
+from plone.restapi.testing import PLONE_RESTAPI_DX_FUNCTIONAL_TESTING
+from plone.restapi.testing import RelativeSession
+
 
 import transaction
 import unittest
 
+PLONE_VERSION = parse_version(api.env.plone_version())
```
lukasgraf: This is unused I think.

erral: Yeah, traces of the code that I had before your changes :)

PR #543, opened 2018-06-22
Fix test_documentation (ReadConflictError)

sunew:
Fixes https://github.com/plone/plone.restapi/issues/470

The issue was the generation of new transaction ids in ZODB.utils.newTid.
For the documentation test only we patch the newTid to use the unpatched time and gmtime methods, in a new testing layer.

### DIFF
--- a/src/plone/restapi/tests/test_documentation.py
+++ b/src/plone/restapi/tests/test_documentation.py
@@ -23,8 +23,8 @@ from plone.namedfile.file import NamedBlobFile
 from plone.namedfile.file import NamedBlobImage
 from plone.registry.interfaces import IRegistry
 from plone.restapi.testing import PAM_INSTALLED
-from plone.restapi.testing import PLONE_RESTAPI_DX_FUNCTIONAL_TESTING
-from plone.restapi.testing import PLONE_RESTAPI_DX_PAM_FUNCTIONAL_TESTING
+from plone.restapi.testing import PLONE_RESTAPI_DX_PAM_FUNCTIONAL_TESTING_FREEZETIME  # noqa
+from plone.restapi.testing import PLONE_RESTAPI_DX_FUNCTIONAL_TESTING_FREEZETIME  # noqa
 from plone.restapi.testing import register_static_uuid_utility
 from plone.restapi.testing import RelativeSession
 from plone.testing.z2 import Browser
@@ -142,11 +142,9 @@ def save_request_and_response_for_docs(name, response):
 
 class TestDocumentation(unittest.TestCase):
 
-    layer = PLONE_RESTAPI_DX_FUNCTIONAL_TESTING
+    layer = PLONE_RESTAPI_DX_FUNCTIONAL_TESTING_FREEZETIME
 
     def setUp(self):
-        if PLONE_VERSION.base_version >= '5.1':
-            self.skipTest('Do not run documentation tests for Plone 5')
         self.app = self.layer['app']
         self.request = self.layer['request']
         self.portal = self.layer['portal']
@@ -1232,11 +1230,9 @@ class TestDocumentation(unittest.TestCase):
 
 class TestCommenting(unittest.TestCase):
 
-    layer = PLONE_RESTAPI_DX_FUNCTIONAL_TESTING
+    layer = PLONE_RESTAPI_DX_FUNCTIONAL_TESTING_FREEZETIME
 
     def setUp(self):
-        if PLONE_VERSION.base_version >= '5.1':
-            self.skipTest('Do not run documentation tests for Plone 5')
         self.app = self.layer['app']
         self.request = self.layer['request']
         self.portal = self.layer['portal']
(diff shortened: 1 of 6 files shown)


src/plone/restapi/tests/test_documentation.py
```diff
@@ -23,8 +23,10 @@
 from plone.namedfile.file import NamedBlobImage
 from plone.registry.interfaces import IRegistry
 from plone.restapi.testing import PAM_INSTALLED
-from plone.restapi.testing import PLONE_RESTAPI_DX_FUNCTIONAL_TESTING
-from plone.restapi.testing import PLONE_RESTAPI_DX_PAM_FUNCTIONAL_TESTING
+from plone.restapi.testing import (
+    PLONE_RESTAPI_DX_PAM_FUNCTIONAL_TESTING_FREEZETIME,
+    PLONE_RESTAPI_DX_FUNCTIONAL_TESTING_FREEZETIME
```
lukasgraf: Could you please split these imports onto seperate lines? That allows people to quickly keep imports sorted by just sorting lines.

docs/source/_json/content_get.resp
```diff
@@ -49,5 +49,6 @@ Content-Type: application/json
   "table_of_contents": null, 
   "text": null, 
   "title": "My Document", 
-  "version": "current"
+  "version": "current", 
+  "versioning_enabled": true
```
lukasgraf: I'm a bit confused about why these changes in the req/resp files happen here, even though this PR
- doesn't bump the Plone version
- doesn't remove the `skipTest()` from `test_documentation.py`

src/plone/restapi/tests/test_documentation.py
```diff
@@ -23,8 +23,10 @@
 from plone.namedfile.file import NamedBlobImage
```
lukasgraf: Hmm - the `skipTest` for Plone >= 5.1 still doesn't get removed in this PR though, no? Is this intentional?

src/plone/restapi/tests/test_documentation.py
sunew: yep - I had to make them two-line because of the line length requirement.

I had this, but thought it ugly:

```
from plone.restapi.testing import (
    PLONE_RESTAPI_DX_FUNCTIONAL_TESTING_FREEZETIME
)
from plone.restapi.testing import (
    PLONE_RESTAPI_DX_PAM_FUNCTIONAL_TESTING_FREEZETIME
)
```

sunew: btw I use isort - it handles the sorting for me ... :)

sunew: and I added the standard plone .isort config in https://github.com/plone/plone.restapi/pull/537
makes things easier...

lukasgraf: You can just slap a `# noqa` at the end of the import. [See here](https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/services/auth/login.py#L7) for example. Having a as-stupid-as-they-get sort-tool being able to get the job done instead of requiring isort is preferable :)

src/plone/restapi/tests/test_documentation.py
sunew: no - I'll look at that now

src/plone/restapi/tests/test_documentation.py
sunew: yep I know `# noqa`  - added.

src/plone/restapi/tests/test_documentation.py
sunew: @lukasgraf what skipTest fo you mean? There is no such thing. 
There is 

```
    @unittest.skipIf(not PLONE5, 'Just Plone 5 currently.')
    def test_controlpanels_get_item(self):
```
but this is not related to 5.1 or not.

lukasgraf: [This one](https://github.com/plone/plone.restapi/blob/fb94ad786dfea7c3e9d949f5a2e9ebc0e7ea8dd5/src/plone/restapi/tests/test_documentation.py#L148-L149), and two more further down - those are the ones that skip the tests that previously failed with the `ReadConflict`, but only for Plone >= 5.1

docs/source/_json/content_get.resp
sunew: I was running it with plone 5.1 - that explains it perhaps. 
Currently builds 4 and 5.1 succeeds, while 5.0 fails. (because of documentation artifacts)

lukasgraf: Ah, righ - that's what I would have expected to happen. I would recommend to keep 5.0 as the reference-build for the documentation artifacts (that's defined [here](https://github.com/plone/plone.restapi/blob/master/test-no-uncommitted-doc-changes.in#L16-L21), though the comment didn't get updated when the reference build got switched from 4.3 to 5.0).

I would address actually moving to 5.1 (for the default buildout, and the reference build for docs artifacts, and updating them / checking changes) in a separate PR.

lukasgraf:
Awesome, nice work! :+1: :tada: :cake:

@tisto I think this one is ready.

sunew:
also coredev builds are failing...for instance https://jenkins.plone.org/job/pull-request-5.2/891/consoleFull

PR #547, opened 2018-06-23
Tiles endpoint final implementation and related behavior for storing them and its layout

sneridagh:


setup.py
```diff
@@ -56,6 +56,8 @@
           'plone.api',
           'requests',
           'freezegun',
+          'plone.tiles',
+          'plone.schema'
```
tisto: @sneridagh missing comma.

PR #548, opened 2018-06-23
Plone51 as test documentation basis 

sunew:
Based on fix-documentation-test-on-51
Fist merge https://github.com/plone/plone.restapi/pull/543

### DIFF
--- a/docs/source/_json/types_document.resp
+++ b/docs/source/_json/types_document.resp
@@ -18,6 +18,7 @@ Content-Type: application/json+schema
         "allow_discussion", 
         "exclude_from_nav", 
         "id", 
+        "versioning_enabled", 
         "table_of_contents"
       ], 
       "id": "settings", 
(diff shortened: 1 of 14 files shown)


docs/source/_json/types_document.resp
```diff
@@ -140,36 +140,18 @@ Content-Type: application/json+schema
...
-        "es", 
-        "fr"
+        "en"
       ], 
       "enumNames": [
-        "Deutsch", 
-        "English", 
-        "Espa\u00f1ol", 
-        "Fran\u00e7ais"
+        "English"
```
lukasgraf: Hmm, where did those languages go? They should still include those for, those should be set as available and supported languages for all our tests. ([See testing.py](https://github.com/plone/plone.restapi/blob/5afe77b22bed7ca7f7d7680163866e69448bf4b0/src/plone/restapi/testing.py#L52-L71))

PR #549, opened 2018-06-23
Fix object creation events

buchi:
This makes object creation in plone.restapi behave more like TTW object creation.

New objects are now created in the following way:
1. Create an empty object
2. Deserialize it
3. Fire object created event
4. Add object to container
5. Fire object added and container modified events

Before that change new objects were created in the this way:
1. Create an empty object
2. Fire object created event
3. Add object to container
4. Fire object added and container modified events
5. Deserialize object
6. Fire modified event

### DIFF
--- a/src/plone/restapi/services/content/utils.py
+++ b/src/plone/restapi/services/content/utils.py
@@ -3,6 +3,8 @@ from Acquisition import aq_base
 from Acquisition import aq_parent
 from DateTime import DateTime
 from plone.app.content.interfaces import INameFromTitle
+from plone.app.uuid.utils import uuidToObject
+from plone.uuid.interfaces import IUUID
 from Products.CMFCore.utils import getToolByName
 from Products.CMFPlone.utils import base_hasattr
 from random import randint
@@ -76,9 +78,15 @@ def add(container, obj):
         notifyContainerModified(container)
         return obj
     else:
-        rval = container._setObject(id_, obj)
-        new_id = isinstance(rval, basestring) and rval or id_
-        return container._getOb(new_id)
+        new_id = container._setObject(id_, obj)
+        # _setObject triggers ObjectAddedEvent which can end up triggering a
+        # content rule to move the item to a different container. In this case
+        # look up the object by UUID.
+        try:
+            return container._getOb(new_id)
+        except AttributeError:
+            uuid = IUUID(obj)
+            return uuidToObject(uuid)
 
 
 def rename(obj):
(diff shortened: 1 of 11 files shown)


src/plone/restapi/services/content/utils.py
```diff
@@ -22,17 +28,56 @@ def create(container, type_, id_=None, title=None):
...
+
+    return obj
+
+
+def add(container, obj):
+    """Add an object to a container."""
+    id_ = obj.getId()
+    # Archetypes objects are already created in a container thus we just fire
+    # the notification events.
+    if container == aq_parent(obj):
```
davisagli: Comparing acquisition-wrapped objects with `==` can be problematic (two different wrappers of the same object are not equal). Probably best to do `if aq_base(container) is aq_base(aq_parent(obj))`

And maybe we should raise an exception if obj already has a different parent?

buchi: Thanks, I've amended that.

Not sure, how a different parent would be possible at all? Have you something in mind how this could happen?

tisto:
@buchi this is at least a possible breaking change, correct?

buchi:
@tisto It shouldn't break anything unless you're heavily relying on plone.restapi internals, e.g. customizing the endpoint for adding objects.

buchi:
@tisto Turns out this can break custom content deserializers. I've added a note to the changelog.

tisto:
@buchi thank you! Would you mind adding that section to our upgrade guide as well?

src/plone/restapi/services/content/utils.py
```diff
@@ -22,17 +29,56 @@ def create(container, type_, id_=None, title=None):
...
+    # Archetypes objects are already created in a container thus we just fire
+    # the notification events.
+    if aq_base(container) is aq_base(aq_parent(obj)):
+        notify(ObjectAddedEvent(obj, container, id_))
+        notifyContainerModified(container)
+        return obj
+    else:
+        rval = container._setObject(id_, obj)
+        new_id = isinstance(rval, basestring) and rval or id_
+        return container._getOb(new_id)
```
davisagli: _setObject triggers ObjectAddedEvent which can end up triggering a content rule to move the item to a different container. In this case we might need the same workaround that we use in plone.dexterity.utils.addContentToContainer to work around this edge case:

    try:
        return container._getOb(newName)
    except AttributeError:
        uuid = IUUID(object)
        return uuidToObject(uuid)

buchi:
@tisto added small section in the upgrade guide.

tisto:
@buchi awesome. Thank you!
@davisagli would you mind having another look and approving the merge?

PR #551, opened 2018-06-23
Reindex AT content on PATCH

### LINKED ISSUE #531: PATCH doesn't reindex content
I tried to update object and it have changed but catalog record not. 

I use:
- Plone 5.0.6
- plone.restapi 2.0
- Archetypes 1.11.2

buchi:
Fixes #531

### DIFF
--- a/src/plone/restapi/deserializer/atcontent.py
+++ b/src/plone/restapi/deserializer/atcontent.py
@@ -67,6 +67,7 @@ class DeserializeFromJson(OrderingMixin, object):
                 notify(ObjectInitializedEvent(obj))
                 obj.at_post_create_script()
             else:
+                obj.reindexObject()
                 notify(ObjectEditedEvent(obj))
                 obj.at_post_edit_script()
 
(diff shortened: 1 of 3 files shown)


src/plone/restapi/deserializer/atcontent.py
```diff
@@ -67,6 +67,7 @@ def __call__(self, validate_all=False, data=None):
                 notify(ObjectInitializedEvent(obj))
                 obj.at_post_create_script()
             else:
+                obj.reindexObject()
```
davisagli: Do we not also want this when the creation flag is set?

buchi: When the creation flag is set, the object will be added to a container and reindexed because of that: https://github.com/zopefoundation/Products.CMFCore/blob/master/Products/CMFCore/CMFCatalogAware.py#L272
Thus I would like to avoid indexing twice, though collective.indexing would solve that.

davisagli: Okay!

PR #553, opened 2018-06-23
Create image URLs using the cache optimized URLs

### LINKED ISSUE #494: Image scales are not using cache optimized url form.
The image URLs of the scales should use the UID based format as at other places in Plone to ensure caches are updated after upload of a new image.

What I expect to get:

```
    "image_fullscreen": {
    ...
        "scales": {
            "icon": {
                "download": "http://192.168.101.191:8080/Plone/os/at/vw/model-groups/top-banner/@@images/e38ea213-2059-455b-9294-2c9554b2ba33.jpeg",
                "height": 8,
                "width": 32
            },
    ...
```
What I get from restapi:

```
    "image_fullscreen": {
    ...
        "scales": {
            "icon": {
                "download": "http://192.168.101.191:8080/Plone/os/at/vw/model-groups/top-banner/@@images/image_fullscreen/icon",
                "height": 8,
                "width": 32
            },
    ...
```

erral:
This fixes #494


erral:
I had to patch the uuid4 method, and achieved that using `mock` as dependency, I don't know if that's good practice here... I haven't seen something similar in the codebase.

tisto:
@erral would you mind adding a short note to our upgrade guide? If you don't have time for it I can also take care of it tomorrow...

erral:
Done.

tisto:
@erral the real implementation returns different URLs with different UUIDs for every single scale as described here:

https://github.com/plone/plone.restapi/pull/553/files#diff-772cee6f1ad591450b99fdde08976749

right?

The auto-generated response just has the same URL/UUID for all image scales just because of the mock, is that correct?

erral:
Yes, that's right.

erral:
There is a note on the documentation explaining it: https://github.com/plone/plone.restapi/pull/553/files#diff-c294eb935ad014e7ea890908cbd17e75

PR #554, opened 2018-06-23
Have tests fail on Sphinx documentation errors / warnings

lukasgraf:
Adds a **test that fails if Sphinx build has warnings or errors.**

*(Changelog entry omitted on purpose since it's not a user-visible bugfix/feature)* 

This is so we can keep documentation at a high quality level, and avoid piling up too many warnings. Especially frustrating is documentation that has been carefully written, but never got published because it wasn't included in a toctree.

The test is a simple bash script that checks the output and return code of `bin/sphinxbuild` and errors out to fail the build if either
- `bin/sphinxbuild` exited with return code 1 (hard errors)
- the output contains WARNINGs

In addition, I fixed all the existing Sphinx warnings.

The build failure is displayed in a way on travis where it should be immediately visible to developers / first time contributors why the build failed.

---

Example [travis build failure](https://travis-ci.org/plone/plone.restapi/jobs/395849798#L2700):

sphinx_warnings

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,7 @@ script:
   - bin/sphinxbuilder
   - bin/test
   - bin/test-no-uncommitted-doc-changes
+  - bin/test-no-sphinx-warnings
 after_success:
   - bin/test-coverage
   - pip install coverage==3.7.1 coveralls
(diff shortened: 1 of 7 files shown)


davisagli:
Nice!

PR #561, opened 2018-06-24
Fix change password on Plone 5.2.

sunew:
After the change of MemberData to MemberAdapter, there is no longer a method userSetPassword of the user instance. Using the same approach as the PasswordResetTool.

### DIFF
--- a/src/plone/restapi/services/users/update.py
+++ b/src/plone/restapi/services/users/update.py
@@ -46,6 +46,10 @@ class UsersPatch(Service):
         portal_membership = getToolByName(portal, 'portal_membership')
         return portal_membership.getMemberById(user_id)
 
+    def _change_user_password(self, user, value):
+        acl_users = getToolByName(self.context, 'acl_users')
+        acl_users.userSetPassword(user.getUserId(), value)
+
     def reply(self):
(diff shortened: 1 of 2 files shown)


src/plone/restapi/services/users/update.py
```diff
@@ -46,6 +46,14 @@ def _get_user(self, user_id):
         portal_membership = getToolByName(portal, 'portal_membership')
         return portal_membership.getMemberById(user_id)
 
+    def _change_user_password(self, user, value):
+        try:
+            user.userSetPassword(user.getUserId(), value)
+        except AttributeError:
+            # plone >= 5.2
+            acl_users = getToolByName(self.context, 'acl_users')
+            acl_users.userSetPassword(user.getUserId(), value)
```
davisagli: Any reason not to do it this way on older versions of Plone too?

sunew: hmm, not really, I guess it works there also. 
Did not want to change previous behavior though. But would be cleaner. 
@tisto @lukasgraf opinions?

lukasgraf: If it works, I'm all for removing unnecessary conditionals.

PR #562, opened 2018-06-24
Fix tests on 52

sunew:


src/plone/restapi/tests/test_roles.py
```diff
@@ -22,7 +22,7 @@ def setUp(self):
     def test_roles_endpoint_lists_roles(self):
         response = self.api_session.get('/@roles')
 
-        self.assertEqual([
+        self.assertItemsEqual([
```
tisto: @sunew can you elaborate why "assertEqual" stopped working in Plone 5.2?

sunew: The list order changed in plone 5.2
We should use assertItemsEqual when comparing lists.

tisto: Cool. Thank you @sunew!

PR #566, opened 2018-06-25
Add groups to @user

csenger:
* Add support to save groups in PATCH/POST
* Add conditional support to retrive groups in GET


tisto:
@csenger would you mind fixing the merge conflicts?

@csenger @sneridagh @buchi @lukasgraf I am wondering if we should use "True" as default for the "include_groups" param to be in sync with how we use "include_items":

https://plonerestapi.readthedocs.io/en/latest/content.html?highlight=include_items

This means that the default call might be a bit slower but it would be more consistent IMHO. Opinions?

csenger:
> @csenger would you mind fixing the merge conflicts?

@tisto  I fixed them.

> @csenger @sneridagh @buchi @lukasgraf I am wondering if we should use "True" as default for the "include_groups" param to be in sync with how we use "include_items":
> 
> https://plonerestapi.readthedocs.io/en/latest/content.html?highlight=include_items
> 
> This means that the default call might be a bit slower but it would be more consistent IMHO. Opinions?

We discussed the approach last year and you suggested using an expander back then. I completed the code I had for that and removed the direct parameter. That works fine except now on Plone 5.2 where the breadcrums expander adapter is also found for the user object and fails. Probably due to aquisition changes on the user object. The breadcrumb and maybe other expander need to be fixed. 

Any opinion which way to fix them is better:
* registering them for a specific interfaces that provide a matching object. In this case it needs absolute_url(), so an content specific interface should work. But there is a small risk to break a working use case.
* Fix the adapter code to handle users and other objects that don't provide absolute_url()

jensens:

PR #568, opened 2018-06-27
Properly include HTTP examples in installed egg

lukasgraf:
Make sure to include HTTP examples in the **installed egg**, so `test_documentation` tests also work against an installed release of `plone.restapi`.

We need to do this dynamically because the `data_files` argument to `setup()` doesn't support globs (wildcards).

The examples need to be included [via `data_files`](https://docs.python.org/3/distutils/setupscript.html#installing-additional-files) because they are outside a Python package. So we can't include them using `package_data`, which [only works relative to Python packages](https://docs.python.org/3/distutils/setupscript.html#installing-package-data). (The `MANIFEST` only controls what gets included in the *source distribution*. Listing these files in `data_files` ensures they actually get *copied to the installed `.egg`*).

If the HTTP examples directory is ever moved, the `HTTP_EXAMPLES_PATH` in `setup.py` needs to be updated before a new release is cut. I included a tripwire test that should catch this ahead of release time.

---

I tested the changes to `setup.py` as well as I could locally (ran `python setup.py sdist` as well as `python setup.py install`), so the collecting and copying of the HTTP examples into the egg works.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,9 @@ Changelog
 3.1.1 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Make sure to include HTTP examples in installed egg, so test_documentation
+  tests also work against a installed release of plone.restapi.
+  [lgraf]
 
 
 3.1.0 (2018-06-27)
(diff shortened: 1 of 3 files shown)


tisto:
Thank you @lukasgraf!

Issue #552, opened 2018-06-23
Make tests run on plone 5.2


sunew:
- [x] Documentation tests / freeze time fails on 5.2
- [x] Failure in test_roles
- [x] test_update_user_password:
MemberData' object has no attribute 'userSetPassword'

- [x] Port allocation plugin on Jenkins leads to different ports
- [ ] test_tus_head_on_not_existing_resource_returns_404
- [ ] we are down to 8 skipped tests

Running under coredev / plip-2177-plone-restapi.cfg

tisto:
https://jenkins.plone.org/view/PLIPs/job/plip-plone-restapi/9/#showFailuresLink

I see two main issues:

1) Port allocation plugin on Jenkins leads to different ports and lots of tests have port 55001 hard coded.

2) test_tus_head_on_not_existing_resource_returns_404

sunew:
@tisto fixed 1.
2.: The TUS problem only happens on Jenkins, not in a local coredev 5.2 build. I looked a bit at it, but have no explanation. It could be related to missing cleanup, and a different test order on Jenkins.

tisto:
@sunew awesome!!! I get one test failure locally on the PLIP job:

```
Failure in test test_tus_head_on_not_existing_resource_returns_404 (plone.restapi.tests.test_tus.TestTUS)
Traceback (most recent call last):
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/unittest/case.py", line 331, in run
    testMethod()
  File "/Users/timo/workspace/buildout.coredev-5.2/src/plone.restapi/src/plone/restapi/tests/test_tus.py", line 155, in test_tus_head_on_not_existing_resource_returns_404
    self.assertEqual(404, response.status_code)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/unittest/case.py", line 515, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/unittest/case.py", line 508, in _baseAssertEqual
    raise self.failureException(msg)
AssertionError: 404 != 200
```

It seems TUS returns a 200 instead of a 404 on non-existing resources. Maybe @lukasgraf or @buchi could have a quick look.

tisto:
Seems to be a test isolation issue. No test failure when I run the test individually:

`bin/test -t test_tus_head_on_not_existing_resource_returns_404`

[...]

PR #569, opened 2018-06-27
Cleanup TUS-tests (to avoid test failure on 5.2) 

sunew:
See https://github.com/plone/plone.restapi/issues/552

### DIFF
--- a/src/plone/restapi/tests/test_tus.py
+++ b/src/plone/restapi/tests/test_tus.py
@@ -105,6 +105,7 @@ class TestTUS(unittest.TestCase):
         stored_metadata = upload.metadata()
         self.assertEqual(stored_metadata,
                          {u'length': 8, u'mode': u'create'})
+        upload.cleanup()
 
     def test_tus_post_initialization_with_metadata(self):
         metadata = 'filename {},content-type {}'.format(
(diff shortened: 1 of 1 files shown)


src/plone/restapi/tests/test_tus.py
```diff
@@ -105,6 +105,7 @@ def test_tus_post_initialization(self):
         stored_metadata = upload.metadata()
         self.assertEqual(stored_metadata,
                          {u'length': 8, u'mode': u'create'})
+        upload.cleanup()
```
tisto: @sunew can't we just add this to the test tear down method?

tisto:
@sunew I agree. My fix was just a quick fix and temporary solution.

src/plone/restapi/tests/test_tus.py
sunew: the `upload` or the `tus` variables are local for each test case. The TUDUpload instance is created in different ways in different testcases. 
One option woul be to have a tus=None in the setUp() and then do
`self.tus = TUSUpload('myuid', {'length': 2048})`
instead of 
`tus = TUSUpload('myuid', {'length': 2048})`
(and rename ´upload´ to ´tus´ in some cases)

But i don't know if that's a nice thing to do...?

tisto: @sunew I see. I guess then your version is cleaner and easier to understand. Thanks for the clarification.

PR #573, opened 2018-06-28
Do not include HTTP examples using data_files anymore.

### LINKED ISSUE #572: Writing to disc when installing the egg (buildout or pip)
#568 introduced a write on the disc on egg installation at the root of the project, @lukasgraf is that correct? It might lead to some errors on buildout and pip deployments where the permissions does not allow that.

lukasgraf:
This is a follow up to #568 - including the HTTP example files in `data_files` was a misguided approach. It sort of worked using buildout / `easy_install`, where those files got installed into the `.egg` (apparently this is a [`setuptools` bug](https://github.com/pypa/setuptools/issues/130)).

However, using `pip`, they get installed to the project root (`sys.prefix`) [as specified by the `setuptools` documentation](https://docs.python.org/3/distutils/setupscript.html#installing-additional-files). This was undesired, and can lead to permission issues during deployment, for example in Docker images.

---

This change should hopefully fix this: It **moves the HTTP example `.json` files *inside the Python package* `plone.restapi.tests`** (in a subdirectory `http-examples`, which doesn't have to be a Python package itself)
- that way, they can be treated as first-class package data
- because MANIFEST already includes `graft src/plone/restapi`, we don't even need to list them in `package_data`, they should get picked up and bundled automatically during release
- These files can now be properly referenced using `pkg_resources.resource_filename`

Tested locally using:
- `buildout`
- `pip`

Tests pass, `test_documentation` properly updates the files, Sphinx docs build correctly, the `.egg` created via `pip install` correctly contains the files *inside the egg*, and the `test-no-uncommitted-docs-changes-test` still works.

Fixes #572


sunew:
@lukasgraf looks good!
Why not copy the files over to the docs folder during the build of the documentation? (Or similar approach)
It could be good if the docs (and the genartion of the docs) were independent of the tests being run.

lukasgraf:
@sunew we could - but what exactly would we gain from that?

I see two possible scenarios when building the docs:
- examples are up to date (tests have been run)
- examples are out of date

The result is the same for both those cases, independent of whether we introduce an additional copy-when-building step or not.

The downside is, for every change there's now to redundant sets of files to check in and to maintain, instead of a canonical source of truth.

PR #621, opened 2018-10-01
Run Python 3 on Travis (with allow failures)

tisto:


davisagli:
Hmm, it seems like a waste to try to test with Python 3 on Plone 4.3, 5.0 and 5.1. Is there a way we can avoid that and only run the tests in Python 3 with Plone 5.2? Probably not until there's an alpha release with passing tests...

tisto:
@davisagli had to go to bed before I could exclude those. :)

This is just an experiment that we can merge once there is a final 5.2. Then we will, of course, exclude all non 5.2 versions of Plone from running Py3 tests...

tisto:
@davisagli done. Now we run just the necessary jobs on Travis.

PR #622, opened 2018-10-02
Fix test: normalize difference between Plone 5.2 and previous versions

davisagli:


davisagli:
@tisto do you want a changelog for a test fix like this?

PR #623, opened 2018-10-02
[WIP] Drop collective mockmailhost

gforcada:
I see that @pbauer did port c.MockMailHost to python 3, but if we can avoid it with a simpler solution (from CMFPlone), isn't that all the better? :smile:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,8 +4,8 @@ Changelog
 3.4.6 (unreleased)
 ------------------
 
-- Nothing changed yet.
-
+- Drop collective.MockMailHost, use a util class from CMFPlone.
+  [gforcada]
 
 3.4.5 (2018-09-14)
 ------------------
--- a/setup.py
+++ b/setup.py
@@ -52,7 +52,6 @@ setup(name='plone.restapi',
       ],
       extras_require={'test': [
           'Products.Archetypes',
-          'collective.MockMailHost',
           'plone.app.collection',
           'plone.app.contenttypes',
           'plone.app.robotframework',
--- a/setup.py
+++ b/setup.py
@@ -44,9 +44,9 @@ setup(name='plone.restapi',
       install_requires=[
           'setuptools',
           'python-dateutil',
-          'plone.behavior>=1.1',  # adds name to behavior directive 
+          'plone.behavior>=1.1',  # adds name to behavior directive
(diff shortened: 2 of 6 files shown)


gforcada:
Seems that travis has some  problems? jenkins 5.2 did run fine

tisto:
@gforcada not sure if I agree here. Mocking the mailhost in a separate package was an improvement over manually defining it in every single package. I'd recommend discussing that kind of things first before moving to the implementation to avoid useless work and possible frustration...

gforcada:
@tisto I am using some other package mocking implementation (in my first iteration I did copy the code over, but then as I saw that plone.restapi already depends on cmfplone I just imported it).

The main benefit of this is to remove two dependencies: collective.mockmailhost and products.securemailhost. Relying in less packages is not a good idea?

@pbauer did a port to python 3 of collective.mockmailhost that also drops products.securemailhost, should we instead try to get that merged and use that here in restapi? :thinking:

tisto:
@gforcada so your first idea is to add the mockmailhost to CMFPlone directly? That would also be an option. I just would like to avoid that we duplicate the code to mock the mailhost over and over again...

gforcada:
@tisto no, CMFPlone has already a mock class that mimics collective.MockMailHost, see it yourself: 

https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/tests/utils.py#L37-L57

tisto:
@gforcada ah, ok. Is the implementation identical to c.mockmailhost? Then we can indeed drop that dependency.

tisto:
@gforcada is it possible that the new MockMailHost in CMFPlone is only present since 5.0? Travis fails for older Plone versions. We have to keep p.restapi compatible with older Plone versions as well. We either have to add MockMailHost to older Plone versions as well or keep things as they are for now...

gforcada:
@tisto I haven't checked if it is a 1:1 replacement, though the tests on 5.1 say that they are at least enough to make plone.restapi tests pass.

We can backport the utility function on CMFPlone 4.3 and 5.0, or I can make it conditional here, although then setup.py would be a bit of a problem, should we depend on it or not...

I will backport the utility function, it can not hurt, would it be ok to depend on a checked out version of CMFPlone to run the tests? :thinking:

PR #634, opened 2018-11-11
Proposal: docs for indexes endpoint

davisagli:


tisto:
@davisagli docs look good to me. Now we just need to find someone who wants to work on the implementation. :)

Any takers? That might be a good task for a p.restapi newbie since the endpoint itself is rather trivial.

davisagli:
Closing; information about indexes is available in the querystring endpoint https://plonerestapi.readthedocs.io/en/latest/querystring.html

PR #635, opened 2018-11-11
Standardize errors data structure of email-notification endpoint

cekk:
Errors had a different data structure than the other endpoints

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,10 @@ Changelog
 3.5.1 (unreleased)
 ------------------
 
-- Nothing changed yet.
+Bugfixes:
+
+- Standardize errors data structure of email-notification endpoint.
+  [cekk]
 
 
 3.5.0 (2018-11-06)
--- a/src/plone/restapi/services/email_notification/post.py
+++ b/src/plone/restapi/services/email_notification/post.py
@@ -1,7 +1,8 @@
 # -*- coding: utf-8 -*-
-from plone.restapi.services import Service
-from zope.component import getMultiAdapter
 from plone.restapi.deserializer import json_body
+from plone.restapi.services import Service
+from zExceptions import BadRequest
+from zope.component import getMultiAdapter
 from zope.interface import alsoProvides
 
 import plone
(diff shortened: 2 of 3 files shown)


tisto:
@hvelarde @jensens I appreciate your work on reviewing and merging open pull requests. Though, are you aware that this is actually a breaking change? We are changing the API here, which means we need to do a major release (for just a very minor API change) and we need to write the docs in the upgrade steps. This is IMHO too much work and too much hassle for a very minor improvement. If we fix the error reporting, we should do it all at once or together with other breaking changes IMHO.

There is a reason why I did not merge that pull request right away and there is a reason why the improvement of the error reporting is pending since quite a while. It is complex and needs more documentation and testing.

plone.restapi is very different from other core packages because it will continue to live as an add-on product, it supports four different Plone versions and two different content type frameworks. It also uses, in contrast to plone core, semantic versioning. In addition, we follow a documentation first and no-release without full documentation approach.

We have to take this into account when reviewing and merging pull request IMHO. If we don't we will end up in a situation where the master branch is not releasable without reverting merges and this causes serious problems in the projects where we use plone.restapi. We rely the fact that we can do plone.restapi releases at any time. Otherwise, we would have to create forks or work on git checkouts which I would like to avoid.

Anyways, this is not the first time something like this happened and I am sure you had the best intentions when doing this. As said, I appreciate your work here. We just have to discuss how we treat plone.restapi in the future, to ensure it can continue to move fast (faster than plone core) and keep up with our plone.restapi standards when it comes to releases and documentation.

cc @plone/framework-team we might want to combine that with the Guillotina/plone.restapi API discussion since we also have to take into account that changes in plone.restapi affect Guillotina (and the other way around). Complex topic...

@cekk just for the record: this discussion has nothing to do with your pr, which is fine. I just did not merge/approve it yet because I wanted to prevent a merge from happening. :)

hvelarde:
@tisto I'm sorry to hear that; feel free to revert the pull request in case it causes so many troubles to you.

tisto:
@cekk I am afraid we have to revert this PR merge in order to be able to do another minor plone.restapi release. Are you ok with this? I would like to avoid doing a major version bump just for a minor change like this. I did not had time to look into how to do this but I will try to avoid such unnecessary work in the future.

tisto:
Revert of this PR merge: https://github.com/plone/plone.restapi/pull/666

tisto:
@cekk would you mind giving this another try and creating a new pull request? We have another breaking change in the pipeline and I would like to combine them to make a new 4.0.0 release.

tisto:
@cekk ping :)

cekk:
@tisto damn, i was sure to replied to this issue :(

Sure, i can add a new pr. I've found another branch related to this issue: https://github.com/plone/plone.restapi/tree/fix-email-notification-error-2

Is it the right one right now? Should i sync it with master and then create a new pr?

Issue #636, opened 2018-11-12
Lots of sockets not closed during testing


gforcada:
Have a look at any 3.6/3.7 jenkins jobs, i.e. https://jenkins.plone.org/job/plone-5.2-python-3.6/ or https://jenkins.plone.org/job/plone-5.2-python-3.7/ and you will see that after this line:

``Set up plone.restapi.testing.PloneRestApiDXLayer:Functional in 0.000 seconds.``

There are lots, literally lots, of:

```
zope/testrunner/runner.py:386: ResourceWarning: unclosed <socket.socket fd=8, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('127.0.0.1', 51612), raddr=('127.0.0.1', 50139)>
  test.__dict__.clear()
```

On the up side, it feels like all of them come from the very same place, so with one stone all of them might  be gone :+1:

davisagli:
My guess is something about the StopableWSGIServer started by plone.testing (which is a thin wrapper around waitress).

mauritsvanrees:
So this happens during the call to `test.__dict__.clear()`. In my case it looks like this:

```
(Pdb) pp test.__dict__
{'_cleanups': [],
 '_outcome': None,
 '_subtest': None,
 '_testMethodDoc': None,
 '_testMethodName': 'test_documentation_users_add',
 '_type_equality_funcs': {<class 'list'>: 'assertListEqual',
                          <class 'dict'>: 'assertDictEqual',
                          <class 'set'>: 'assertSetEqual',
                          <class 'frozenset'>: 'assertSetEqual',
                          <class 'tuple'>: 'assertTupleEqual',
                          <class 'str'>: 'assertMultiLineEqual'},
 'api_session': <plone.restapi.testing.RelativeSession object at 0x108d98ef0>,
 'app': <Application at >,
 'browser': <plone.testing._z2_testbrowser.Browser object at 0x110f71b70>,
 'document': <Document at /plone/front-page>,
 'frozen_time': <freezegun.api.FrozenDateTimeFactory object at 0x11105dda0>,
 'portal': <PloneSite at /plone>,
 'portal_url': 'http://localhost:53215/plone',
 'request': <HTTPRequest, URL=None>,
 'time_freezer': <freezegun.api._freeze_time object at 0x108e1d6a0>}
```

I can remove several keys from it without problem:

```
(Pdb) del test.__dict__['_outcome']
[...]

Issue #648, opened 2018-12-20
Jenkins failure


gforcada:
Any idea what could go wrong here? It is failing since yesterday :confused: 
```
Error in test test_documentation_search_fullobjects (plone.restapi.tests.test_documentation.TestDocumentation)
Traceback (most recent call last):
  File "/srv/python2.7/lib/python2.7/unittest/case.py", line 329, in run
    testMethod()
  File "/home/jenkins/workspace/plone-5.2-python-2.7/src/plone.restapi/src/plone/restapi/tests/test_documentation.py", line 499, in test_documentation_search_fullobjects
    response = self.api_session.get('/@search', params=query)
  File "/home/jenkins/.buildout/eggs/cp27m/requests-2.21.0-py2.7.egg/requests/sessions.py", line 546, in get
    return self.request('GET', url, **kwargs)
  File "/home/jenkins/workspace/plone-5.2-python-2.7/src/plone.restapi/src/plone/restapi/testing.py", line 359, in request
    return super(RelativeSession, self).request(method, url, **kwargs)
  File "/home/jenkins/.buildout/eggs/cp27m/requests-2.21.0-py2.7.egg/requests/sessions.py", line 533, in request
    resp = self.send(prep, **send_kwargs)
  File "/home/jenkins/.buildout/eggs/cp27m/requests-2.21.0-py2.7.egg/requests/sessions.py", line 646, in send
    r = adapter.send(request, **kwargs)
  File "/home/jenkins/.buildout/eggs/cp27m/requests-2.21.0-py2.7.egg/requests/adapters.py", line 498, in send
    raise ConnectionError(err, request=request)
ConnectionError: ('Connection aborted.', error(104, 'Connection reset by peer'))
```

tisto:
@gforcada connection timeout when the test tries to do a request against the functional test fixture. Maybe a port allocation problem on Jenkins?

mauritsvanrees:
In the traceback above, the error is in `test_documentation_search_fullobjects`. On [latest Jenkins](https://jenkins.plone.org/job/plone-5.2-python-2.7/lastCompletedBuild/testReport/plone.restapi.tests.test_documentation/TestDocumentation/test_documentation_search/) it is in `test_documentation_search`.

mauritsvanrees:
On Python3 locally I get hundreds of these warnings, each with a different port number:

```
.../zope.testrunner-4.9.2-py3.7.egg/zope/testrunner/runner.py:386: 
[...]

PR #651, opened 2018-12-27
Close the api_session in tests.

mauritsvanrees:
This prevents lots of ResourceWarnings about unclosed sockets.
Fixes https://github.com/plone/plone.restapi/issues/636 and https://github.com/plone/plone.restapi/issues/648.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,12 @@ Changelog
 
 Bugfixes:
 
+- Close the api_session in tests.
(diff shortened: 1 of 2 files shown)


mauritsvanrees:
I see less unclosed socket warnings, but there are still lots of them, so this may need to be done in more places.

Rotonen:
One root cause of seeing those now might be a newer `zope.testrunner` masking less warnings and new infra being wider (allowing for more concurrent builds per server).

PR #652, opened 2018-12-28
Fix more warnings

gforcada:
With this PR all sockets should be closed.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,7 +10,7 @@ Bugfixes:
   This prevents lots of ResourceWarnings about unclosed sockets.
   Fixes issues `636 <https://github.com/plone/plone.restapi/issues/636>`_
   and `648 <https://github.com/plone/plone.restapi/issues/648>`_.
-  [maurits]
+  [maurits, gforcada]
 
 - Standardize errors data structure of email-notification endpoint.
   [cekk]
--- a/src/plone/restapi/tests/test_batching.py
+++ b/src/plone/restapi/tests/test_batching.py
@@ -35,6 +35,9 @@ class TestBatchingDXBase(unittest.TestCase):
         self.api_session.headers.update({'Accept': 'application/json'})
         self.api_session.auth = (SITE_OWNER_NAME, SITE_OWNER_PASSWORD)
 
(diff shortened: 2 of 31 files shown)


CHANGES.rst
```diff
@@ -10,7 +10,7 @@ Bugfixes:
   This prevents lots of ResourceWarnings about unclosed sockets.
   Fixes issues `636 <https://github.com/plone/plone.restapi/issues/636>`_
   and `648 <https://github.com/plone/plone.restapi/issues/648>`_.
-  [maurits]
+  [maurits] [gforcada]
```
tisto: @gforcada we usually use a comma-separated list here, right?

src/plone/restapi/tests/test_batching.py
```diff
@@ -35,6 +35,10 @@ def setUp(self):
         self.api_session.headers.update({'Accept': 'application/json'})
         self.api_session.auth = (SITE_OWNER_NAME, SITE_OWNER_PASSWORD)
 
+    def tearDown(self):
```
tisto: @gforcada is there a specific reason why you call "super" after the other calls? Just wondering...

tisto:
@gforcada LGTM. Apart from my two questions. Maybe somebody with more in-depth knowledge of our current issues with the test layers can do a second review. Thanks for your work!

CHANGES.rst
gforcada: No idea, I will change it

src/plone/restapi/tests/test_batching.py
gforcada: Just in case there is another `tearDown` method up in the chain? I will remove them though, if tests are usually not stacked

tisto: I get why you call "super" in general. I was just wondering why "super" is not the first line in the "tearDown" method.

gforcada: Oh, sorry, I call it last, as it is tearing down: first you clean your local changes and then let the higher up layers clean, just the opposite as in setUp

gforcada:
@tisto changes done

Issue #660, opened 2019-01-13
Dexterity deserialization acquisition problem


thet:
When deserializing nested content structures which all set the same field, fields in child nodes are not set unless they defer from the parent.

This is due to the deserializer checks via the z3c.form DataManager if the attribute already exists and if ihe value difers from what is going to be set. See:

https://github.com/plone/plone.restapi/blob/48bb26bbaa8ef30a9e4fe518e1ba45cd7ca814aa/src/plone/restapi/deserializer/dxcontent.py#L97

The DataManager adapter has to be called with an Acquisition unwrapped context.

A PR is in the works: #

jensens:
Closing as outdated (6 years old). Nested DX acquisition bug on Plone 5.x. Deserialization reworked significantly.
If you think this issue is still relevant please reopen it. Thank you for the original report!

PR #661, opened 2019-01-13
Fix problem when deserializing for nested Dexterity content structure…

thet:
…s and setting values for the same field.

When creating a nested content structure and setting the same value for same fields, a value for a sub node was not set when it was the same value as it's parent node due to Acquisition. The attribute for the sub node doesn't exist at time of creation, so the datamanger compares the value to be set against the parent's attribute value.
Fixes: #660

### DIFF
--- a/src/plone/restapi/deserializer/dxcontent.py
+++ b/src/plone/restapi/deserializer/dxcontent.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+from Acquisition import aq_base
 from AccessControl import getSecurityManager
 from plone.autoform.interfaces import WRITE_PERMISSIONS_KEY
 from plone.dexterity.interfaces import IDexterityContent
@@ -55,7 +56,10 @@ class DeserializeFromJson(OrderingMixin, object):
                     continue
 
                 if name in data:
-                    dm = queryMultiAdapter((self.context, field), IDataManager)
+                    dm = queryMultiAdapter(
+                        (aq_base(self.context), field),
+                        IDataManager
+                    )
                     if not dm.canWrite():
                         continue
 
@@ -94,7 +98,14 @@ class DeserializeFromJson(OrderingMixin, object):
                             'message': e.doc(), 'field': name, 'error': e})
                     else:
                         field_data[name] = value
-                        if value != dm.get():
+
+                        dm_value = object()  # Unique value to compare against.
+                        try:
+                            dm_value = dm.get()
+                        except AttributeError:
+                            pass
+
+                        if value != dm_value:
                             dm.set(value)
                             # Collect the names of the modified fields
                             # Use prefixed name because z3c.form does so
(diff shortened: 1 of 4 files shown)


src/plone/restapi/deserializer/dxcontent.py
```diff
@@ -25,6 +26,9 @@
 from .mixins import OrderingMixin
 
 
+MARKER = object()
```
tisto: @thet could you elaborate why you define this outside of the class?

thet: No specific reason - only, that this feels more efficient. Initialized once, reusable anywhere.
Do you prefer to have it within the scope of the block where it is used? I'm fine with that.

thet:
FYI: I plan to add some tests.

src/plone/restapi/deserializer/dxcontent.py
tisto: @thet I'd prefer readability over premature optimization. :) Therefore I'd put it next to where it is used.

tisto:
@thet any updates here? :)

tisto:
@thet merge conflict due to the towncrier merge. Can you have a look? Let me know if you want me to review again.

tisto:
@thet any updates? :)

thet:
@tisto tests added, but I cannot reproduce the problem within tests :/

src/plone/restapi/profiles/testing/types.xml
```diff
@@ -6,5 +6,6 @@
  <object name="ATTestFolder"
     meta_type="Factory-based Type Information with dynamic views"/>
  <object name="DXTestDocument" meta_type="Dexterity FTI" />
+ <object name="DXTestFolder" meta_type="Dexterity FTI" />
```
tisto: @thet would it be possible to get away without adding that text folder here? I personally really dislike testing profiles because they hide the test setup in a remote place instead of making it obvious in the test itself what is set up. They are also super inflexible. If you change something in the testing profile you always risk to break multiple tests. Getting rid of the test profile has been on my todo list for quite some time and I would like to avoid adding additional work here for me. This is not your fault, of course, since the testing profile was there already. You would do me a personal favor if you could refactor that. :)

tisto:
@thet tests are passing. minor code analysis violations. :)

src/plone/restapi/profiles/testing/types.xml
thet: I can easily get rid of the DXTestFolder when changung DXTestDocument's base class from ``Item`` to ``Container``.
I can also check if I can just use Folder+DXTestDocument. Should be possible if I can leave the ``title`` without a value for testing.

thet:
Updated the tests - but there are still minor code violations.

I can't still reproduce the initial problem in tests :/

I got it when creating content like here, using from ``plone.restapi.services.content.utils.create`` https://github.com/collective/collective.contentcreator/blob/master/src/collective/contentcreator/__init__.py#L102

The behavior's field I tried to set but which failed was not required with no default value.

PR #667, opened 2019-02-05
Plone 5.1.5

tisto:


src/plone/restapi/tests/http-examples/registry_get_list.resp
```diff
@@ -495,5 +495,5 @@ Content-Type: application/json
       "value": false
     }
   ], 
-  "items_total": 1673
+  "items_total": 1683
```
tisto: @plone/framework-team did we add additional registry entries in a minor version upgrade of Plone? The plone.restapi documentation tests detected this change when upgrading from Plone 5.1.4 to Plone 5.1.5.

Issue #671, opened 2019-02-15
Feature: Improve Plone Site Root object API


sneridagh:
- Improved serialization/deserialization of the Plone Site root to make possible an opt-in to support Tiles attributes on it.
- @workflow endpoint now answers an empty object instead of a 404 for the Plone Site Root object, hiding the implementation details (the Plone Site Root has no workflow for historical reasons) and equalizing it to the other content types, making the API more consistent.

PR #672, opened 2019-02-15
Improve Plone Site Root object API

sneridagh:
Addresses #671 

/cc @tisto


sneridagh:

src/plone/restapi/services/workflow/info.py
```diff
@@ -28,6 +29,13 @@ def __call__(self, expand=False):
         if not expand:
             return result
 
+        # Prevent 404 on site root on workflow request
+        # Although 404 will be more semantic, for the sake of uniformity of the
+        # API we fake the response to the endpoint by providing an empty
+        # response instead of a 404.
+        if IPloneSiteRoot.providedBy(self.context):
+            result['workflow'].update({'history': [], 'transitions': []})
```
lukasgraf: Immediately `return result` here? No point in even trying all the rest, right? That would make it clearer what happens here.

src/plone/restapi/tests/test_deserialize_siteroot.py
```diff
@@ -0,0 +1,46 @@
+# -*- coding: utf-8 -*-
```
lukasgraf: There already is an existing [`src/plone/restapi/tests/test_site_deserializer.py`](https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/tests/test_site_deserializer.py) - can you please integrate your testing changes into that one instead? (And also rename it to something like `TestSiteRootDeserializer` - the Plone site is not DX content).

src/plone/restapi/tests/http-examples/siteroot.resp
```diff
@@ -15,6 +15,7 @@ Content-Type: application/json
   },
```
lukasgraf: While the http examples have been updated, the narrative documentation seems to be out of date. https://plonerestapi.readthedocs.io/en/latest/tiles.html needs a major update I think. I'm honestly quite confused at this point when trying to determine which *tiles* related functionality is implemented, and where, and where the documentation is just stubbed out, talks about future ideas that may or may not happen, or is outdated.

Please, please try to keep the documentation in sync with API changes - it's so very hard to even to attempt to do it after the fact, after things have changed over time.

tisto: @lukasgraf you are absolutely correct. Thanks for pointing that out. I created a separate issue for that since it is not directly related to the PR. https://github.com/plone/plone.restapi/issues/674

src/plone/restapi/services/workflow/info.py
tisto: Fixed. Thx for the hint!

src/plone/restapi/tests/test_deserialize_siteroot.py
tisto: I moved the tests into test_site_deserializer.py and renamed the class name. Thx for the pointer!

news/671.feature
```diff
@@ -0,0 +1,7 @@
+Improved serialization/deserialization of the Plone Site root to make
+possible an opt-in to support Tiles attributes on it. The @workflow endpoint
+now answers an empty object instead of a 404 for the Plone Site Root object,
+hiding the implementation details (the Plone Site Root has no workflow for
+historical reasons) and equalizing it to the other content types, making the
+API more consistent.
+[sneridagh]
```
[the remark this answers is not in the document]
tisto: I tried to improve the wording and I split up the entry. @sneridagh please double check if you are happy with my amendment.

tisto:
@lukasgraf thanks for your detailed and thoughtful review (as always)! I will look into the tiles endpoint update asap.

tisto:

news/671.feature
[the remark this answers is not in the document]
lukasgraf: @tisto thank you! IMHO that summarizes the changes perfectly. From what I understand, the "opt-in" part just referred to the fact that that the tiles-specific fields aren't required in the deserializer, which is the case in most places in the API, so that can be omitted.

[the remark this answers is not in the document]
sneridagh: @tisto Absolutelly fine with me!!

sneridagh:
@lukasgraf Thanks for the review!! Agreed that we have to improve the documentation in both plone.restapi and Volto itself (almost not existent) that's a must that we should address in the next major sprint.

PR #678, opened 2019-02-28
Add values to vocabulary GET responses

Rotonen:
This is required to, for example, build choice fields in the front end.

Seems `six` is now also pinned to something too new for 4.3 based on the Travis build?


Rotonen:

src/plone/restapi/serializer/vocabularies.py
```diff
@@ -43,8 +43,10 @@ def __call__(self, vocabulary_id):
         term = self.context
         token = term.token
         title = term.title if ITitledTokenizedTerm.providedBy(term) else token
+        value = term.value
         return {
             '@id': '{}/{}'.format(vocabulary_id, token),
             'token': token,
-            'title': title
+            'title': title,
+            'value': value
```
lukasgraf: I think it would be prudent to run `value` through the [`json_compatible()`](https://github.com/plone/plone.restapi/blob/1d4d683704f5f4b319e0bff111dcdb5984a8d308/src/plone/restapi/serializer/converters.py#L54-L72) helper function. This will ensure that any value is serialized the same way to JSON like it is in other places in `plone.restapi`.

Rotonen: Amended. Thank you.

Rotonen:

Rotonen:
And it'd seem the WSGI layer does not automatically pick the port and multiple layers failed due to that.

davisagli:
This doesn't make sense to me. A Zope vocabulary term value could be anything, including a full content object, and is not necessarily JSON-serializable. That's why the token exists, to serve as a unique serializable identifier of the vocabulary term.

jensens:
@Rotonen I agree with @vangheem - the value is meant to keep in the backend as it can be anything (I used binary images as values already). The token is made as a bridge to the form-ready representation, translated forth and back. 

May you explain your use case to get us some insight why you want to do this?

lukasgraf:
@davisagli @jensens good point, I see why this can't work, we're barking up the wrong tree here.

Our use case was that we've got a `Choice` field with a vocabulary that is a list of `Int`s:

```python
    retention_period = schema.Choice(
        title=_(u'label_retention_period', u'Retention period (years)'),
        source=u'lifecycle_retention_period_vocabulary',
        required=True,
    )
```
Vocabulary (simplified):

```python
retention_period_choices = SimpleVocabulary([
    SimpleTerm(value=5, token='5', title='5'),
    SimpleTerm(value=10, token='10', title='10'),
    SimpleTerm(value=30, token='30', title='30'),
])
```
If the JS frontend uses the token (string) for that field during a POST or PATCH, field validation will fail because `plone.restapi` will on deserialization validate the field's value against the vocabulary's *values* (which only contains integers), and not its tokens.

The [DX field deserializers](https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/deserializer/dxfields.py) in  `plone.restapi` currently make no attempt at transforming the token to a value on ingest. So it looks to me like this is the underlying issue that would need to be addressed here.

Would you agree?

Rotonen:
Back to the drawing board.

I'll try to make the tests of the current master green first and I'll only then come back to this.

Help / ideas / an implementation is also most welcome in the meanwhile.

davisagli:
@lukasgraf that makes sense to me as the underlying issue. Values would also need to be transformed to tokens during serialization.

PR #684, opened 2019-03-05
Fix release

gforcada:
- the release version must use periods between micro and suffix (if any), otherwise setuptools keeps complaining about it
- universal wheels can not be built due to a check on setup.py regarding which python version is being used to install the package

### DIFF
--- a/news/682.feature
+++ b/news/682.feature
@@ -1,2 +1,0 @@
-- Add group roles to @groups serializer
-  [sneridagh]
--- a/news/684.bugfix
+++ b/news/684.bugfix
@@ -1,0 +1,1 @@
+Fix release to not create universal (Python 2/3) wheels.
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,0 @@
-[bdist_wheel]
-universal = 1
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@ from setuptools import setup, find_packages
 
 import sys
 
-version = '3.7.1-dev.0'
+version = '3.7.1.dev.0'
 
 long_description = (
     open('README.rst').read() + '\n' +


tisto:
Will merge and release as soon as Jenkins is green...

tisto:
@mauritsvanrees @gforcada I am still unable to use towncrier on plone.restapi. I synced the versions with buildout.coredev but still nothing happens when I run "bin/fullrelease".

mauritsvanrees:
@tisto Maybe you are hit by the same bug as Guido, see https://github.com/collective/zestreleaser.towncrier/issues/6. Well, in his case he gets an actual error, so the `fullrelease` fails. I fixed that in version 1.0.2 yesterday. Maybe that helps for you as well.
I haven't updated the coredev version yet, because a new feature/fix is coming in shortly.

BTW, the root cause was that `pyroma` was messing up `sys.argv`, which meant we could not find `towncrier`. So when `fullrelease` asks if you want to run `pyroma`, answer 'No' and see if that helps.

mauritsvanrees:
On Gil's branch, I have removed an outdated snippet and added a new one for this fix.

mauritsvanrees:

tisto:
@gforcada @mauritsvanrees the two Jenkins failures in robot tests seem to be unrelated. Shall I merge an release now?

mauritsvanrees:
I hate those Jenkins test failures where for some reason browser setup randomly fails in one or two tests... But that is for another day.

Yes, fine to merge and release. Thanks!
(You could try zestreleaser.towncrier 1.1.0.)

tisto:
@mauritsvanrees upgraded to zestreleaser.towncrier. 1.1.0 and it is still not working:

```
bin/fullrelease 
INFO: Starting prerelease.
Run pyroma on the package before tagging? (Y/n)? n
Do you want to run check-manifest? (Y/n)? n
WARNING: Found more than one file, picked the shortest one to change: CHANGES.rst, docs/source/history.rst
Changelog entries for version 3.7.1:

3.7.1 (unreleased)
------------------

- Nothing changed yet.


3.7.0 (2019-03-04)
------------------
Enter version [3.7.1]: 
WARNING: Changelog contains "- Nothing changed yet.". Are you sure you want to release? (y/N)? N
INFO: You can use the 'lasttaglog' command to see the commits since the last tag.
```

Could it be that this fails because zest.releaser can not identify the changelog file?

mauritsvanrees:
No, that is not the point. It finds two files and picks the correct one.
Let me run the `plone.restapi` buildout and see if I get the same problem, because I haven't done that yet.

Aha! `zestreleaser.towncrier` is not working here, because it is actually not installed!
The `base.cfg` installs `zest.releaser[recommended]`, but `zestreleaser.towncrier` is not in there.

When I add it, it works.
Well, apparently I need to explicitly install `towncrier` itself too, even though the module is installed. We currently need the script itself to be installed, which does not happen unless we set `dependent-scripts = true` in the `releaser` part.

I will make a PR for good measure.

PR #687, opened 2019-03-05
Install zestreleaser towncrier

mauritsvanrees:


mauritsvanrees:
This does not affect Jenkins, so let's not run any of its jobs.

mauritsvanrees:
The explicit inclusion of `towncrier` in the `release` part should not be needed anymore with `zestreleaser.towncrier` version 1.2.0, just released.

PR #688, opened 2019-03-05
use environment-markers instead of python-logic

pbauer:
See https://www.python.org/dev/peps/pep-0508/#environment-markers

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,7 @@
 from setuptools import setup, find_packages
 
+import sys
+
 version = '3.7.3-dev.0'
 
 long_description = (
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,5 @@
 from setuptools import setup, find_packages
 
-import sys
-
 version = '3.7.1.dev.0'
 
 long_description = (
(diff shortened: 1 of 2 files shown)


gforcada:
@pbauer @mauritsvanrees do wheels respect these markers? i.e. can we turn on again the universal wheel then with this? :thinking:

mauritsvanrees:
I suppose so, but I haven't tried these markers.

tisto:
Nice. Can we try out the markers on another (not so super important) package first maybe? :)

pbauer:
We use these marker in Plone: https://github.com/plone/Plone/blob/master/setup.py
Not not important 😉

pbauer:

gforcada:
And if you are worried, just make an alpha release, so that nobody accidentally gets that release.

setup.py
```diff
@@ -1,8 +1,6 @@
 from setuptools import setup, find_packages
 
-import sys
-
-version = '3.7.2-dev.0'
+version = '3.7.1.dev.0'
```
jensens: Version decrease by accident?

jensens:
Environment markers in its current version are from active [PEP 508](/#environment-markers) but are dating back to a more basic version (like we use them) in accepted [PEP 345](https://www.python.org/dev/peps/pep-0345/#environment-markers). Wheels are designed to respect whatever metadata is provided (see [PEP 427](https://www.python.org/dev/peps/pep-0427)) and package it. Pip reads the metadata and is [responsible to respect](https://pip.pypa.io/en/stable/reference/pip_install/#requirement-specifiers) it.

So, yes, this is the way to go. With or without wheel. And best put [all data in a `setup.cfg` (example) ] (https://github.com/collective/collective.es.index/blob/master/setup.cfg) and a one liner in ``setup.py`` to never ever have the idea again in future to use logic for metadata.

setup.py
tisto: Fixed.

tisto:

tisto:

PR #550, opened 2018-06-23
Filtering and batching for @vocabularies

davisagli:
This implements the first and second requirements for #535

ToDo:

- [x] Refactor q -> query (https://github.com/plone/plone.restapi/pull/713)
- [x] Make decision on how to link the vocabulary (plain vs. hypermedia link) (https://github.com/plone/plone.restapi/pull/714)
- [x] Write upgrade guide (https://github.com/plone/plone.restapi/pull/715)

PR #692, opened 2019-03-10
Use tokens for serialization and deserialization of vocabulary terms

### LINKED ISSUE #691: Serialization and deserialization of vocabulary terms should use tokens
Currently the serialization of a field with a vocabulary (e.g. ChoiceField) returns the stored value while the deserialization expects the value. So far this works well for values that are JSON serializable.

However the vocabularies endpoint only returns the token and the title of a vocabulary term. This makes it impossible to set a new value from the vocabulary if the token and the value are not equal.
Also term values that are not JSON serializable can not be handled this way.

To solve this, serialization should return tokens instead of values and deserialization should handle tokens.

buchi:
Fixes #691 

Deserialization from term values still works, however serialization always returns the token.
This may break compatibility with API clients if the term's value is not equal to it's token.

### DIFF
--- a/docs/source/upgrade-guide.rst
+++ b/docs/source/upgrade-guide.rst
@@ -7,6 +7,47 @@ This upgrade guide lists all breaking changes in plone.restapi and explains the 
 Upgrading to plone.restapi 4.x
 ------------------------------
 
+Serialization and Deserialization of fields with vocabularies
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The serialization of fields with vocabularies (e.g. ``Choice``) now returns the
+token of the vocabulary term instead of the stored value. For text values this
+should not make much a difference as the token and the value are usually the
+same. However if the term's value is not equal to it's token it may be neccessary
+to adopt the client implementation.
+
+Example:
+
+The date and time controlpanel previously returned a number for the
+``first_weekday`` property::
+
+  {
+    "@id": "http://localhost:55001/plone/@controlpanels/date-and-time",
+    "data": {
+        ...
+        "first_weekday": 0,
+        ...
+    }
+    ...
+  }
+
+Now it returns a string::
+
+  {
+    "@id": "http://localhost:55001/plone/@controlpanels/date-and-time",
+    "data": {
+        ...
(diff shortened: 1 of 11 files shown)


sneridagh:
@buchi is this any related with #550? I guess we will have to merge them both at some point.

/cc @davisagli what's the status there? Waiting for the Volto side implementation?

buchi:
@sneridagh no, it isn't. But it would be a good opportunity to include #550 as well in the next release with breaking changes.

buchi:

sneridagh:
@buchi I'm preparing some work before the Sorrento Sprint and I would like to see this PR and #550 merged and released. I'm also preparing https://github.com/plone/volto/pull/214 to support right away the new vocabularies in Volto and cover the breaking change.

However, I'm integrating `react-select` component and I just realised of one thing: in fact we need both `token` and `title` (display value) of them for existing values in order to render the component. Not only one of them.

If we only provide the token, how did you plan to know the display value? I am missing something? 

What do you think?

/cc @tisto @robgietema

sneridagh:
@tisto @buchi @lukasgraf @jaroel Which was the rationale for "additionalItems" in the types schema? I thought that it has to do if the field allows for values different than the ones in the options (aka subject field) but it seems that it's set to True always... :( 
https://github.com/plone/plone.restapi/blob/99c83047c44d5e17ea568f8ac4b886512bd700d8/src/plone/restapi/types/adapters.py#L206

I know that in subjects widget we just hardcode that even in Plone itself, but I wonder if is there any way to make the API more expressive on this matter? It would be great if at some point we can tell just from the API the right widget to use in any situation (and the ability to override it, which we are already able to do that)

jaroel:
@sneridagh I guess it should be treated as False if it's not defined? Not sure if that makes sense, just a gut feeling.
A quick search doesn't seem to show anything related in Volto, though I might be missing something here: https://github.com/plone/volto/search?q=additionalItems&unscoped_q=additionalItems

tisto:
@buchi I'd like to release p.restapi 4.0.0 tomorrow, wrapping up all pending breaking changes. Would you mind adding a short upgrade guide, or at least some pointers about the breaking changes?

tisto:

tisto:
Thank you @buchi! I will do a 4.0.0 release with this soon.

Issue #526, opened 2018-05-03
When using this package as a dependency in plone.importexport, I get zope.configuration.config.ConfigurationExecutionError: <class 'zope.interface.interfaces.ComponentLookupError'>: (<InterfaceClass zope.security.interfaces.IPermission>, 'plone.app.controlpanel.Overview')


idgserpro:
Please check https://github.com/collective/plone.importexport/issues/13. I only solved it by explicitly adding:

`plone.restapi-2.0.0-py2.7.egg/plone/restapi/services/controlpanels/configure.zcml`:

```
  <include package="plone.app.controlpanel" file="permissions.zcml"/>
```

And to `plone.restapi-2.0.0-py2.7.egg/plone/restapi/services/multilingual/configure.zcml`:

```
    <include package="plone.app.multilingual" file="permissions.zcml"/>
```

Should this be added here? Why the problem happens in a plone.importexport context but [not in your CI](https://travis-ci.org/plone/plone.restapi/builds/371875639)?

tisto:
@idgserpro please provide us with more information so we can help. Try to provide us with a minimal example that we can use to re-produce the problem.

idgserpro:
Just by running https://github.com/collective/plone.importexport#installation this error would raise. But I just tested and it seems the error is gone. If you test it like it's stated in the link and `bin/instance fg` works I think this issue an be closed.

tisto:
Fixed and released with https://pypi.org/project/plone.restapi/3.7.4/#id1

PR #694, opened 2019-03-12
Fix a problem on ZCML loading depending on how the policy package is …

sneridagh:
…named, related to the load of permissions in control panels and multilingual.

Difficult create a test for it...

### DIFF
--- a/src/plone/restapi/services/multilingual/configure.zcml
+++ b/src/plone/restapi/services/multilingual/configure.zcml
@@ -3,6 +3,8 @@
     xmlns:plone="http://namespaces.plone.org/plone"
     xmlns:zcml="http://namespaces.zope.org/zcml">
 
+    <include package="plone.app.multilingual" file="permissions.zcml"/>
+
     <adapter factory=".pam.Translations" name="translations"/>
 
     <plone:service
--- a/news/526.bugfix
+++ b/news/526.bugfix
@@ -1,0 +1,3 @@
+- Fix a problem on ZCML loading depending on how the policy package is named,
+  related to the load of permissions in control panels and multilingual.
+  [sneridagh]
--- a/setup.py
+++ b/setup.py
@@ -70,6 +70,7 @@ setup(name='plone.restapi',
           'plone.schema >= 1.2.0',  # new json field
           'PyJWT',
           'pytz',
+          'plone.app.controlpanel'  # to remove in Plone 6.
       ],
       extras_require={'test': TEST_REQUIRES},
       entry_points="""
--- a/src/plone/restapi/services/controlpanels/configure.zcml
+++ b/src/plone/restapi/services/controlpanels/configure.zcml
@@ -2,6 +2,8 @@
     xmlns="http://namespaces.zope.org/zope"
     xmlns:plone="http://namespaces.plone.org/plone">
 
+  <include package="plone.app.controlpanel" file="permissions.zcml" />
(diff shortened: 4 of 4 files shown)


sneridagh:
Related to #526.

sneridagh:
@mauritsvanrees @pbauer @esteele 
5.2 fails, it doesn't find `plone.app.controlpanel`, is it possible we do not load it by default in tests anymore? It's a problem if we need to support it in both 4.3 and 5.2.

https://travis-ci.org/plone/plone.restapi/jobs/505331283#L1400

I can add a zcml condition, but not ideal...

sneridagh:

sneridagh:
ok, the last commit did the trick.

src/plone/restapi/services/multilingual/configure.zcml
```diff
@@ -3,6 +3,8 @@
     xmlns:plone="http://namespaces.plone.org/plone"
     xmlns:zcml="http://namespaces.zope.org/zcml">
 
+    <include package="plone.app.multilingual" file="permissions.zcml"/>
```
tisto: @sneridagh don't we have to add p.a.multilingual to setup.py as well then? I mean at least for Plone 4? Or we wrap it in a conditional zcml...

tisto: ok, it is already wrapped properly: https://github.com/plone/plone.restapi/blob/55bfaf38c62a041c18ca94b60a703c5eacecbbee/src/plone/restapi/services/configure.zcml#L29

sneridagh: @tisto it is already wrapped https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/services/configure.zcml#L29

pbauer:
Nonono! This pulls `plone.app.controlpanel` into Plone but it is no longer a dependency. Please make it conditional. This broke the coredev:

```
Uninstalling test.
Uninstalling instance.
Unused options for buildout: 'config-directory' 'template-directory'.
Updating _mr.developer.
Installing instance.
Getting distribution for 'plone.app.controlpanel'.
warning: no previously-included files matching '*pyc' found anywhere in distribution
Got plone.app.controlpanel 4.0.0.
While:
  Installing instance.
  Getting distribution for 'plone.app.controlpanel'.
Error: Picked: plone.app.controlpanel = 4.0.0
```

sneridagh:
Then we have a problem because we have to make it conditional only for 5.2. And then, in ZCML we have to lookup in two different places depending on if it's 5.2 or not. Do we have some way to look that up?

mauritsvanrees:
We can use `zcml:condition="have plone-52"`. See for example [`plone.app.upgrade`](https://github.com/plone/plone.app.upgrade/blob/2.0.22/plone/app/upgrade/configure.zcml#L11-L13).

But wouldn't `zcml:condition="installed plone.app.controlpanel"` be enough? So something like this (untested):

```
   <include package="plone.app.controlpanel"
        file="permissions.zcml"
        zcml:condition="installed plone.app.controlpanel"/>
 ```

sneridagh:
@mauritsvanrees Fixed in #697, waiting for merge

mauritsvanrees:
Ah, I only saw a link to issue #696 above, and did not notice the PR #697 yet.
Approved and merged. Thanks!

PR #709, opened 2019-04-06
DX container site root

jaroel:


tisto:
@jaroel I merged the PR. Can we delete the branch or will that break your PLIP job? I can also make a p.restapi release if that helps...

jaroel:
Thanks for merging!

Oh, given that my PLIP work isn't green it really doesn't matter much :D
I'm sure a p.restapi release is done before this really becomes an issue for me. I'll delete the branch and adjust the PLIP buildout thingies.

PR #712, opened 2019-04-19
Return widget info for boolean fields

MrTango:


MrTango:

tisto:
@MrTango can you elaborate on why you/we need to return the "widget" value in addition to the "type" value we already provide? I mean the frontend can render a boolean widget based on the "type" value. This is what we do in Volto.

MrTango:
@tisto maybe talk to @robgietema he needs to do a lot of mapping on the client side, because of this missing information. The field type is not always enougth.

tisto:
@MrTango I talk to @robgietema on a daily basis and I still don't understand why we need to duplicate the information. There is no other option for boolean fields, so what additional information will a "widget" field provide with the same value?

jensens:

PR #716, opened 2019-04-21
Add query by token and translation of the returned title to the vocab…

sneridagh:
… endpoint

This is required in cases that the initial value is a token, and we need the title. We only can access them via the options, but might be that the options are batched, and we need to get the title anyways. We can't set it in the value, since we need to preserve the serializer-deserializer chain working.

e.g plone.app.vocabularies.Weekdays it returns:

```js
data: {
  available_timezones: ["Europe/Berlin"]
  first_weekday: 0
  portal_timezone: "Europe/Berlin"
}
```

So for the initial value we have "0" but no title whatsoever. In this case, we could find in the options again, but if the results are batched, might be that the token is not in the loaded options tokens.

### DIFF
--- a/src/plone/restapi/serializer/vocabularies.py
+++ b/src/plone/restapi/serializer/vocabularies.py
@@ -3,6 +3,7 @@ from plone.restapi.batching import HypermediaBatch
 from plone.restapi.interfaces import ISerializeToJson
 from zope.component import adapter
 from zope.component import getMultiAdapter
+from zope.i18n import translate
 from zope.interface import implementer
 from zope.interface import Interface
 from zope.schema.interfaces import ITitledTokenizedTerm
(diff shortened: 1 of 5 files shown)


sneridagh:
@tisto this one is ready, contains some required additions to the original PR

tisto:
@sneridagh we only find exact matches of the vocabulary title or token, correct? Maybe we should use title/token instead of query/token. What do you think?

Another idea would be to issue a warning when an API consumer uses both params at the same time. Or make the API return an error code in that case.

sneridagh:
For title we are doing partial matches, and for token as well, although it’s true that we should do the last one exact.

+1 for changing the params to title, token and making them more explicit exclusive.

PR #722, opened 2019-04-25
Fail when renewing invalid authentication tokens.

### LINKED ISSUE #721: ``@login-renew`` doesn't fail when token has already been expired.
When I try to renew an invalid or expired token with ``@login-renew`` the endpoint returns a new token.
IMO it should fail with a ``401`` error.

thet:
When renewing an expired or invalid authentication token with ``@login-renew`` fail with a ``401`` error instead of returning a new authentication token.

Fixes #721

### DIFF
--- a/src/plone/restapi/services/auth/renew.py
+++ b/src/plone/restapi/services/auth/renew.py
@@ -32,6 +32,12 @@ class Renew(Service):
                          plone.protect.interfaces.IDisableCSRFProtection)
 
         mtool = getToolByName(self.context, 'portal_membership')
+        if (bool(mtool.isAnonymousUser())):
+            # Don't generate authentication tokens for anonymous users.
+            self.request.response.setStatus(401)
+            return dict(error=dict(
+                type='Invalid authentication token',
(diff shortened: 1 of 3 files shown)


thet:

tisto:
@buchi would you mind having a look at this PR?

src/plone/restapi/services/auth/renew.py
```diff
@@ -32,6 +32,12 @@ def reply(self):
                          plone.protect.interfaces.IDisableCSRFProtection)
 
         mtool = getToolByName(self.context, 'portal_membership')
+        if (bool(mtool.isAnonymousUser())):
+            # Don't generate authentication tokens for anonymous users.
+            self.request.response.setStatus(401)
+            return dict(error=dict(
+                type='Invalid authentication token',
```
tisto: @thet should we say here "Invalid or expired authentication token" and in the message "The provided authentication token is invalid or expired."

tisto: @thet FYI: I fixed those minor issues myself in master.

PR #735, opened 2019-05-24
Add system endpoint

tisto:


base.cfg
```diff
@@ -99,6 +99,7 @@ mode = 755
 recipe = plone.recipe.codeanalysis
 directory = ${buildout:directory}/src
 flake8-exclude=bootstrap.py,bootstrap-buildout.py,docs,bin,*.egg,setup.py,overrides
+flake8-max-line-length = 88
```
jensens: Black has a soft line length of 88, but lines may be longer (but iirc respect the 120 max). So this may produce false reports.

src/plone/restapi/services/system/configure.zcml
```diff
@@ -0,0 +1,13 @@
+<configure
+    xmlns="http://namespaces.zope.org/zope"
+    xmlns:plone="http://namespaces.plone.org/plone">
+
+  <plone:service
+    method="GET"
+    for="zope.interface.Interface"
```
jensens: Wasn't it IRequest where we usually bound to?

src/plone/restapi/services/system/configure.zcml
```diff
@@ -0,0 +1,13 @@
...
+    factory=".get.SystemGet"
+    name="@system"
+    permission="zope2.View"
```
jensens: I am sure we do not want to expose this on view permission level. Use same permission as the view, see https://github.com/plone/Products.CMFPlone/blob/d1fcfd07972b65a8aeeb6a2b9f0f0f0da378c310/Products/CMFPlone/controlpanel/browser/configure.zcml#L26

tisto:
@jensens thank you for the review! Good catches! Will work on fixing them...

src/plone/restapi/services/system/configure.zcml
tisto: Fixed

src/plone/restapi/services/system/configure.zcml
tisto: Fixed

base.cfg
tisto: Not sure why this was present in this PR. Anyways, we set the line length in master to 200 to not interfere with black...

tisto:

tisto:

tisto:

Issue #745, opened 2019-06-06
POSException.ReadConflictError on 5.1


Rotonen:
Currently the tests get stuck on a test teardown on 5.1.

https://github.com/plone/plone.restapi/pull/471
https://github.com/plone/plone.restapi/issues/470

There is something subtle going on with time freezing and that breaking MVCC on the exact combinations of circumstances. This does not currently go boom on the main 5.1 build.

A circumvention should be sought.

PR #746, opened 2019-06-06
Circumvent ZODB some read conflicts arising from (hopefully) time freezing

Rotonen:
I noticed it's only the time frozen test layers causing issues and that one is luckily freezing time without any reason to do so. I simply stopped freezing time and moved the tests to other layers.

Seems the ZODB DemoStorage on 5.1 somehow gets angry at using (at least) `.invokeFactory()` or `plone.api.user` in a time frozen context and loses(?) the object (or connection?) after the first access to it.

Closes https://github.com/plone/plone.restapi/issues/745

I'm not very fond of this, as this still leaves me with a nagging feeling of whether or not the ZODB stack on 5.1 actually works in regards to MVCC.

https://github.com/zopefoundation/ZODB/issues/139

The worst case is this did not go well or has brought forth silently assumed operational conditions, which are not in place for 5.1.

### DIFF
--- a/news/745.bugfix
+++ b/news/745.bugfix
@@ -1,0 +1,2 @@
+Circumvent ZODB some read conflicts arising from (hopefully) time freezing.
+[Rotonen]
--- a/src/plone/restapi/testing.py
+++ b/src/plone/restapi/testing.py
@@ -27,6 +27,7 @@ from Products.CMFCore.utils import getToolByName
 from requests.exceptions import ConnectionError
 from six.moves.urllib.parse import urljoin
 from six.moves.urllib.parse import urlparse
+from time import time
 from zope.component import getGlobalSiteManager
 from zope.component import getUtility
 from zope.configuration import xmlconfig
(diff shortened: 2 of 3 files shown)


Rotonen:
What worked for me locally, gets hung up on the CI, I'll try to avoid more freezing.

tisto:
@lukasgraf aren't we using freeze in our documentation tests to make sure that we can detect outdated docs? If I recall correctly this was the reason we added freeze in the first place, correct?

Rotonen:
@tisto if so, that'd be for not having to bother with time stamps in documentation deltas. Nothing seems to trip up when it produces a delta on disk. It's always changed all of the `.resp` files on disk for me and never tripped up on that.

It is thus likely I'm breaking the delta simplification here.

lukasgraf:
@tisto exactly.

Rotonen:
For anyone looking into this later, you get a better view and the interpreter does not lock up if you patch the `getattr()` here to be a `hasattr()` (or patch it out all together):

https://github.com/plone/plone.app.testing/blob/188b3fdff6f3b00782c8e8efe3be7dbc97966b7a/src/plone/app/testing/layers.py#L366

tisto:
@Rotonen thank you for looking into this. That is highly appreciated! Freeze caused problems multiple times already when bleeding into other tests and stuff. Though, we kind of need the functionality for the test documentation. Maybe there are other ways to accomplish the same results with a less intrusive approach.

lukasgraf:
I agree.
@Rotonen even though this specific approach proved to be a bit trigger-happy 😉, I do appreciate you looking into this. The Plone 5.1 test failures are a real issue for the upcoming sprint.

@tisto a less intrusive approach could be to canonicalize any dates in dumped responses, so attack the issue with dynamic timestamps at the level of the `test_documentation` mechanism. This would involve some clunky pre/postprocessing of the payloads getting dumped, and add some overhead effort for future tests, but it should be possible to keep it conceptually simple (so that new contributors can understand what's going on, and craft new tests themselves).

But what I don't like about that approach is that it sort of masks the actual problem. Freezing time *should* work, in my opinion. We've been using that pattern for years now, in literally hundreds of tests, and have never quite run into issues like this.

So, when I first looked into this last year, it looked like the main issue of the interaction of time freezing and ZODB persistence was the fact that our way of freezing time violated the inherent assumption in ZODB's transaction IDs that time moves monotonically (always increases). However, back then when I changed our freezing setup to ensure this, somehow that didn't solve the problem.

But now that @sunew fixed a teardown related issue during last sprint, maybe it's now the lack of monotonic time that surfaces. So we could possibly look into that again.

Rotonen:
https://github.com/plone/plone.restapi/blob/1dfae340368a24a3ff9043ee382dff2f0ae2cf0e/src/plone/restapi/testing.py#L130-L139

The implementation looks like it'd tick forwards. But when putting a pdb there, it tends to return the same timestamp for every time you end up there, suggesting the test teardown / env teardown (where it also sometimes goes boom), also tears that utility down. Thus you'd always hit the first branch?

tisto:
@lukasgraf I couldn't agree more. Time freezing is an important feature in testing and it should work.

PR #750, opened 2019-06-13
types endpoint: Fix support for context aware default factories

### LINKED ISSUE #748: Defaults calculated in types endpoint don't respect context aware default factories
In `zope.schema` fields, one can use **context aware default factories** - callables that get passed the field's context and may use it to return a context dependent default value.

For example:

```python
from zope.schema.interfaces import IContextAwareDefaultFactory
from zope import schema

@provider(IContextAwareDefaultFactory)
def uppercased_title_default(context):
    return context.title.upper()

field = schema.TextLine(
    title=u"My field", description=u"My great field",
    defaultFactory=uppercased_title_default
)
```

In order for this to work however, the field needs to be bound to a context before accessing `field.default`, [which the `DefaultJsonSchemaProvider` currently doesn't do](https://github.com/plone/plone.restapi/blob/1dfae340368a24a3ff9043ee382dff2f0ae2cf0e/src/plone/restapi/types/adapters.py#L41).

lukasgraf:
A `defaultFactory` might be context aware (if it implements `IContextAwareDefaultFactory`).

However, in order for these context aware default factories to be invoked with the correct context, the field needs to be bound first before attempting to access `field.default`. The `DefaultJsonSchemaProvider` used in the `@types` endpoint currently doesn't do this, and therefore always invokes `field.default` on unbound fields.

This change makes sure the `DefaultJsonSchemaProvider` (and providers inheriting from it) always operates on bound fields.

Fixes #748

### DIFF
--- a/news/748.bugfix
+++ b/news/748.bugfix
@@ -1,0 +1,2 @@
+@types endpoint: Fix support for context aware default factories.
+[lgraf]
--- a/src/plone/restapi/tests/test_types.py
+++ b/src/plone/restapi/tests/test_types.py
@@ -16,6 +16,8 @@ from unittest import TestCase
 from z3c.form.browser.text import TextWidget
 from zope import schema
 from zope.component import getMultiAdapter
+from zope.interface import provider
+from zope.schema.interfaces import IContextAwareDefaultFactory
 from zope.schema.vocabulary import SimpleTerm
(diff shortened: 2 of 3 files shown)


lukasgraf:

tisto:
Thank you @lukasgraf!

PR #752, opened 2019-06-15
Fix PATCH on @user properties if the portrait is empty, returns 400

sneridagh:


sneridagh:

buchi:
@sneridagh I've tested this on Plone 5.2, 5.1 and 4.3 and for me it seems to work.
Not sure what exactly your issue was.

A user portrait can be deleted by providing `null`: 
```
{
    "portrait": null
}
```

To keep the current portrait, omit `portrait` in the request.

tisto:
@buchi @sneridagh this kind of raises the general question of how we set null values in plone.restapi. I guess we have to check how we do this on other endpoints and then document that properly for the future.

PR #758, opened 2019-06-21
Improve documentation for how to set relations

### LINKED ISSUE #732: How to set relatedItems?
The documentation for setting a relation  https://plonerestapi.readthedocs.io/en/latest/serialization.html#relations is a bit short.

On the source side we have  a list of UIDs (from AT content) and need to set them as `relatedItems` on the related resource. Source and target system use the same UIDs. 

What is the correct data format for settings UIDs? Any example?

buchi:
Closes #732


buchi:

PR #759, opened 2019-06-21
Add @querystring endpoint that dumps p.a.querystring config

### LINKED ISSUE #754: Add @querystring
For the querystring widget used in the collections we will need a new endpoint. See docs at: https://github.com/plone/plone.restapi/blob/add-querystring/docs/source/querystring.rst

lukasgraf:
The `@querystring` endpoint returns the querystring config of `plone.app.querystring`.

Instead of simply exposing the querystring related field and operation entries from the registry, it serializes them the same way the `@@querybuilderjsonconfig` from `p.a.querystring` does.

This form is structured in a more conventient way for frontends to process:

- **Vocabularies** will be resolved, and their values will be inlined in the values property
- **Operations** will be inlined as well. The operations property will contain the list of operations (dotted names), and the operators property will contain the full definition of each of those operations supported by that field.
- Indexes that are flagged as **sortable** are listed in a dedicated top-level property `sortable_indexes`

See [Docs build of the documentation / example response](https://plonerestapi.readthedocs.io/en/add-querystring/querystring.html)

Closes #754


lukasgraf:
Thanks! Updated & force pushed 👍

lukasgraf:

lukasgraf:

bloodbare:
@tisto We not going to support querystring at guillotina project, we already discussed to have a search endpoint that supports all kind of indexes and operations.

https://docs.google.com/document/d/1xooubzJKBnUzVlsa2f0GSwvuE1oir9hUdWUf1SU9S6E/edit?usp=sharing

bloodbare:
And we have a collection that that supports the search format to store the query.

sneridagh:
Just tested it with the corresponding Volto branch and it works well.

Only one question: For the vocabularies we used "the old way", like just include them in the response. This is different on how we implemented it this spring, so we provide a source URL for retrieving the values. I wonder if we might find the same culprits we found in the vocabularies widgets. e.g. Users vocabulary, and if you use a custom criteria with a custom vocabulary that returns an insane amount of results.

This might complicate the widget, but it's something that we might think of upfront.

/cc @tisto @lukasgraf @robgietema @buchi

tisto:
@plone/restapi-team opinions on @sneridagh's post above? Otherwise I'd merge and release this now since it is needed for the new collection type in Volto 4.

lukasgraf:
@sneridagh sorry, I was on holidays and didn't read my mails for a while.
I don't quite understand your question though: Vocabulary contents aren't (shouldn't) be inlined in the response of the `@querystring` endpoint - only the vocabulary *name* is included in the response. Or did I misunderstand what you mean?

sneridagh:
@lukasgraf oh ok since in your first comment in this thread you said: `Vocabularies will be resolved, and their values will be inlined in the values property` I thought they were inlined and I didn't check them out. So that's settled, sorry for the confusion!

lukasgraf:
@sneridagh oh, now I see what you mean, you're right.

That is indeed a bit inconsistent, and could potentially be undesirable for large custom vocabs.

We already have one other inconsistency like that though: Sources (instead of vocabularies) currently get inlined in the `@types` endpoint response. They have to be, because there's no `@sources` endpoint yet :)

So, I'm currently working on that anyway (new `@sources` and `@querysources` endpoints). I would recommend that we first add these new endpoints and link to them (non-breaking change), and at a later point stop inlining vocab values everywhere, at least by default (breaking change).

Issue #760, opened 2019-06-21
Effective Date not set properly on workflow transition


wkbkhard:
Fix for:
https://github.com/plone/volto/issues/761

buchi:
Fixed in #761

PR #761, opened 2019-06-21
Effective date not set

wkbkhard:
https://github.com/plone/volto/issues/761
#760


src/plone/restapi/services/workflow/transition.py
```diff
@@ -16,6 +16,7 @@
 from zope.interface import implementer
 from zope.publisher.interfaces import IPublishTraverse
 from zope.publisher.interfaces import NotFound
+from DateTime import DateTime
```
buchi: Ordering of imports does not conform to the Plone style guide (https://docs.plone.org/develop/styleguide/python.html#grouping-and-sorting). But that was already the case before that change.

tisto: I will look into this...

tisto:
@jenkins-plone-org please run jobs.

PR #762, opened 2019-06-21
Fix flake8 plugins

tisto:


tisto:
@buchi @lukasgraf the way black formats imports is incompatible with isort it seems: https://github.com/timothycrosley/isort/issues/694 cc @gforcada

base.cfg
```diff
@@ -100,6 +100,9 @@ recipe = plone.recipe.codeanalysis
 directory = ${buildout:directory}/src
 flake8-exclude=bootstrap.py,bootstrap-buildout.py,docs,bin,*.egg,setup.py,overrides
 flake8-max-complexity = 25
+# flake8-ignore = E203, E266, E501, W503
```
idgserpro: Would you mind telling the motivation of removing the .flake8 file to just commiting a comment here in a base.cfg?

We're following black adoption in plone core to make a well thought decision on how to use in internal packages, but we got confused about this commit.

PR #776, opened 2019-06-24
Fix 204 no content

### LINKED ISSUE #775: Responses with status 204 (No Content) shouldn't return content
Currently various services (e.g. DELETE, PATCH) return `None` which is serialized to `null`.
In the case of 204 the response body should be omitted.

buchi:
Closes #775

### DIFF
--- a/news/775.bugfix
+++ b/news/775.bugfix
@@ -1,0 +1,3 @@
+Return empty response for status 204 (No Content).
+[buchi]
+
--- a/src/plone/restapi/services/__init__.py
+++ b/src/plone/restapi/services/__init__.py
@@ -35,3 +35,7 @@ class Service(RestService):
            the no content marker if the response body should be empty.
         """
         return _no_content_marker
+
+    def reply_no_content(self, status=204):
+        self.request.response.setStatus(status)
+        return _no_content_marker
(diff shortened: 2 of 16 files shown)


buchi:

tisto:
@lukasgraf @buchi I am wondering if that is a breaking change. We are technically changing the API. Though, that means we have to bump a major release and write an upgrade guide section.

lukasgraf:
I would consider this a bugfix.
`204 No Content` was never *supposed* to return any body content. That was always the intended behavior, and some inconsistencies regarding that have now been fixed.

Yeah, we're technically changing the API. But if we're setting the bar that low for what we consider breaking, then we're going to have a painful road ahead of us. Because if you look at the concept of a "breaking change" with a narrow enough set of eyes, you can see it everywhere:

Fixed an unhandled exception? Breaking change, because the endpoint now doesn't return 500 with the serialized exception any more. If a client were to rely on this exact behavior, it would break.

And it starts getting absurd at that point. I very much appreciate us being rather strict about breaking changes and semantic versioning, but there's always going to be situations where we have to use our best judgment. And in this case my gut says it's quite unlikely that someone is relying on that particular behavior (or at least it would be quite [ill-advised](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204)).

PR #778, opened 2019-06-25
Return status 400 if a referenced object can not be resolved during deserialization

### LINKED ISSUE #777: Message for referencing object that can't be resolved should be returned with 400 instead of 500
With plone/plone.restapi#756 the returned status code (when a object is referenced that can't be resolved) is now `500`. It should instead be `400` since it's a client error.

lukasgraf:
Return status `400` (instead of `500`)  if a referenced object can not be resolved during deserialization since it's a client error.

Fixes #777.

### DIFF
--- a/news/777.bugfix
+++ b/news/777.bugfix
@@ -1,0 +1,2 @@
+Return status 400 if a referenced object can not be resolved during deserialization.
+[lgraf]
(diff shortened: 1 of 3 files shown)


lukasgraf:

PR #779, opened 2019-07-05
Docs: Add 'How to write a custom service' in customization.rst

mbarde:
"How to write a custom service" by @MrTango 
Belongs here imho


MrTango:
You are right, didn't have the time so far. Note that you can use the plonecli/bobtemplates.plone to create a service ;)

tisto:
@mbarde thank you for your contribution. From past experiences I'd say it would be better if we copy over the howto to our docs instead of just adding a link that might stop working at some point. Adding this to our docs also allows people to improve or amend the docs if things change over time. We just need a volunteer for that task. :)

Issue #780, opened 2019-07-10
@sharing does not work on the plone site root


csenger:
```
2019-07-10 09:27:11 ERROR Zope.SiteErrorLog 1562743631.070.120943757926 http://localhost:8080/Plone/@sharing
Traceback (innermost last):
  Module ZPublisher.Publish, line 138, in publish
  Module ZPublisher.mapply, line 77, in mapply
  Module ZPublisher.Publish, line 48, in call_object
  Module plone.rest.service, line 23, in __call__
  Module plone.restapi.services, line 21, in render
  Module plone.restapi.services.content.sharing, line 60, in reply
  Module plone.restapi.deserializer.local_roles, line 58, in __call__
TypeError: ('Could not adapt', <PloneSite at /Plone>, <InterfaceClass Products.CMFCore.interfaces._content.ICatalogAware>)
```
The plone @@sharing view can be used on the site root and is calling .reindexObjectSecurity without checks for interfaces.

jensens:
this one was fixed with #781

PR #781, opened 2019-07-10
sharing on site root

csenger:
With this change @sharing works on the site root like @@sharing does. Local roles are set correctly.

Though PloneSite.reindexObjectSecurity does nothing so catalog search results for allowedRolesAndUsers are still wrong. But that is the same bug/feature behavior plain Plone's @@sharing has.

See #780 (branch is named wrong)


tisto:

PR #787, opened 2019-08-20
Feature relations service

ukretschmer:
I added a simple expandable service `@relations`, which adapts objects implementing `IHasRelations` and by default renders incoming relations (summary).

It is also possible to render incoming and outgoing relations, by bypassing the request param `allrelations`.

### DIFF
--- a/plone-5.2.x.cfg
+++ b/plone-5.2.x.cfg
@@ -1,7 +1,7 @@
 [buildout]
 extends =
     base.cfg
-    http://dist.plone.org/release/5.2rc3/versions.cfg
+    http://dist.plone.org/release/5.2/versions.cfg
 find-links += http://dist.plone.org/thirdparty/
 versions=versions
 
--- a/src/plone/restapi/services/configure.zcml
+++ b/src/plone/restapi/services/configure.zcml
@@ -20,6 +20,7 @@
   <include package=".locking" />
   <include package=".principals"/>
   <include package=".registry"/>
+  <include package=".relations"/>
   <include package=".roles"/>
   <include package=".search"/>
   <include package=".types"/>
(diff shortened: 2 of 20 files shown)


tisto:
@ukretschmer thank you for your contribution! This is much appreciated! 

I just had a very brief look at your PR so far. If I understand this correctly your use case is that you want to show all "inbound" relations that point to the current object. Would you mind elaborating how you are using this plone.restapi feature? Is that for link integrity checking or a specific use case you have?

ukretschmer:
@tisto thank you for your response and the questions, which I'd like to answer.

We are currently developing a SPA based on Vue and Vuetify as frontend for our virtual archive and use Plone as backend. We have highliy interrelated content types and want to make explicit and strong use of those relations between the different objects. At the moment it is only possible to act with relations pointing from a specific object to others, e.g. showing persons as authors from a publication. But if I want to list all publications of an author (backlinks), there is no way to do that trivially.
This also allows to lookup and use the complex network graph of relations and finally use those data for visualization, e.g. a network of persons with different types of relations (friend, father, coworker and so on).

I'm not just interested in the incoming relations, but I made this the default case, to not overload the default output. Some notes to the implementation:

1.The service is added as expandable component, so it doesn't change the actual behaviour of the API response, there is just a new component `@relations` listed. 
2. If the component is expanded, via `expand` param or by requesting `.../@relations`, it will list all objects directly pointing to the current context object (incoming relations). Outgoing relations, that is pointing from context object to others, will not be listed by the service, because this is usually done by the object serialization itself (attributes with relations, serialize to a summary of the target object).
3. By adding the request param `allrelations`, the service also lists all outgoing relations. Usecase here is to access `@relations` to get all relation information without the need of serializing the whole context object or to get attributes using relations directly accessable, without the need to know, that they are relations.
4. Related objects are grouped by the source attribute of the relations.

Our usecase is to make our highly interrelated objects accessable (linking) and by that explorable along their relations. By that, we offer a third "navigation" possibility.

1. Traverse through object hierarchy (static)
2. Search (dynamic, but you need to know, what you are looking for)
3. Relation network (dynamic, start at any object and discover (all) others by navigating through the graph)

tisto:

davisagli:
Closing in favor of #1431 which is more complete and has a signed contributor agreement

Issue #535, opened 2018-06-19
Complete the experience when using choices/lists tied to vocabulary/sourceCatalog fields


sneridagh:
Currently choices or list-choices fields are returning all the possible values in the content type JSON Schema representation, and that could lead to returning big datasets that the browser have to deal with (and probably fail to do so, hanging up).

Instead, we have to find out to provide a better experience with those, in which could be easily integrated with an autocomplete widget that would query to the original vocabulary, filtering the results on the fly.

Also, the sourceCatalog vocabulary field implementation is missing. Given the fact that is the recommended way to go when dealing with content in vocabularies (plone.app.vocabularies.Catalog), it's sensible to give support to them.

It might be needed to discuss how to implement that. Also, the UI can't alone determine which Component have to use for the widget (e.g subjects (Tags) field). We have to deliver some sensible defaults for the most common field-widget combinations as we did for z3c.form, but it would be great to be able to fine grain specify the type of the widget (in the UI or in Plone itself, when creating the field). More like

```
    directives.widget(
        'subjects',
        AjaxSelectFieldWidget,
        vocabulary='plone.app.vocabularies.Keywords'
    )
```

but thinking specifically on the UI part.

davisagli:
The way that this currently works with mockup is:

1. The AjaxSelectWidget pattern options have `vocabularyUrl` set to an endpoint for looking up vocabulary terms. For fields with a vocabulary this is `[form path]/@@getVocabulary?field=[field name]` (so that permission to access the vocabulary can be checked based on the form context). For fields with a source this is `[widget path]/@@getSource`
2. The select2 and relateditems widgets in mockup call these endpoints, passing either a string or a querystring-formatted query.
3. The getVocabulary and getSource views are implemented in plone.app.content. They handle a few different things:
    - Get the vocabulary.
    - Check permission using either a IFieldPermissionChecker adapter or a hardcoded mapping of vocab name to permission.
    - Prepare a catalog query. The query is found in the `query` request parameter, which can be either a string or JSON including a `criteria` querystring, `sort_on`, and `sort_order`.
    - Search the vocabulary:
        - If the vocabulary factory takes a `query` parameter, call it passing the query.
        - If the vocabulary has a `search_catalog` method, call it passing the query.
        - If the vocabulary has a `search` method, call it with the SearchableText from the query.
        - Otherwise (or if no query) iterate through all terms.
    - Batch results.
[...]

Issue #753, opened 2019-06-20
@types: Distinguishing add and edit contexts


lukasgraf:
In Plone, both vocabularies as well as default values may be context-sensitive.

An `IVocabularyFactory` (or an `IContextSourceBinder`) will be passed a context, as will an `IContextAwareDefaultFactory`.

Now, depending on whether the user is creating a new object, or editing an existing one, these might get passed different contexts.

In classic Plone, the way this pans out is that when

- **adding**, the context will be the **container** to which the object will be added
- **editing**, the context is the **object itself**

In other words:
Because the `++add++ ...` view is invoked on the container that the object will be added to, a `IVocabularyFactory` will be passed the parent of the object that is about to be created, and is expected to produce the vocabulary for a particular field *for* that newly created object, *based on* information from its parent.

The edit case is much simpler: The object already exists, and the item itself is the context that will be passed to the `IVocabularyFactory`.

This however leads to **ambiguity**: The `IVocabularyFactory` has no frame of reference to determine  whether the context it's looking at is the item itself, or the container of the item in question.

It could do `container = aq_parent(context)` - but that would be wrong during the add form. Or it could do `container = context`, but that's wrong in the edit form.

This issue only crops up in quite advanced uses of context-aware vocabularies and default factories. In our application, we have so far addressed this by hooking into all the different ways to create content in Plone (plone.api, invokeFactory / createContentInContainer, transmogrifier, REST API, Forms), and adding a `IDuringContentCreation` marker interface to the request.

That way, the vocabulary factories / default factories that need to know this can do

```python
if IDuringContentCreation.providedBy(request):
    # Add form
    container = context
else:
    # Edit form
    container = aq_parent(context)
```

In `plone.restapi` we now face the same issue. The `@types/:portal_type:` endpoint is likely to be used to both render edit as well as add forms. In the most recent version of plone.restapi a field's vocabulary will contain a pointer to a context-sensitive URL to the `@vocabulary` endpoint for that field's vocabulary. And there we face the same issue:

Is the resource that `@vocabulary` is invoked on both the target item and the source context? (edit form). Or is it just the source context, and the target item doesn't yet exist? (add form)

The `@types` endpoint is also a bit awkward to use in the edit form case: It needs to be invoked with a `portal_type` path parameter by the frontend, even though the context's portal_type is very much known to the backend, and completely unambigous.

This possibly could address both these issues by:

- Introducing a `@type` endpoint (singular):
[...]

PR #791, opened 2019-09-08
Add @sources and @querysources endpoints

### LINKED ISSUE #790: Support sources and querysources
_(Placeholder issue for towncrier)_

Add proper support for enumerating / querying terms for fields with an`ISource` or `IQuerySource`.

lukasgraf:
This PR introduces two new endpoints,
- `@sources` and
- `@querysources`

and wires up links to them in the `@types` endpoint response in a **backwards-compatible** way.

## Status Quo

Currently (before this change) we have a `@vocabularies` endpoint, that enumerates terms for **named vocabularies**. This endpoint is linked to from the JSON schema in a `@types` response.

If a field is backed by an `ISource` that is not a vocabulary however, possible choices are inlined in the JSON schema (because we have no corresponding endpoint yet to link to). Querying of `IQuerySources` isn't supported at all yet.

## New endpoints @sources and @querysources

After quite a bit of back and forth, this change reflects what I would currently propose as the way to start adding support for these vocabulary-like concepts.

_(Please see the [updated docs on Vocabularies and Sources](https://plonerestapi.readthedocs.io/en/querysources/vocabularies.html) for a less abstract overview, as well as the [updated note at the very bottem of the @types docs](https://plonerestapi.readthedocs.io/en/querysources/types.html))_

With this change, we would then have three endpoints:
- `@vocabularies/<vocab_name>` - addresses vocabularies via their **global name**, enumerates terms
- `@sources/<field_name>` - addresses sources via the **field name**, enumerates terms
- `@querysources/<field_name>?query=<search_term>` - addresses `IQuerySources` via the **field name**, requires a `?query=` parameter, doesn't allow enumeration of all terms.

The rationale behind three different endpoints is that they fundamentally differ in how the vocab-like set is addressed: Either via a global name, which only vocabularies have, or via the field that the source is bound to. The `@querysource` endpoint also differs from the other two in that it _can't_ be invoked without a query (to just enumerate all terms), but instead needs to be supplied with a query term. This makes its usage different enough that it makes sense to me to not cram everything into one endpoint. Also see [David's comment (point 4. at the very bottom)](https://github.com/plone/plone.restapi/issues/535#issuecomment-399669706)

These three endpoints then get wired up in the JSON schema of the `@types` endpoint as follows:
- If a field is backed by an `IQuerySource`, a property `querysource` will be included, which contains an URL that points to the `@querysource` endpoint with the appropriate field name. That endpoint **must** be queried by the client by supplying a `?query=` parameter - full enumeration for terms is not supported. The client (frontend) therefore needs to include the necessary logic in the widget, and treat fields that include a `querysource` property accordingly.
- If a field is backed by any other vocab-like (vocabulary or  iterable source), a property `vocabulary` is included, with an URL to either the `@vocabularies` or `@sources` endpoint. These two endpoints have the exact same semantics:
  - They enumerate all terms
  - They are batched, if necessary
  - They support simple filtering (via `title` or `token`)
  For a client application it therefore doesn't matter, to which endpoint the URL in the `vocabulary` property points - it's supposed to treat them both exactly the same.

## Possible future changes

This PR doesn't remove the inlining of choices in the JSON schema for sources that are iterable. It keeps this behavior, in addition to the link to the `@sources` endpoint, for backwards compatibility.

I would eventually like to get rid of inlining these choices though. Especially sources tend to be quite dynamic and possibly large, so we could end up dumping large amounts data in these choices. We could at a later point introduce a mechanism similar to expansion for vocabulary terms:
 - New query string param for `@types` that takes a list of field names for which to inline vocab terms, or  some kind of marker to inline all fields.
 - Defaults to inlining all (sources) first, until we decide to do a breaking change

Closes #790

@plone/restapi-team


lukasgraf:

lukasgraf:

lukasgraf:
@tisto Plone tests are currently failing because of `IOError: [Errno 28] No space left on device` (`Node1`, see [this failure](https://jenkins.plone.org/job/pull-request-5.2/516/console) for example)

lukasgraf:
I'll rebase this one once the dust on #795 settles.

lukasgraf:

lukasgraf:
@jensens thanks for your review! 💚 
Rebased onto `master`, and fixed `vocabularies.rst` and `types.rst` according to the Documentation Styleguide (one sentence per line, no hard line wraps).

sneridagh:
@lukasgraf I just tested it in Volto, and it breaks the edit view because of now the allow_discussion field is exposing a vocabulary.

```
allow_discussion: {choices: [["True", "Ja"], ["False", "Nein"]],…}
choices: [["True", "Ja"], ["False", "Nein"]]
description: "Erlaubt Diskussionsbeiträge für diesen Inhalt."
enum: ["True", "False"]
enumNames: ["Ja", "Nein"]
title: "Diskussion erlauben"
type: "string"
vocabulary: {@id: "http://localhost:8080/Plone/@sources/allow_discussion"}
@id: "http://localhost:8080/Plone/@sources/allow_discussion"
```

How is that possible?

sneridagh:
@lukasgraf btw, the `http://localhost:8080/Plone/@sources/allow_discussion` is answering a 404.

lukasgraf:
@sneridagh could you give me a bit more context on this? Edit view for what type? What endpoint's response did change for you in order to cause the breakage (I'm assuming `@types` for a particular object).

sneridagh:
@lukasgraf the `@types` call for the schema when editing an object, has the `allow_discussion` field. This field now is modeled differently, so it also has a "vocabulary" property, as shown in my first message.

Volto comes in, and tries to access that vocabulary `http://localhost:8080/Plone/@sources/allow_discussion` that does not exists and throws a 404.

The the edit view behaves badly.

lukasgraf:
@sneridagh I believe we're now facing the consequences of the issue I described in #753.

The schemas returned by the `@types` endpoint do not distinguish between **add** form and **edit** form scenarios. (The way `@types` currently is implemented suggests add semantics).

The `@sources` endpoint however currently only works for **edit** scenarios - because it needs to look up the field via the schema from a context, it needs to be invoked on an existing object:

sources_edit

In order for it to also work for **add** scenarios, it would need an alternative way of invoking it. Something along these lines:

sources_add

Then the schemas from the `@types` endpoint could be modified so that the vocabulary-URLs they return are properly build, including the `?portal_type` parameter to render these sources and vocabularies for add-scenarios.

sneridagh:
@lukasgraf the discussion behavior:

```
options = SimpleVocabulary([
    SimpleTerm(value=True, title=_(u'Yes')),
    SimpleTerm(value=False, title=_(u'No')),
])

...

    allow_discussion = schema.Choice(
        title=_(u'Allow discussion'),
        description=_(u'Allow discussion for this content object.'),
        vocabulary=options,
        required=False,
        default=None,
    )
```

This is a static vocabulary, for static vocabularies we were still returning them on the serialization. Did we change it?

sneridagh:
I'm guessing now, but I think that it might be a bug when the `vocabulary` property is set, and it is not checking if there is a named vocab assigned to the field or not. If a static vocab is set, then it shouldn't set it. Notice in the serialization, the full vocab enumeration are still there:

```allow_discussion: {
choices: [["True", "Ja"], ["False", "Nein"]],…}
choices: [["True", "Ja"], ["False", "Nein"]]
description: "Erlaubt Diskussionsbeiträge für diesen Inhalt."
enum: ["True", "False"]
enumNames: ["Ja", "Nein"]
title: "Diskussion erlauben"
type: "string"
vocabulary: {@id: "http://localhost:8080/Plone/@sources/allow_discussion"}
@id: "http://localhost:8080/Plone/@sources/allow_discussion"
}```

**Comment by lukasgraf** [comment]
@sneridagh the fact that the static choices are still there is intentional - exactly because I didn't want to break backwards compatibility.

What isn't intentional though is providing a `@sources` URL for a field with a static vocabulary. I'm currently looking into why this is happening.

**Comment by sneridagh** [comment]
@lukasgraf yeah, there's a decider in Volto how a vocabulary has to be rendered, if it has a `vocabulary` property, then it gets the URL and renders the response properly. Vocabularies has more precedence than static vocabs. That's why it's failing, tries to get the URL, it's not there, then the render fails.

**Comment by lukasgraf** [comment]
@sneridagh I just checked, the `vocabulary` URL is included because the fields vocabulary can actually be accessed as a source, even though it's static:
```python
(Pdb) self.field
<zope.schema._field.Choice object at 0x118545828>
(Pdb) self.field.getName()
'allow_discussion'
(Pdb) self.field.source
<zope.schema.vocabulary.SimpleVocabulary object at 0x1134466a0>
(Pdb)
```
(That's just how `zope.schema` works).

And the `@sources` endpoint would usually also work for iterable vocabularies. In this case it doesn't though, because the endpoint isn't registered on the Plone site root. And registering it wouldn't help, because without explicitly passing it a `portal_type` parameter or something like that, it would have no way of knowing what schema to look up `field_name` on.

lukasgraf:
@sneridagh I'm leaving for lunch now, but I'll keep thinking of a quick short term solution how we can fix this that it doesn't break Volto.

sneridagh:
@lukasgraf from the Volto part, we could investigate if we can change the decider and give more precedence to the choices above the vocabularies.

However, we will still have the fact that the URL is not accessible in some scenarios.
I still don't get why it's `http://localhost:8080/Plone/@sources/allow_discussion` relative to root, since the object being edited was `/Plone/whatever` ... Shouldn't it be `Plone/whatever/@sources/allow_discussion`?

lukasgraf:
@sneridagh as outlined in my linked issues or the comments above: There's no single URL that is correct.

The last URL you posted would be correct for an **edit** form for `/Plone/whatever`. It would however be incorrect for the add form. And currently the `@types` endpoint, implicitly, judging from the way its built, only really supports schemas for **add**.

Or in other words:
- For edit forms, the URLs should be relative to the `object`
- For add forms, the URLs need to be relative to the `container`
 
Relative to the container results in relative to root for top-level objects. So that URL is not technically incorrect (though incomplete), but it's simply not supported yet.

lukasgraf:
@sneridagh possibly we could also make attempts in the `@types` endpoint to only return Vocabulary-URLs that have been checked in some way (not sure how yet) to be valid.

But honestly, changing the preference order in Volto sounds like a good idea anyway to me: If `choices` / `enumNames` are already included in the response, it's quite wasteful to do an extra request anyway, no?

sneridagh:
@lukasgraf Yeah I know that seems more sensible to have the order of the precedence the other way around, but might be that it should be like this because of something (some weird field/widget combination that forced us to decide on this originally). Also, some months ago all the vocabs were serialized inline... might be that we can now safely swap their order in the decider.

I’ll take a look!

sneridagh:
Implemented this PR: https://github.com/plone/volto/pull/921/
for changing the order of the widget render decider.

PR #792, opened 2019-09-09
First working version

sneridagh:


lukasgraf:
I'm having a look at this right now - batching seems a bit strange at first glance.

sneridagh:
@lukasgraf I had to do this because the current batching is taking the params from the qs, not from the body, so it has to cope with the use case that the batching params are in the body.

sneridagh:
@lukasgraf there are also some changes merged from another branch that will become master soon. We are changing to default Plone 5.2 in p.restapi and we found something strange in the buildout that behaves differently in 5.2 and it didn't get the developed egg correctly.

lukasgraf:
Yeah, I saw. It seems to be based on #795. The `auto-checkout = plone.restapi` seems quite wrong on that one, and also causes issues for me locally, but I'll comment on that other PR later.

PR #795, opened 2019-09-09
Default to Plone 5.2

tisto:


tisto:
@sneridagh this branch still has the problem that it does not run the tests on the code in /src but the released version. Still don't understand what goes on. Seems this is a Plone 5.2 issue since I can reproduce the problem with py2 as well.

sneridagh:
@tisto estrange, I merged https://github.com/plone/plone.restapi/pull/795/commits/80f198f4f801e8d2a3531d5cb5e1bd03b72b36dd in my branch and it works well, all changes are reflected and taken into account.

sneridagh:
@tisto mmmm bin/test shows different numbers in travis than locally now:
locally: `Total: 948 tests`
3.7 + 5.2 travis: `Total: 944`.

sneridagh:
@tisto Forget about this last one, the tests are equal locally and in travis:
https://travis-ci.org/plone/plone.restapi/jobs/582740441#L594

sneridagh:
@tisto I added the ZSERVER_PORT fixing for documentation, and updated the documentation to Plone 5.2 as well.

base.cfg
```diff
@@ -19,9 +19,10 @@ parts =
     sphinx-python
     deploy-to-heroku
     omelette
-develop = .
-sources-dir = extras
+# develop = .
+sources-dir = src
 auto-checkout =
+    plone.restapi
```
lukasgraf: This doesn't seem right to me (and also causes issues when testing this locally, `mr.developer: Directory name for existing package 'plone.restapi' differs. Expected 'plone.restapi'.`). Why would we check out `plone.restapi` _again_ in a source checkout of that very package?

base.cfg
```diff
@@ -65,6 +66,7 @@ recipe = zc.recipe.testrunner
 eggs = ${instance:eggs}
 initialization =
     os.environ['TZ'] = 'UTC'
+    os.environ['ZSERVER_PORT'] = '55001'
```
lukasgraf: I see the need for a stable port in the documentation HTTP examples, but doesn't this cause issues with parallel runs on Jenkins when the Jenkins port allocator plugin isn't in full control of handing out ports?

sneridagh: @lukasgraf I also thought about it but then I realised that the coredev has its own buildout. Or am I wrong?

lukasgraf:
> @sneridagh this branch still has the problem that it does not run the tests on the code in /src but the released version. Still don't understand what goes on. Seems this is a Plone 5.2 issue since I can reproduce the problem with py2 as well.

@tisto I've looked into this. Seems like the version pin for `plone.restapi = 4.3.2` from the Plone 5.2 KGS is the issue (since plone.restapi is now included in core).

Resetting the version pin seems to do the trick (as in, checked out package gets developed properly, and is used for tests instead of the `4.3.2` egg):

```diff
diff --git a/plone-5.2.x.cfg b/plone-5.2.x.cfg
+[versions]
+plone.restapi =
```

(Tested with Py2 and Py3)

base.cfg
tisto: @lukasgraf tried to find the cause for the release. Reverted.

tisto:
@lukasgraf thanks for looking into it. This was indeed the cause.

tisto:
@lukasgraf @sneridagh I think this is ready to merge now.

base.cfg
tisto: @sneridagh I think this is already set in .travis.yml.

tisto: ok, we need this locally as well of course. Ignore my previous comment.

tisto:
Ok, seems we have a problem with the groups endpoint in python3. The results are not reproducible and change on every single run.

PR #805, opened 2019-09-20
Testing: Drop freezegun and instead selectively patch some timestamp accessors.

### LINKED ISSUE #803: Consider dropping freezegun
We've had some issues in the past, and recently I've seen testing problems crop up again that seem to be related to our use of `freezegun` in `plone.restapi`.

Specifically, I've seen builds (both on Travis and Plone Jenkins) fail because they got deadlocked (and eventually killed by the runner) that seem to be related to `freezegun`. From [this failure](https://travis-ci.org/plone/plone.restapi/jobs/583984273) here for example it looks like it's a threading issue involving some of [`freezegun`s bookkeeping data structures](https://github.com/spulec/freezegun/blob/d414467b7a9daf16e1f4810f5d08cb17fff66ad4/freezegun/api.py#L25-L45) (which aren't thread local, and therefore are subject to concurrency issues).

I will therefore prepare a PR that drops `freezegun` entirely, and replaces it with an approach that does much more selective patching (`created` and `modified` dates, mostly) of just what we need. I like the convenience of time freezing for tests, but if it's regularly causing issues that we need to invest time to solve, and its problems bleed through to the coredev buildout, it needs to go IMHO.

lukasgraf:
This drops the use of `freezegun` in our testing, and instead introduces a more selective approach to patch mainly `creation` and `modification` date accessors in order to get stable tests and dumped HTTP examples.

### Rationale

Unfortunately, `freezegun` has causes some issues in the past, and recently I've seen tests deadlocking again in a way that I would attribute to `freezegun` (see #803 for more details).

I like the convenience of the time freezing approach, but if it keeps causing issues and problems bleed through to the coredev buildout tests, it needs to go.

### Implementation

This PR introduces a `StaticTime` context manager (only used in `test_documentation` tests that dump HTTP responses). This context manager selectively patches `DexterityContent` in a way that DX object's `creation` and `modification` dates accessors return static times.

The same is also done for a couple other places where dynamic times are originating from other sources (p.a.discussion comments, workflow history, content history viewlet and lock infos). 

### Risks

Time freezing basically supplied us with stable and easy to control frozen time for the `test_documentation` tests once and for all. With this approach, there may be some work in the future where we need to extend the `StaticTime` helper to patch more things, depending on what functionality we introduce.

Closes #803


lukasgraf:

PR #809, opened 2019-10-05
Rename tiles to blocks.

tisto:
@sneridagh cc


sneridagh:
Some comments we might think about:

Behavior short naming: `plone.tiles` -> `plone.blocks` We might want to not tie this name to neither tiles nor blocks, since might cause confusion with the related technologies there. I suggest `plone.tiles` -> `volto.blocks`

I think we shouldn't rename @tiles endpoint, because it already does what it promises (on Mosaic tiles), we didn't use it in the final Volto Blocks implementation and I hardly believe Volto Blocks are going to have a backend side anytime soon. So I would leave it as it is.

Same for the documentation on that endpoint.

We should add proper docs for the Volto Blocks behavior, explaining the why and how.

We should not forget the upgrade step.

tisto:
@sneridagh ok, so we rename the ITiles behavior to volto.blocks, the tiles attribute to blocks and the tiles_layout to blocks_layout, correct?

We should also mark the tiles endpoint as deprecated for future releases. AFAIK there is no consumer for Mosaic via REST API.

PR #815, opened 2019-10-26
Fix issue with serialization of vocabulary items

### LINKED ISSUE #788: Serialization of multiple vocabulary entries leads to Server Error
https://github.com/plone/plone.restapi/blob/cbf3c07c94c8b9d4d624b6e31c4d21f9f3b94a6d/src/plone/restapi/serializer/dxfields.py#L71

**Error**

```
2019-08-24 19:06:54 ERROR Zope.SiteErrorLog 1566666414.130.724885414754 http://localhost:8080/Plone/test/GET_application_json_
Traceback (innermost last):
  Module ZPublisher.Publish, line 138, in publish
  Module ZPublisher.mapply, line 77, in mapply
  Module ZPublisher.Publish, line 48, in call_object
  Module plone.rest.service, line 23, in __call__
  Module plone.restapi.services, line 21, in render
  Module plone.restapi.services.content.get, line 18, in reply
  Module plone.restapi.serializer.dxcontent, line 125, in __call__
  Module plone.restapi.serializer.dxcontent, line 83, in __call__
  Module plone.restapi.serializer.dxfields, line 71, in __call__
TypeError: unhashable type: 'dict'
```

**Steps to reproduce**
Create a DX contenttype with a multi selection field, configure it to use a vocabulary for available options. Than create an object of that type with at least one chosen term.

**Problem**
At plone.restapi.serializer.dxfields, line 71 there is `value = self.field._type(values)` this code. I'm not sure, what the intention was what to do here, with this  `_type` function. IMO at this line values should just be assigned to value and all works well.

So replacing that line by `value = values` fixes the error. But as I said, not sure about the initial idea.

buchi:
Vocabulary items are not hashable and thus we can't cast them to the
fields type. We always return lists as JSON doesn't support other
container types anyway.

Closes #788

### DIFF
--- a/news/788.bugfix
+++ b/news/788.bugfix
@@ -1,0 +1,2 @@
+Fix serialization of vocabulary items for fields that need hashable items (e.g. sets).
+[buchi]
--- a/src/plone/restapi/serializer/dxfields.py
+++ b/src/plone/restapi/serializer/dxfields.py
@@ -68,7 +68,7 @@ class CollectionFieldSerializer(DefaultFieldSerializer):
             for v in value:
                 term = value_type.vocabulary.getTerm(v)
                 values.append({u"token": term.token, u"title": term.title})
-            value = self.field._type(values)
+            value = values
         return json_compatible(value)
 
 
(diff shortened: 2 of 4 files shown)


pbauer:
👍 Thanks, this fixes the issue in the Mastering Plone training.

sneridagh:
LGTM

PR #819, opened 2019-10-26
Remove IAPIRequest marker interface

fredvd:
It's not used in plone.restapi and should be imported from plone.rest.interfaces instead. If anybody was using this marker, it didn't do anything.

We coul alias/import it from plone.rest to here, but then it suddenly would start to work..... I don't know if that's better than removing it.

### DIFF
--- a/news/819.bugfix
+++ b/news/819.bugfix
@@ -1,0 +1,1 @@
+Remove IAPIRequest marker interface from plone.restapi. The correct interface should be imported from plone.rest.interfaces instead. If anybody was using this marker Interface, it didn't do anything.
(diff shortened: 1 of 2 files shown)


fredvd:

tisto:
Note: I hesitate to merge and release this because it could be considered to be a breaking change. Even though the chances that this is breaking for anybody are super low. Anyways, will leave this open for now to see if we have another breaking change coming up we can combine this with. I guess it is nothing that anybody really needs. Just a necessary cleanup that we should get in if possible.

jensens:
we have now "target" labels.

PR #824, opened 2019-10-31
Drop py3.6 and add py3.8 support.

tisto:


ale-rt:
I might be wrong but I do not think we can already drop Python 3.6

tisto:
@ale-rt right, we are still testing 3.6 on Jenkins so I will keep it...

PR #828, opened 2019-11-04
Prevent converting bytestring ids to unicode ids when reordering on python2.

### LINKED ISSUE #827: Reordering objects with python2 creates unicode ordering ids, should use bytestring ids.
When reordering objects in `plone.restapi` (as per https://plonerestapi.readthedocs.io/en/latest/content.html#reordering-sub-resources ) the ids reordered will be converted to unicode from a bytestring in python2. With  we use the `obj_id` from a request and thus most likely output by a `json.loads` to reorder(and also maybe `subset_ids`). The ids are unicode at that moment. The ids are then used to update ordering via the `plone.folder.interfaces.IExplicitOrdering` interface in:https://github.com/plone/plone.restapi/blob/3870e5b992e278e4ccd670a127455454b5f558d7/src/plone/restapi/deserializer/mixins.py#L36-L43
With python2 object ids in plone are an ascii bytestring. After updating the ordering via API the order which is kept in object annotations will contain mixed unicode and bytestrings. A call to `objectIds` will return a list with mixed types.

I have attempted to reproduced this with `plone5.2` and `python2` with the following steps (`python3` is not affected as it now uses unicode strings internally):

- create a folder with some children, e.g. pages in it
- reorder content via rest api, e.g. move one item down
```js
{
    "ordering": {"obj_id": "page3", "delta": 1}
}
```
- output `objectIds`, will now contain mixed types
```python
folder = plone.unrestrictedTraverse('folder')
folder.objectIds()
>>> ['page1', 'page2', 'page4', u'page3']
```
- See also failing tests in https://github.com/plone/plone.restapi/commit/f74a72afc7c4703b2a50f279ef9fa2304f2ddc1a

Impact:
- Calls to `objectIds` (and `keys` and `contentIds`) will return mixed types once content is reordered
- the object id itself and ids in the tree are not affected. Thus the path and catalog index is also not affected by default.

@tisto we have discussed this thing together with @lukasgraf at `ploneconf2019`. 

We need to write down a conclusion about the upgrade path that we want to take here. IIRC we discussed this and we tend to not upgrade anything by default as an upgrade would have to inspect all orderable content items which is very expensive. We expect that most content won't actually be affected and if it is that consequences are low.

For the fix, we are of the opinion those ids should be `ascii` instead. The ids used for ordering are later output in a call to `objectIds` where usually bytestrings are returned. The result of a call to `objectIds` may be used for traversal by the client where unicode paths segments will fail. Thus we should make sure that we perform input validation and use bytestring ids when calling methods of a `plone.folder.interfaces.IExplicitOrdering`.

deiferni:
This PR fixes an issue where ids of reordered content are converted to unicode on python 2. We now convert input ids when necessary and make sure the ordering stored on a container contains only bytestring object ids.

When changing object ordering via PATCH request we used to incorrectly store ids of reordered resources as unicode instead of a bytestring (on python2). This lead to mixed types being stored in the ordering annotations and subsequently mixed types being returned when calling `objectIds` of a container.

The problem only exists with python2 as python3 uses unicode internally.

Things we should probably discuss in this PR:
- Is the suggested upgrade-path feasible?
- The current suggestion pushes responsibility to upgrade ordering to the client, see suggestion in upgrade-guide.

Fixes #827.

### DIFF
--- a/docs/source/upgrade-guide.rst
+++ b/docs/source/upgrade-guide.rst
@@ -3,6 +3,33 @@ Upgrade Guide
 
 This upgrade guide lists all breaking changes in plone.restapi and explains the necessary steps that are needed to upgrade to the lastest version.
 
+
+Upgrading to plone.restapi X.y
+------------------------------
+
+All versions before plone.restapi x.y.z are potentially affected by an issue
+that converts ids of reordered content to unicode from a bytestring when
+running on python2.
+
+You may be affected by this issue and should run the fix if:
+
+- You used the PATCH "ordering" functionality of plone.restapi
+- Were using Python 2 at that point
+- Are seeing issues with objectIds() returning mixed string types
+
+If you need to fix object ids you can do one of the following:
+
+- Use the browser-view ``@@plone-restapi-upgrade-fix-ordering`` as a "Manager"
+  to fix all folderish content types in your Plone site.
+- Run the helper function
+  ``ensure_child_ordering_object_ids_are_native_strings``
+  from ``plone.restapi.upgrades.ordering`` for all affected objects. You could
+  do this in a custom upgrade-step implemented in your policy.
+
+We expect that most content won't actually be affected. See
+https://github.com/plone/plone.restapi/issues/827 for more details.
+
+
 Upgrading to plone.restapi 5.x
 ------------------------------
 
--- a/src/plone/restapi/tests/mixin_ordering.py
+++ b/src/plone/restapi/tests/mixin_ordering.py
@@ -195,6 +195,38 @@ class OrderingMixin:
             self.folder.contentIds(),
         )
 
+    def test_ordering_preserves_native_string_obj_id(self):
+        # sanity check, initial situation
+        for id_ in self.folder.objectIds():
+            self.assertIsInstance(id_, str)
+
+        # reorder
+        data = {"ordering": {"delta": "top", "obj_id": "doc9"}}
+        self.deserialize(body=json.dumps(data), context=self.folder)
+
+        # reordering should preserve bytestring ids in PY2 and unicode ids in PY3
+        for id_ in self.folder.objectIds():
+            self.assertIsInstance(id_, str)
+
+    def test_ordering_preserves_native_string_subset_ids(self):
+        # sanity check, initial situation
+        for id_ in self.folder.objectIds():
+            self.assertIsInstance(id_, str)
+
+        # reorder and also provide subset_ids
+        data = {
+            "ordering": {
+                "delta": "bottom",
+                "obj_id": "doc1",
+                "subset_ids": ["doc1", "doc2", "doc3"],
+            }
+        }  # noqa
+        self.deserialize(body=json.dumps(data), context=self.folder)
+
+        # reordering should preserve bytestring ids in PY2 and unicode ids in PY3
+        for id_ in self.folder.objectIds():
+            self.assertIsInstance(id_, str)
+
     def test_reorder_subsetids(self):
         # sanity check, initial situation
         self.assertEqual(
--- a/src/plone/restapi/tests/test_upgrade_ordering.py
+++ b/src/plone/restapi/tests/test_upgrade_ordering.py
@@ -1,0 +1,53 @@
+# -*- coding: utf-8 -*-
+from plone.restapi.testing import PLONE_RESTAPI_DX_INTEGRATION_TESTING
+from plone.restapi.upgrades.ordering import ensure_child_ordering_object_ids_are_native_strings
+
+import unittest
+import six
+
+
+class TestUpgradeOrdering(unittest.TestCase):
+
+    layer = PLONE_RESTAPI_DX_INTEGRATION_TESTING
+
+    def setUp(self):
+        self.portal = self.layer["portal"]
+        self.request = self.layer["request"]
+
+        self.folder = self.portal[
+            self.portal.invokeFactory("Folder", id="folder1", title="Folder")
+        ]
+        for x in range(1, 4):
+            self.folder.invokeFactory(
+                "Document", id="doc" + str(x), title="Test doc " + str(x)
+            )
+
+    def test_upgrade_ensure_child_ordering_object_ids_are_native_strings(self):
+        ordering = self.folder.getOrdering()
+
+        # use incorrect type for ordering, results in mixed type ordering ids
+        # on folder
+        ordering.moveObjectsToBottom([six.text_type('doc1')])
+
+        ensure_child_ordering_object_ids_are_native_strings(self.folder)
+
+        self.assertEqual(
+            [
+                "doc2",
+                "doc3",
+                "doc1",
+            ],  # noqa
+            self.folder.objectIds(),
+        )
+
+        # upgrade helper should ensure bytestring ids in python2 and do nothing
+        # on python3
+        for id_ in self.folder.objectIds():
+            self.assertIsInstance(id_, str)
+
+    def test_upgrade_can_be_called_with_nonetype(self):
+        ensure_child_ordering_object_ids_are_native_strings(None)
+
+    def test_upgrade_can_be_called_with_not_annotatable(self):
+        ensure_child_ordering_object_ids_are_native_strings(object())
+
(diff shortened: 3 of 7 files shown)


deiferni:

src/plone/restapi/tests/mixin_ordering.py
```diff
@@ -195,6 +195,38 @@ def test_reorder(self):
             self.folder.contentIds(),
         )
 
+    def test_ordering_preserves_bytestring_obj_id(self):
```
lukasgraf: `bytestring` -> `native_string`
(See https://portingguide.readthedocs.io/en/latest/strings.html#the-native-string)

src/plone/restapi/tests/mixin_ordering.py
```diff
@@ -195,6 +195,38 @@ def test_reorder(self):
...
+
+        # reorder
+        data = {"ordering": {"delta": "top", "obj_id": "doc9"}}
+        self.deserialize(body=json.dumps(data), context=self.folder)
+
+        # reordering should preserve bytestring ids in PY2 and unicode ids in PY3
+        for id_ in self.folder.objectIds():
+            self.assertIsInstance(id_, str)
+
+    def test_ordering_preserves_bytestring_subset_ids(self):
```
lukasgraf: bytestring -> native_string

src/plone/restapi/tests/test_upgrade_ordering.py
```diff
@@ -0,0 +1,46 @@
...
+
+        self.folder = self.portal[
+            self.portal.invokeFactory("Folder", id="folder1", title="Folder")
+        ]
+        for x in range(1, 4):
+            self.folder.invokeFactory(
+                "Document", id="doc" + str(x), title="Test doc " + str(x)
+            )
+
+    def test_upgrade_ensure_child_ordering_object_ids_are_bytestrings(self):
```
lukasgraf: bytestrings -> native_strings

src/plone/restapi/upgrades/ordering.py
```diff
@@ -0,0 +1,40 @@
+# -*- coding: utf-8 -*-
+from persistent.list import PersistentList
+from zope.annotation.interfaces import IAnnotations
+from plone.folder.default import DefaultOrdering
+
+import six
+
+
+def ensure_child_ordering_object_ids_are_bytestrings(parent):
```
lukasgraf: `bytestrings` -> `native_strings`
I would also prefer `container` instead of `parent` (parent of what?)

docs/source/upgrade-guide.rst
```diff
@@ -3,6 +3,24 @@ Upgrade Guide
...
+All versions before plone.restapi x.y.z are affected by an issue that
+converts ids of reordered content to unicode from a bytestring when running on
+python2.
+
+- Only content that was reordered is affected.
+- Only ordering metadata stored in object annotations is affected by default.
+- See https://github.com/plone/plone.restapi/issues/827 for more details.
+
+We expect that most content won't actually be affected. Nevertheless in
+https://github.com/plone/plone.restapi/commit/6a3258d771bdb8dc91f693c181f6f09d14d680d6
```
lukasgraf: -> `plone.restapi.upgrades.ordering`

deiferni:

tisto:
@deiferni what would be the possible downside of running the upgrade step by default?

deiferni:
@tisto it would be a potentially long-running upgrade for all content. We expect most content to be unaffected by the issue. Also we expect low impact for affected content. IMHO running the upgrade for all content by default poses a larger risk.

So as far as i can tell if we would want to be sure to fix all potentially affected content we would have to run the upgrade for every resource where theoretically content could have been reordered. That would most likely mean loading most/all content and perform the same duck-typing checks as in:https://github.com/plone/plone.restapi/blob/d00494f31a90a5d5070bea5e6f96254413148f54/src/plone/restapi/deserializer/mixins.py#L45-L52
Or maybe take a shortcut and load all `IFolderish` or something like that. But still for larger sites that would mean loading potentially thousands of objects to check if the ordering is broken (which it most likely is not as we expect most content has not been reordered via API).

Also as far as we can tell now the mixed types in the ordering annotations do not cause a default plone site to malfunction. IIRC we already discussed it a bit at plone conf and came to the conclusion that it is most likely not worth the risk to write and run an upgrade for all content. But if other users run into issues we can provide an upgrade helper which can be called with potentially affected content objects. And leave it to the clients to narrow down or define content which is potentially affected for them.

FYI we broke it in client code by relying on output of `objectIds` for traversal.

deiferni:

deiferni:
am a bit confused with the test errors. i think it is something unrelated to my changes, at least according to the failed tests.

deiferni:

docs/source/upgrade-guide.rst
```diff
@@ -3,6 +3,23 @@ Upgrade Guide
 
 This upgrade guide lists all breaking changes in plone.restapi and explains the necessary steps that are needed to upgrade to the lastest version.
 
+
```
tisto: @deiferni we should add instructions here how to actually run the migration step (via ZMI). The alternative would be to provide a browser view that runs the upgrade step.

docs/source/upgrade-guide.rst
```diff
@@ -3,6 +3,23 @@ Upgrade Guide
...
+Upgrading to plone.restapi X.y
+------------------------------
+
+All versions before plone.restapi x.y.z are affected by an issue that
+converts ids of reordered content to unicode from a bytestring when running on
+python2.
+
+- Only content that was reordered is affected.
```
lukasgraf: Thinking of it, this could also do with some clarification. We should make it easy and unambigous for people to determine whether or not they need to run the fix or not. Maybe something like this:

```
You may be affected by this issue and should run the fix if:

- You used the PATCH "ordering" functionality of plone.restapi
- Were using Python 2 at that point
- Are seeing issues with objectIds() returning mixed string types
```

docs/source/upgrade-guide.rst
deiferni: @tisto ok, will amend the following:
- implement an upgrade that checks everything that `is_folderish` according to the catalog (currently i only have a function that fixes one single object, the intention was that clients could implement their own upgrade. i see that we should provide a bit more convenience here) 
- browser view to manually run the upgrade

deiferni:
For the record, i have verified what happens if we do a `zodbupdate` from python2 to python3 with content with mixed string types in the ordering annotations. The `zodbupdate` script seems to be able to get rid of the mixed typing and correctly converts all the ids to native strings (i.e. `str` in python3), so upgrading from python2 to python3 will also get rid of potential issues. I think that reinforces the point not to run an upgrade by default.

deiferni:

deiferni:
@tisto i have updated this PR with the following changes:
- added a browser-view in case someone needs the fix and prefers to run it TTW
- improved the description in the upgrade-guide, clarified when people are affected as suggested by @lukasgraf and also mentioned the two possible paths to run the fix

What is your stance on the upgrade strategy? We don't object rewriting this as a "proper" upgrade-step if someone does see a good reason to do so. Currently we think that not running the upgrade by default is the better strategy for the following reasons:
- we expect most content is not affected by the issue
- it only surfaces on python2
- even if content is affected it does not AFAICT have any serious consequences (unless you do very specific things with `contentIds` in client code)
- as soon as you migrated to python3 the issue will disappear, the current version of `zodbupdate` seems to update the object ids to native strings (i.e. unicode strings) correctly
- running the upgrade for all content by default may pose a certain risk if many objects need to be migrated, especially due to the potentially long runtime and loading of many objects into memory.

So from our side this would be good to 🚢. If anyone has objections please let me know so we can address/discuss the issue.

tisto:
@deiferni thank you! I will try to find time in the next days to review this and I will also discuss the upgrade issue in the FWT and with @esteele. My personal opinion is that the browser view is sufficient.

tisto:
@deiferni @lukasgraf sorry for not handling this in an appropriate timeframe. I promise to do a release early next week.

tisto:
@deiferni @lukasgraf plone.restapi 6.0.0 released with the fix. Sorry it took so long.

lukasgraf:
Thanks @tisto! :)

PR #844, opened 2019-12-11
Add SearchableText indexing for text in blocks

luca-bellenghi:
Used in volto for documents and objects implementing IBlock interface and using blocks

### DIFF
--- a/src/plone/restapi/indexers.py
+++ b/src/plone/restapi/indexers.py
@@ -2,11 +2,7 @@
 from Products.CMFPlone.utils import safe_nativestring
 from plone.restapi.behaviors import IBlocks
 from plone.indexer.decorator import indexer
-try:
-    from plone.app.contenttypes.indexers import SearchableText
-    HAS_PAC = True
-except ImportError:
-    HAS_PAC = False
+from plone.app.contenttypes.indexers import SearchableText
 
 
 def _extract_text(block):
@@ -17,8 +13,6 @@ def _extract_text(block):
 
 @indexer(IBlocks)
 def SearchableText_blocks(obj):
-    if not HAS_PAC:
-        return ""
     std_text = SearchableText(obj)
     blocks = obj.blocks
     blocks_text = [
--- a/src/plone/restapi/indexers.py
+++ b/src/plone/restapi/indexers.py
@@ -1,0 +1,33 @@
+# -*- coding: utf-8 -*-
+from plone.restapi.behaviors import IBlocks
+from plone.indexer.decorator import indexer
+from plone.app.contenttypes.indexers import SearchableText
+import six
+
+
+def _extract_text(block):
+    result = ""
+    for paragraph in block["text"]["blocks"]:
+        text = paragraph["text"]
+        if six.PY2:
+            if isinstance(text, six.text_type):
+                text = text.encode("utf-8", "replace")
+            if text:
+                result = " ".join((result, text))
+        else:
+            result = " ".join((result, text))
+    return result
+
+
+@indexer(IBlocks)
+def SearchableText_blocks(obj):
+    std_text = SearchableText(obj)
+    blocks = obj.blocks
+    blocks_text = [
+        _extract_text(blocks[block_uid])
+        for block_uid in obj.blocks
+        if blocks[block_uid].get("@type", "") == "text"
+    ]
+    blocks_text.append(std_text)
+    text = " ".join(blocks_text)
+    return text
--- a/src/plone/restapi/indexers.py
+++ b/src/plone/restapi/indexers.py
@@ -2,7 +2,11 @@
 from Products.CMFPlone.utils import safe_nativestring
 from plone.restapi.behaviors import IBlocks
 from plone.indexer.decorator import indexer
-from plone.app.contenttypes.indexers import SearchableText
+try:
+    from plone.app.contenttypes.indexers import SearchableText
+    HAS_PAC = True
+except ImportError:
+    HAS_PAC = False
 
 
 def _extract_text(block):
@@ -13,6 +17,8 @@ def _extract_text(block):
 
 @indexer(IBlocks)
 def SearchableText_blocks(obj):
+    if not HAS_PAC:
+        return ""
     std_text = SearchableText(obj)
     blocks = obj.blocks
     blocks_text = [
--- a/src/plone/restapi/indexers.py
+++ b/src/plone/restapi/indexers.py
@@ -1,22 +1,14 @@
 # -*- coding: utf-8 -*-
+from Products.CMFPlone.utils import safe_nativestring
 from plone.restapi.behaviors import IBlocks
 from plone.indexer.decorator import indexer
 from plone.app.contenttypes.indexers import SearchableText
-import six
 
 
 def _extract_text(block):
-    result = ""
-    for paragraph in block.get("text").get("blocks"):
-        text = paragraph["text"]
-        if six.PY2:
-            if isinstance(text, six.text_type):
-                text = text.encode("utf-8", "replace")
-            if text:
-                result = " ".join((result, text))
-        else:
-            result = " ".join((result, text))
-    return result
+    return " ".join(
+        safe_nativestring(paragraph["text"]) for paragraph in block.get("text").get("blocks")
+    )
 
 
 @indexer(IBlocks)
--- a/src/plone/restapi/indexers.py
+++ b/src/plone/restapi/indexers.py
@@ -7,7 +7,7 @@ import six
 
 def _extract_text(block):
     result = ""
-    for paragraph in block["text"]["blocks"]:
+    for paragraph in block.get("text").get("blocks"):
         text = paragraph["text"]
         if six.PY2:
             if isinstance(text, six.text_type):
--- a/src/plone/restapi/indexers.py
+++ b/src/plone/restapi/indexers.py
@@ -1,14 +1,22 @@
 # -*- coding: utf-8 -*-
-from Products.CMFPlone.utils import safe_nativestring
 from plone.restapi.behaviors import IBlocks
 from plone.indexer.decorator import indexer
 from plone.app.contenttypes.indexers import SearchableText
+import six
 
 
 def _extract_text(block):
-    return " ".join(
-        safe_nativestring(paragraph["text"]) for paragraph in block.get("text").get("blocks")
-    )
+    result = ""
+    for paragraph in block.get("text").get("blocks"):
+        text = paragraph["text"]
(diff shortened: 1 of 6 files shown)


luca-bellenghi:

luca-bellenghi:
@tisto can't understand why first jenkins test fails.
It's here: https://jenkins.plone.org/job/pull-request-5.2-3.6/730/
Seems to fail due to problems with GenericSetup

tisto:
@luca-bellenghi trying to re-run the job: https://jenkins.plone.org/job/pull-request-5.2-3.6/731/

jensens:
The same is valid for the Blocks behavior. It breaks separation of concerns and does not belong into the Restapi. Why not put it into `plone.app.dexterity`  or `plone.app.contenttypes` where we have all the other base behaviors? This cries for refactoring.

tisto:
@jensens you are correct and I agree with you that keeping the block behavior and the indexer in plone.restapi are indeed code smells. 

This decision was made after a longer architectural discussion where we just couldn't find a perfect solution tbh. This happened when we implemented the blocks and the indexers are just a follow up of this, if you will. 

The reasoning behind our decision to put this code here is that plone.restapi acts as the connection point between classic Plone and Volto. plone.restapi will continue to move at a faster pace than Plone for the foreseeable future. 

We added, removed and refactored Volto-related code in plone.restapi in the past (e.g. the blocks behavior and endpoint) and we expect this will continue to happen in the future. Keeping things together in plone.restapi allowed us to do changes quickly and even remove functionality with almost no cost, because Volto is the only consumer of this code right now. We usually mark those code parts as experimental, to make it clear that this is subject to change in the future. This will be true for the indexers as well.

If we spread functionality across packages, removing or refactoring will become a lot harder for us and we might end up in a situation where we have to create multiple branches of multiple packages and maintain that over a longer period of time. We saw this happening in multiple Plone core packages and this comes with high costs for the maintainers and leads to lots of confusion for developers and clients.

We do not claim that this is the final destination for that code. Refactoring this in the future is definitely a valid option. As soon as things settled a bit in the Plone 6 branch. Does that sound acceptable to you?

Apart from that, I am looking forward to discussing the finer details of the Plone core package architecture (aka spreading functionality randomly across packages) with you over a beverage next time we meet. :)

tisto:

src/plone/restapi/indexers.py
```diff
@@ -0,0 +1,33 @@
...
+    result = ""
+    for paragraph in block.get("text").get("blocks"):
+        text = paragraph["text"]
+        if six.PY2:
+            if isinstance(text, six.text_type):
+                text = text.encode("utf-8", "replace")
+            if text:
+                result = " ".join((result, text))
+        else:
+            result = " ".join((result, text))
```
ale-rt:
```
from Products.CMPPlone.utils import safe_nativestring
" ".join(result, safe_nativestring(text))
```
does the same with lot less lines.
Making a single `" ".join` operation at the end would also be more efficient.
Something like:
```
return " ".join(safe_nativestring(paragraph["text"]) for paragraph in block.get("text").get("blocks"))
```

src/plone/restapi/indexers.py
```diff
@@ -0,0 +1,33 @@
...
+    return result
+
+
+@indexer(IBlocks)
+def SearchableText_blocks(obj):
+    std_text = SearchableText(obj)
```
ale-rt: I am not sure that calling `SearchableText` is good here.
I think this is going to clash with the `collective.dexteritytextindexer` behavior.

src/plone/restapi/indexers.zcml
```diff
@@ -0,0 +1,11 @@
+<configure
+    xmlns="http://namespaces.zope.org/zope"
+    xmlns:five="http://namespaces.zope.org/five"
+    xmlns:i18n="http://namespaces.zope.org/i18n"
+    xmlns:genericsetup="http://namespaces.zope.org/genericsetup"
```
ale-rt: Those are not needed

src/plone/restapi/tests/test_blocks_searchable_text.py
```diff
@@ -0,0 +1,98 @@
...
+
+        fti = queryUtility(IDexterityFTI, name="Document")
+        behavior_list = [a for a in fti.behaviors]
+        behavior_list.append("volto.blocks")
+        fti.behaviors = tuple(behavior_list)
+
+        self.doc = createContentInContainer(
+            self.portal, u"Document", id=u"doc", title=u"A document"
+        )
+        transaction.commit()
```
ale-rt: I think you are calling commit to have the catalog queue processed.
It is better then to call the `processQueue` function.

src/plone/restapi/tests/test_blocks_searchable_text.py
```diff
@@ -0,0 +1,98 @@
...
+from plone.restapi.testing import RelativeSession
+from zope.component import queryUtility
+
+import transaction
+import unittest
+
+
+class TestSearchTextInBlocks(unittest.TestCase):
+
+    layer = PLONE_RESTAPI_DX_FUNCTIONAL_TESTING
```
ale-rt: Isn't it better to use an integration test here? Or is a functional test mandatory for the restapi? I am not really into it, just asking

tisto:
@ale-rt thank you for your in-depth review. You rock! :)

ale-rt:
The code looked good anyway, most are just improvements

cekk:
just my 2 cents: an option could be creating a "Volto" package (plone.volto?) with all the code needed: behaviors (remove text, add blocks fields), indexers, restapi endpoints, etc? And when Volto will be officially in the core, we can move pieces into the core parts. With that, we get rid of kitconcept's Volto packages and we don't put dirty things into the core or restapi.

jensens:
@tisto I know the pain. So, putting code in a place where it does not belong to is in the long term another pain. 

I watched what is happening with the block behavior in here for a while. Since it didn't hurt until now and was kind of experimental it was fine. But know we're going to make this Plone 6 and Volto a very central element.

I agree that our code base is split, and I agree it is often split up in a strange way. Dealing with branches is not that difficult (thanks to our great Coredev Buildout and Jenkins setup, really!). Anyway its more work. 

I think having our indexers in CMFPlone and the type-code is in `p.a.[dexterity|contenttypes]` was some decision far in the past, as we supported both, Archetypes and dexterity. So, it always has a reason. But blocks are primary Volto/Content-type specific and RESTAPI is the wrong place for them. Full stop. I am sure you agree here.

Regarding the behavior, this should really go into one of `p.a.[dexterity|contenttypes]` (I would go for contenttypes, but I am ok with dexterity too).

Regarding the indexer: Here we come to one long wanted feature to go into core:  `collective.dexteritytextindexer` [Plip 2780](https://github.com/plone/Products.CMFPlone/issues/2780)
Then the implementation for Blocks could nicely plug in there.

Since we are talking about Plone core code I prefer a clean solution instead of putting code into wrong places. I can not approve this PR, because we are at the point where refactoring is needed (with or w/o before moving on. For Plone 6 this "chaos" is not acceptable.

jensens:
> option could be creating a "Volto" package (plone.volto?) with all the code needed: behaviors (remove text, add blocks fields),

Please please please not another package. Plone already has packageritis.

tisto:
@jensens this is not going to become Plone 6. We have to ship and stabilize Volto 4 first. Stabilizing means that we will continue to move fast and that we maybe have to refactor code further. The Plone 6 branch is empty for a good reason right now.

When we are done with this experimental phase, we will move the gs profile code from kitconcept.voltodemo into CMFPlone and the block behavior and the indexers to their correct places. In the meantime, we mark this part as experimental so that only people who know what they are doing rely on this.

We want to be sure that we did things right before we start Plone 6. If we start too early with this we risk putting lots of work into this and gain nothing.

The alternative to this PR is that we put all that stuff (gs profile, blocks, indexers) into kitconcept.voltodemo and continue to work there...Though, kitconcept.voltodemo was always meant as an example and temporary add-on, not a vital part of the Volto roadmap.

jensens:
@tisto, great to get a picture of the roadmap. It makes sense, but wasn't communicated well.

Please mark all that code "experimental", maybe with a line or two as in your comment above. Actually I can not find any comment or docstring in the code now, i.e. here: https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/behaviors.py

At the current state I would not move code around. just make sure code and docs a communicating well why the code is in here and that it is experimental.

tisto:
@jensens yeah. You are right (again). Sorry for not putting a sufficient amount of effort into the communication part of this. We go step by step right now and our main focus is still finishing Volto 4. Next step then is to publish the concrete roadmap and ask the Plone community and the FWT for feedback on our ideas. Maybe a good task for the Alpine City sprint. :)

tisto:
@luca-bellenghi did you see @ale-rt's comments? Do you plan to continue to work on this?

tisto:
@luca-bellenghi @pnicolli ping :)

src/plone/restapi/indexers.zcml
tisto: Fixed.

src/plone/restapi/tests/test_blocks_searchable_text.py
tisto: In plone.restapi tests we usually have to rely on functional tests. You usually do a low-level test setup and then do an API call to check if the expected response is there. This is kind of a functional browser test. We could use integration tests but that would be more work and my personal opinion is that this would lead us to testing on the wrong level. Of course this is debatable for every single use case. Though, we usually stick to functional tests.

src/plone/restapi/tests/test_blocks_searchable_text.py
tisto: When we use a functional test, does that make much of a difference?

src/plone/restapi/indexers.py
tisto: @ale-rt would you mind elaborating how the indexer could clash with c.dexteritytextindexer?

src/plone/restapi/indexers.py
tisto: @ale-rt fixed.

tisto: @ale-rt I had to revert my fix. plone.restapi supports Plone 4.3 which does not have safe_nativestring. :(

tisto:

tisto:

src/plone/restapi/indexers.py
wesleybl: @ale-rt @tisto this really caused problems. See: #1764

PR #858, opened 2020-01-22
Fix Plone 4.3.x build

lukasgraf:
This adds version constraints to `plone-4.3.x.cfg` needed for Python 2.7 compatibility.

Fixes build failures like these:
https://travis-ci.org/plone/plone.restapi/jobs/640324488


lukasgraf:

lukasgraf:
@ale-rt I deliberately left out the changelog since it doesn't contain any functional changes to the code. Quite the opposite, it would produce an entry that would just be distracting for someone combing through the CL to look for API changes. That's why we usually haven't added CL entries for these kinds of changes.

Issue #856, opened 2020-01-20
Error thrown when a content object has an unknown vocabulary item on Edit route


sneridagh:
The API answers with an uncaught 400 error. We might want to gracefully catch this error and allow the request go through.

PR #860, opened 2020-01-24
Degrade gracefully when a term set in a content field does not exists…

sneridagh:
… in the assigned vocabulary


sneridagh:
@tisto @buchi I've logged the error for this one, although we are not doing it for the choices field.
Should we log both or none?
https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/serializer/dxfields.py#L49

sneridagh:

sneridagh:
Fixes: #856

PR #865, opened 2020-02-12
Resolveuid in blocks functional test

tisto:


src/plone/restapi/tests/test_resolveuid_functional.py
```diff
@@ -0,0 +1,114 @@
...
+                        "ca5908a5-3f58-4cd5-beb7-9bd1539d6744",
+                        "791bf004-7c88-4278-8490-13b85c3fa4b4",
+                    ]
+                },
+            },
+        )
+        transaction.commit()
+
+        self.assertEqual(
+            "http://localhost:55001/Plone/target",
```
tisto: @sneridagh @buchi I'd expect this to be a resolveUID link or am I missing something obvious here? The serializer should convert the path to a resolveUID that is stored internally, correct? It is super odd that the serializer/deserializer tests work but nothing is actually stored in the database.  cc @wkbkhard

cekk: @tisto FOUND IT!
Deserializer stores the value correctly, but searializer is actual overriding that value when it convert the uid into a path.

The reason could be that the we are working with a shallow copy of the field's value, and every change into it, will be reflected into the original one.
With a deep copy, everything works fine.

tisto: @cekk w00t!!!

tisto: @cekk would you mind pushing your solution to this branch?

cekk:
@tisto  updated this branch with the fix.

@buchi @ale-rt is it ok for you using a deepcopy here? other possible solution were:
- json encode/decode dictionary
- iterate over dictionary items and make a deepcopy on each value

tisto:
@cekk the thing that I do not understand is why is my functional test still passing? If the problem is fixed, the test should actually fail because it checks that a path without revolveUID is stored in the database.

cekk:
@tisto easy: because you set "Plone" as sitename in your functional test, but the deserializer convert the path to uid only if in the url there is the right sitename (that is "plone") ;)

If you want to keep these functional tests, i can fix them to not fail with the right sitename

tisto:
@cekk yeah. please go ahead!

tisto:
@cekk I fixed the problem. It had nothing to do with the portal URL. The Plone testing layer uses "plone" instead of the default "Plone" that you got when you create a Plone site manually. This was all correct in the acceptance test as far as I can see. Though, I was checking the REST API response first which was serialized again, so no UUID there. I just removed that line and checked the object on database level and it all works fine. I also checked the functionality and it works like a charm. Thank you for the fix!!!

I merged this PR into the main one. Now I will write a few more Cypress tests and then I guess we are good to go.

PR #870, opened 2020-02-20
Resolveuid in block more generic

cekk:
Refactored methods to be more generic and work also with standard blocks that are not text ones.

### DIFF
--- a/src/plone/restapi/deserializer/blocks.py
+++ b/src/plone/restapi/deserializer/blocks.py
@@ -12,20 +12,34 @@ from zope.interface import implementer
 from zope.publisher.interfaces.browser import IBrowserRequest
 
 
-def path2uid(context, path):
+def path2uid(context, portal, href):
     # unrestrictedTraverse requires a string on py3. see:
     # https://github.com/zopefoundation/Zope/issues/674
-    if not isinstance(path, str):
-        path = path.decode("utf-8")
-    obj = context.unrestrictedTraverse(path, None)
+    if not href:
+        return ''
+    portal_url = portal.absolute_url()
+    portal_path = '/'.join(portal.getPhysicalPath())
+    path = href
+    context_url = context.absolute_url()
+    relative_up = len(context_url.split("/")) - len(portal_url.split("/"))
+    if path.startswith(portal_url):
+        path = path[len(portal_url) + 1 :]
+    if not path.startswith(portal_path):
+        path = '{portal_path}/{path}'.format(
+            portal_path=portal_path, path=path.lstrip("/")
+        )
+    obj = portal.unrestrictedTraverse(path, None)
     if obj is None:
-        return None, None
+        return href
     segments = path.split("/")
     suffix = ""
     while not IUUIDAware.providedBy(obj):
         obj = aq_parent(obj)
         suffix += "/" + segments.pop()
-    return IUUID(obj), suffix
+    href = relative_up * "../" + "resolveuid/" + IUUID(obj)
+    if suffix:
+        href += suffix
+    return href
 
 
 @implementer(IFieldDeserializer)
@@ -38,9 +52,6 @@ class BlocksJSONFieldDeserializer(DefaultFieldDeserializer):
         portal = getMultiAdapter(
             (self.context, self.request), name="plone_portal_state"
         ).portal()
-        portal_url = portal.absolute_url()
-        context_url = self.context.absolute_url()
-        relative_up = len(context_url.split("/")) - len(portal_url.split("/"))
         if self.field.getName() == "blocks":
             for block in value.values():
                 if block.get("@type") == "text":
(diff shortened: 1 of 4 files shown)


src/plone/restapi/deserializer/blocks.py
```diff
@@ -12,20 +12,34 @@
 from zope.publisher.interfaces.browser import IBrowserRequest
 
 
-def path2uid(context, path):
+def path2uid(context, portal, href):
     # unrestrictedTraverse requires a string on py3. see:
     # https://github.com/zopefoundation/Zope/issues/674
-    if not isinstance(path, str):
```
sneridagh: @cekk are you observing this "unrestrictedTraverse requires a string on py3. see:" comment later in the new code? I cannot see the check.

src/plone/restapi/deserializer/blocks.py
```diff
@@ -12,20 +12,34 @@
...
-        path = path.decode("utf-8")
-    obj = context.unrestrictedTraverse(path, None)
+    if not href:
+        return ''
+    portal_url = portal.absolute_url()
+    portal_path = '/'.join(portal.getPhysicalPath())
+    path = href
```
sneridagh: ?

src/plone/restapi/deserializer/blocks.py
```diff
@@ -12,20 +12,34 @@
 from zope.publisher.interfaces.browser import IBrowserRequest
 
 
-def path2uid(context, path):
+def path2uid(context, portal, href):
```
tisto: @cekk do we pass a full URL "href" to path2uid? If we name the function "path2uid" one of the input params should be "path" in my opinion.

src/plone/restapi/deserializer/blocks.py
tisto: I'd like to understand as well why we switched from unrestrictedTraverse to getPhysicalPath

src/plone/restapi/deserializer/blocks.py
```diff
@@ -12,20 +12,34 @@
...
+    obj = portal.unrestrictedTraverse(path, None)
     if obj is None:
-        return None, None
+        return href
     segments = path.split("/")
     suffix = ""
     while not IUUIDAware.providedBy(obj):
         obj = aq_parent(obj)
         suffix += "/" + segments.pop()
-    return IUUID(obj), suffix
```
tisto: @cekk can you elaborate why we changed the path2uid return from a tuple to returning a full path/href?

tisto: ok, got it now. you moved logic into the path2uid function...

src/plone/restapi/deserializer/blocks.py
```diff
@@ -38,25 +52,28 @@ def __call__(self, value):
...
-                                    entity["data"]["href"] = href
-                                    entity["data"]["url"] = href
-                                print("DESERIALIZE " + before + " -> " + href)  # noqa
+                            deserialized_href = path2uid(
+                                context=self.context, portal=portal, href=href
+                            )
+                            entity["data"]["href"] = deserialized_href
+                            entity["data"]["url"] = deserialized_href
+                else:
+                    # standard blocks can have an "url" or "href" field
```
tisto: @cekk @sneridagh is it fair to assume all blocks have a url/href field? I am just wondering here if we have a unit test that tests what happens if a block does not have those fields. From what I can see in the code blocks this should work...

src/plone/restapi/tests/test_resolveuid.py
```diff
@@ -23,7 +23,15 @@ def setUp(self):
             )
         ]
         self.doc2 = self.portal[
-            self.portal.invokeFactory("Document", id="doc2", title="Target Document")
+            self.portal.invokeFactory(
```
tisto: @cekk just wondering why this happens here. Should black prevent changes like this in PRs? I mean if black auto-formats things this isn't something that should change in any way...

tisto:
@cekk could you please rebase with master? The Travis config in master stopped working when you branched this PR. Therefore we need to rebase for Travis to run again... @sneridagh and I are more than happy to help with getting this merged into master. We plan to do a plone.restapi 7.0.0alpha1 release with this soon.

src/plone/restapi/tests/test_resolveuid.py
cekk: i think that this is my problem. I set black to auto-format on save in my editor, but i forgot to disable it when i was working on this pr.
Should i revert these changes?

src/plone/restapi/deserializer/blocks.py
cekk: yep ;)

src/plone/restapi/deserializer/blocks.py
cekk: if they don't have these fields, they don't look for them, so it shouldn't be a problem.
I left this part of code untouched (setting both href and url), but probably we could do a better job setting only the needed one and save some bytes

tisto: I think we are fine.

src/plone/restapi/tests/test_resolveuid.py
tisto: auto-formatting in the ide is fine and recommended. do you have flake8 auto-formatting enabled with a line-length limit? we set line-length to 200 in flake8 in the volto config because black takes care of that as well. if your ide black/flake8 picks up the custom config of the volto project things should be fine. maybe we should have a pre-commit hook that runs black to prevent something like this to happen on the CI...

src/plone/restapi/deserializer/blocks.py
cekk: Just checked: it seems that volto returns a full url. In that method i try to manage both cases (full url or path).
Let me know what naming do you prefer and i'll fix it ;)

src/plone/restapi/deserializer/blocks.py
cekk: @sneridagh  it was an old comment made by who started this feature.
@tisto first of all (rows 20-30) i'm trying to get the right object path because i'm handling full urls (that could be without the site name) and paths. I replace the portal_url from the given string and append the portal_path to. have the right string to pass to unrestrictedTraverse to check if the object exists.
More clear now?

src/plone/restapi/deserializer/blocks.py
cekk: See comment below ;)

src/plone/restapi/tests/test_resolveuid.py
cekk: i'm not working on volto project for this pr, so probably that's the main problem.
And yes, i have a limit set to 80 because of old habits.
If you want i can re-format this file

cekk:
@tisto this is a branch from "resolveuid-in-blocks" branch. Should i have to rebase it with master anyway?

tisto:
@cekk yes, please. Sorry for the late response.

tiberiuichim:
There is a conflict of approaches between this branch and the block transformer branch here: https://github.com/plone/plone.restapi/pull/915

My proposal is that the generic approach is not implemented and instead specific block converters are implemented for the blocks that need them. The code here can be structured as a generic function that can be called from those block converters.

tisto:
@cekk @tiberiuichim since https://github.com/plone/plone.restapi/pull/915 is merged now I guess we have to refactor this PR. The PR itself is fine, it just needs rebase/refactoring.

cekk:
this has been replaced by #952

PR #872, opened 2020-02-21
Allow objects paths and UIDs to link translations.

### LINKED ISSUE #645: POST to @translations requires full URL for translated content (which is nonsense)
Setting a translation via the `@translations` endpoint requires the full URL of the translation content:

https://plonerestapi.readthedocs.io/en/latest/translations.html

```
POST /plone/en/test-document/@translations HTTP/1.1
Accept: application/json
Authorization: Basic YWRtaW46c2VjcmV0
Content-Type: application/json

{
    "id": "http://localhost:55001/plone/es/test-document"
}
```

This makes no sense:
- a translated resource always lives within in the same Plone site
- on the client side I usually deal with resource paths inside given Plone site

Fix:
- specify the translation objects through their relative path inside the Plone site only

erral:
Fixes #645

### DIFF
--- a/src/plone/restapi/services/multilingual/pam.py
+++ b/src/plone/restapi/services/multilingual/pam.py
@@ -7,11 +7,13 @@ from plone.restapi.services import Service
 from Products.CMFCore.utils import getToolByName
 from Products.CMFPlone.interfaces import ILanguage
 from zope.component import adapter
+from zope.component import getMultiAdapter
 from zope.interface import alsoProvides
 from zope.interface import implementer
 from zope.interface import Interface
 
 import plone.protect.interfaces
+import six
 
 
 @implementer(IExpandableElement)
@@ -55,6 +57,14 @@ class LinkTranslations(Service):
     """ Link two content objects as translations of each other
     """
 
+    def __init__(self, context, request):
+        super(LinkTranslations, self).__init__(context, request)
+        self.portal = getMultiAdapter(
+            (self.context, self.request), name="plone_portal_state"
+        ).portal()
+        self.portal_url = self.portal.absolute_url()
+        self.catalog = getToolByName(self.context, "portal_catalog")
+
     def reply(self):
         # Disable CSRF protection
         if "IDisableCSRFProtection" in dir(plone.protect.interfaces):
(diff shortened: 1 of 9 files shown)


erral:

src/plone/restapi/services/multilingual/pam.py
```diff
@@ -92,17 +102,24 @@ def reply(self):
...
-        portal_url = portal.absolute_url()
-        if url.startswith(portal_url):
-            content_path = url[len(portal_url) + 1 :]
-            content_path = content_path.split("/")
-            content_item = portal.restrictedTraverse(content_path)
-            return content_item
-
-        return None
+    def get_object(self, key):
+        if isinstance(key, six.string_types):
```
tisto: @erral may I ask why you check for a string here? What other possibilities do we have? What happens if this is not a string? Just wondering...

src/plone/restapi/tests/test_translations.py
```diff
@@ -81,7 +81,7 @@ def setUp(self):
         )
         transaction.commit()
 
-    def test_translation_linking_succeeds(self):
+    def test_translation_linking_by_url_succeeds(self):
```
tisto: @erral nitpicking: I'd remove the "_succeeds" here since it does not add any kind of information IMHO. Not your fault. You just continued to use what was there already.

src/plone/restapi/services/multilingual/pam.py
erral: I copied the code from the @copy endpoint: https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/services/copymove/copymove.py#L27

Browsing in the history of changes, on previous versions we have an option to use intids as a way to identify objects:

https://github.com/plone/plone.restapi/blame/2c8fae7e412e6606f16617ab869f6b3c0b7d2c69/src/plone/restapi/services/copymove/copymove.py#L27

We removed the use of intids but didn't remove the string check:

https://github.com/plone/plone.restapi/commit/c8740e3e04f0fc66e8ae2362912235fd3beb290d

Which later was migrated to a six.string_types check.

I can do another PR remove that string_types check both here and in @copy endpoints.

src/plone/restapi/tests/test_translations.py
erral: done :)

src/plone/restapi/services/multilingual/pam.py
tisto: @erral that would be great. We can do it after the merge...

tisto:

PR #876, opened 2020-03-02
Remove uneeded stringtype checks

### LINKED ISSUE #875: Unneeded string checks
In @translations and @copymove endpoints we keep some ```isinstance(key, six.string_types)``` checks for ages.

The reason behind it is that we used to have an option to get the object using its intid. See: https://github.com/plone/plone.restapi/blame/2c8fae7e412e6606f16617ab869f6b3c0b7d2c69/src/plone/restapi/services/copymove/copymove.py#L27

We removed that feature on https://github.com/plone/plone.restapi/commit/c8740e3e04f0fc66e8ae2362912235fd3beb290d but did not remove the string check. 

Now we do not need those checks because those endpoints can only get objects by path or UID which in any case are strings.

erral:
Fixes #875

### DIFF
--- a/news/875.feature
+++ b/news/875.feature
@@ -1,0 +1,2 @@
+Remove unneeded stringtype checks
+[erral]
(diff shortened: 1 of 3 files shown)


erral:

tisto:

ale-rt:
This is a change in behavior with respect to the previous implementation.
```
self.get_object(None)
```
will raise an exception now (before it was just returning `None`).

I have no idea if this is good or bad :)

PR #878, opened 2020-03-03
Add a catalog serializer guard when returning fullobjects in case the…

sneridagh:
… object doesn't exist anymore because for some reason it failed to uncatalog itself.


tisto:

PR #883, opened 2020-03-13
Upgrade to Plone 5.2.1

tisto:


src/plone/restapi/tests/http-examples/translated_messages_object_history.resp
```diff
@@ -22,9 +22,9 @@ Content-Type: application/json
     "action": "Crear", 
     "actor": {
       "@id": "http://localhost:55001/plone/@users/test_user_1_", 
-      "fullname": "", 
+      "fullname": "test_user_1_", 
       "id": "test_user_1_", 
-      "username": "test-user"
+      "username": null
```
tisto: @pbauer @jensens @esteele it seems plone.restapi started to return null for username when upgrading from Plone 5.2.0 to 5.2.1. I could not figure out why. Any idea or pointers where to look? This isn't a problem for plone.restapi AFAIK. Though, maybe something we might want to investigate.

PR #885, opened 2020-03-19
[WIP] dxcontent: add targetUrl for primary file fields

csenger:
https://github.com/plone/plone.restapi/issues/886

### DIFF
--- a/news/886.feature
+++ b/news/886.feature
@@ -1,0 +1,2 @@
+Add targetUrl to the dxcontent serializer for primary file fields to be able to download a file directly.
+[csenger]
--- a/src/plone/restapi/interfaces.py
+++ b/src/plone/restapi/interfaces.py
@@ -52,6 +52,21 @@ class IFieldSerializer(Interface):
         """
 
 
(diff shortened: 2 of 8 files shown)


sneridagh:
@carsten targetUrl it is then?

csenger:
> @carsten targetUrl it is then?

@sneridagh Yes, it is. 

The logic in the endpoint now mirrors the behavior implemented in plone 4. It adds targetUrl only if the user is not able to edit the content. That is the behavior I talked about with Rob and I think it is convenient for Volto (as it directly fits the use case). Though while implementing it I not sure it is the best implementation for the API. For a more generic use we'd have to add that url every time and name the primary field.

We can't just add that info to the file field object as there is a second use case for link types where the remoteUrl should be used, but this is not an object but a string

So we might want to add an extra property in the future regardless of the edit permission like 

```
primaryField: {
  name: "file",
  target: "http://...."
}
```

Please review the PR.

tisto:
@sneridagh this is what we agreed upon, right?

tisto:

PR #888, opened 2020-03-20
Link translation on content creation feature

sneridagh:


src/plone/restapi/services/content/add.py
```diff
@@ -29,6 +33,8 @@ def reply(self):
         type_ = data.get("@type", None)
         id_ = data.get("id", None)
         title = data.get("title", None)
+        translationOf = data.get("translationOf", None)
```
sneridagh: @tisto crap, just realised of the naming... Poisoned by JS :) Did we agreed on the naming of parameters on the RESTAPI?

tisto: I noticed that as well. :) We agreed not to have a naming convention and to just map over Plone var names (which are a horrible mess as we all know). Though, we should come up with a naming convention for new var names. At some point, we might even want to be able to run plone.restapi in a strict mode that uses a consistent naming. Though, that would break lots of things...

src/plone/restapi/services/content/add.py
```diff
@@ -88,3 +111,27 @@ def reply(self):
...
 
         return serialized_obj
+
+    def get_object(self, key):
+        portal = getMultiAdapter(
+            (self.context, self.request), name="plone_portal_state"
+        ).portal()
+        catalog = getToolByName(self.context, "portal_catalog")
+
+        if isinstance(key, six.string_types):
```
erral: We are trying to get rid of this `six.string_types` check in https://github.com/plone/plone.restapi/pull/876 (sorry I need to review @ale-rt's comment in there).

sneridagh:
@erral Thanks! What is the recommendation instead?

src/plone/restapi/services/content/add.py
tisto: @erral it is not possible to use the CMFPlone function that does this because plone.restapi needs to work with other Plone versions as well.

tisto: @erral sorry, got this wrong. Ignore my comment!

sneridagh:
@erral oh ok, I see it in the other PR. Maybe we could extract this one to an util helper to reuse it anywhere. @tisto what do you think?

@erral @tisto other question. I'm implementing a basic working Multilingual support for Volto :) I have one good one to debate. It's the translation locator. We need to know in Volto where the translation is going to be created. This is delegated to the ITranslationLocator adapter in PAM. We need to use the same logic to place the upcoming translation, but we need to know it upfront :) I thought into stuff this info in the expansion, however, there is another useful info, which is the already translatied content links (which we need also for the language selector).

How about to extend the expansion with this information? Then build the link according to this information?

erral:
Another option would be to delegate the translation creation to a new endpoint, which will decide, according to the configuration in Plone where to create the translation.

sneridagh:
@erral I've drafted a new endpoint `translation-location` specs. It should call the ITranslatorLocator adapter, then return the proper location. What do you think?

https://github.com/plone/plone.restapi/pull/888/commits/5fe662adc5168f6ee755d95b34a21a20cd67f358

sneridagh:
@erral @tisto ready for review!

src/plone/restapi/services/content/add.py
sneridagh: @erral mmm I forgot this, @tisto maybe, do we want to convert this get_object into a reusable utils function?

I can refactor it and include #876 updates. However, I don't get if @ale-rt 's comment is something that we want to improve.

tisto:

src/plone/restapi/tests/test_translations.py
```diff
@@ -206,3 +206,74 @@ def test_calling_with_an_untranslated_content_gives_400(self):
...
+        response = requests.get(
+            "{}/@translation-locator?targetLanguage=de".format(
+                self.es_content.absolute_url()
+            ),
+            headers={"Accept": "application/json"},
+            auth=(SITE_OWNER_NAME, SITE_OWNER_PASSWORD),
+        )
+        self.assertEqual(200, response.status_code)
+
+        self.assertEqual("http://localhost:55001/plone/de", response.json().get("@id"))
```
tisto: @sneridagh you can not assume that the testing ZSERVER_PORT is always 55001. The jenkins.plone.org testrunner randomly assigns the ZSERVER_PORT. That's why the Plone coredev Jenkins jobs fail.

tisto:

src/plone/restapi/tests/test_translations.py
tisto: Fixed.

src/plone/restapi/services/content/add.py
tisto: @sneridagh I gave this some thought. 90% of our attributes are snake_case. We have some ancient lowerCamelCase attributes from AT and Zope. Though, if we ever want to become consistent in naming we will most likely force snake_case to be consistent between backend and frontend (Python and JavaScript). Therefore I'd say we make snake_case the default. I have to admit that I have a slight personal preference towards snake_case because it is a lot more readable than camelCase. It could be that the torture with Java in my early career plays a role in that as well. :)

tisto:

src/plone/restapi/services/content/add.py
sneridagh: :) ok then! I will amend it!

sneridagh:
@tisto I think it's ready now, merged #876 changes too, no reusable utils in order to simplify.

tisto:

src/plone/restapi/services/content/add.py
pnicolli: Just dropping a quick comment here. I'm curious about why forcing the consistency among python and js, which have different preferred standards in their communities (afaik snake_case is usually preferred in python, while camelCase is for sure preferred in javascript, especially in the react world). Imho, we should let js be js and keep all the camelCasing in volto (and everything else) as it is, while working on consistency in the Plone python code of course.
To add more on that, I think that if we want to make it easier to onboard js devs in the Plone world with volto, keeping it as close to the community standards should be the goal.

tisto: @pnicolli we decided to continue to use snake_case for plone.restapi (see https://github.com/plone/plone.restapi/commit/114bb91298e5b3d06287848adb43e805dafa44bf). I guess we could convert snake_case that comes from the REST API to camelCase and be consistent in Volto. I'd love to be consistent in Plone REST API as well but that was never really an option since Plone/Zope itself is a mess and forcing a certain style in the REST API would have been even more confusing.

tisto: @pnicolli could you please create an issue for documenting and enforcing JS naming conventions in Volto?

pnicolli: Here it is, thanks to @ksuess https://github.com/plone/volto/issues/1317#issuecomment-606464898

pnicolli: btw, [graphene](https://github.com/graphql-python/graphene) converts its snake case methods to camel case graphql attributes automatically.. it's a great way to integrate seamlessly to both worlds, not sure it's worth the hassle here.

tisto: @pnicolli the problem here is that we can not expect the backend to provide snake_case and this could make the implementation brittle. In Plone we use every possible way to define vars. I am open to suggestions here or someone who wants to explore the options we have...

PR #890, opened 2020-03-24
Make discussion endpoint return content that is deserialized via port…

tisto:
…al transforms.


tisto:

PR #897, opened 2020-04-01
Fix for the use case while updating user properties in the @user endp…

sneridagh:
…oint, and the portrait is already previously set but the request includes the (previously) serialized value as a string because the user are not updating it


sneridagh:

PR #904, opened 2020-04-08
Fix not searched test

cekk:


cekk:
@rodfersou this fix your errors on tests.

The problem was that on 4.3 search settings are not on the registry but in site_properties ;)

rodfersou:
@cekk thanks

PR #912, opened 2020-04-15
Fix richtext encoding in deserializer

cekk:
This will fix the [#1250](https://github.com/plone/volto/issues/1250) in Volto.

Basically if we post/patch some html with strange ascii/entities characters (for example &#x27;), Plone breaks in portal_transforms.

This simple fix cleans the html before storing it into Plone.


cekk:
wrong branch

PR #913, opened 2020-04-15
Fix richtext encoding in deserializer

cekk:
This will fix [#1250](https://github.com/plone/volto/issues/1250) for Volto.

We could receive html with some html entities that potentially broke some Plone parts (portal_transform in this case).

With this fix, we perform a cleanup before store data to Plone


cekk:

PR #915, opened 2020-04-16
Added block transformer concept

tiberiuichim:
There's now an adapter based mechanism to transform blocks on ObjectCreated and ObjectModified events


src/plone/restapi/interfaces.py
```diff
@@ -163,3 +163,15 @@ def parse(idx_query):
         have been reconstructed to the proper data types that the adapted
         index expects.
         """
+
+
+class IBlockTransformer(Interface):
+    """ A multiadapter that can post-process (transform) the value of a block
+    """
+
+    def transform(value, event):
```
tiberiuichim, line 172: Instead of having the event passed here as an argument, would you rather have the event being provided in the adapter lookup, for a three way adapter?

```
queryMultiAdapter(
            [obj, request, event], IBlockTransformer, name=block.get('@type', ''))
```

sneridagh: Seems fine for me!

cekk:
@sneridagh @tiberiuichim that's really interesting!
This is gonna replace the the serializer/deserializer for blocks already implemented in resolveuid branches?

sneridagh:
@cekk that also crossed my mind too, since I love the adapter on specific block types, and the chainability that we get because of this. We might want to have a chat about it during this week. Are you planning to sprint during these days?

cekk:
The remote sprint? unluckily not, but will be available for a chat about this topic because i'm struggling with it in several projects ;)

sneridagh:
/cc @tisto @tiberiuichim @avoinea @buchi @lukasgraf 

Do you have time for a meeting/reflect on this one?

tisto:
@sneridagh yeah, I do. I would love to hear what @buchi and @lukasgraf think about this...

sneridagh:
@tiberiuichim I love this concept. So now the resolveuid thing is working with the transformers, right? @cekk if we merge this, you'll need to refactor your PR #870, right?

@buchi @lukasgraf It would be great if you can take a look and check on this, before merging. @tisto and me love this new approach, since it's pluggable and extensible and more ZCA-ish.

sneridagh:

tiberiuichim:
I've renamed the concept to IBlockConverter. If any of you can think of better naming, let me know and I'll adjust. I'll also write some documentation, so don't merge it yet.

tiberiuichim:

tiberiuichim:

lukasgraf:
Sorry for the late reply - looks good to me! 👍

sneridagh:
@tiberiuichim We found a slight issue in the current implementation. What if you want to have at the same time two converters over the same block type? We already have this use case: https://github.com/plone/plone.restapi/pull/931

Maybe we should put in place something similar to the portal transforms chaining?

tiberiuichim:
@sneridagh I can convert the adapters to subscribers, then we can discover and run them all. While I think it would be nice to have a possibility to declare a pipeline of transformation, maybe we can just declare a "sort position" value in each transfomer?

In any case, I've been thinking more and I'd like to rename it back to "Transformer". We need ``IBlockFieldInboundTransformer`` and also ``IBlockFieldOutboundTransformer``. My use case is this: my block field values depend on the result of a query in a database. It makes sense to put them there, at the outbound.

sneridagh:
I would explorenthe possibility of doing it like the trasnform chain does.

+1 to the splitting in two, Inbound/Outbound. However isn’t the Serializer/Deserializer naming more appropiate?

sneridagh:
I dig into Plone's code:
https://github.com/plone/plone.transformchain/blob/master/plone/transformchain/transformer.py#L48

The only question is how to introduce the block name in the multiadapter... @csenger had some ideas.

csenger:
> I dig into Plone's code:
> https://github.com/plone/plone.transformchain/blob/master/plone/transformchain/transformer.py#L48
> 
> The only question is how to introduce the block name in the multiadapter... @csenger had some ideas.

transformchain does what @tiberiuichim mentioned.The adapter has an attribute "order" that is used for sorting. 

I don't know if subscribers will work. There are no named subscribers IIRC. 

@tiberiuichim I'd go with `IBlockFieldSerializationTransformer` and `IBlockFieldDeserializationTransformer` to keep restapi's naming.

src/plone/restapi/interfaces.py
```diff
@@ -84,6 +84,38 @@ def __call__(value):
         """
 
 
+class IBlockFieldDeserializationTransformer(Interface):
+    """Convert/adjust raw block deserialized value into block value.
+    """
+
+    order = Attribute("A number used in sorting value transformers. "
```
csenger, line 91: You should add block_type as an attribute to the interface class

tiberiuichim: Good catch, thank you!

tiberiuichim:

tiberiuichim:

tisto:
@tiberiuichim I'd like to merge and release this PR today. Travis is failing due to black. Could you run black or do you want me to do it?

tisto:
@tiberiuichim I can't run black easily because you did a fork.

tisto:

Issue #911, opened 2020-04-15
Make @querystring-search context aware


sneridagh:
Follow up with:
https://github.com/plone/volto/issues/1166

The issue: In a Volto listing block, it's easy to come up with a query that lists also the current object, which is weird. This is happening in Plone as well, but it's eased by the fact that the queries are done only from `Collections` and that you can filter per content type. The freedom that the almighty Volto listing block features is the one to blame :)

So by allowing the endopoint to be context aware, then we know the element we want to exclude.

sneridagh:
Ok I was working on it this morning. These are my findings:
The initial idea was to filter the results "the dirty way" after the querystringbuilder call:

https://github.com/plone/plone.restapi/blob/a58394e95cdb9e90f87a034af96bcaa3bf296e0b/src/plone/restapi/services/querystringsearch/get.py#L31-L39

with a comprehension list, (omitting the boring implementation). However, querybuilder results can return different objects, LazyMaps, Batch etc... for which we have serializers, but we don't have a serializer for brains from LazyMaps, so it breaks. **(returning here later)

Then, I recalled about the `not` ZCatalog capabilities, so I tried it out using the UID of the element we want to exclude... 

```
(Pdb) catalog({'portal_type': {'query': ['Document']}, 'path': {'query': ['/plone/']}, 'sort_limit': 1000, 'UID': {'not': '83011c5b147647ccaa6f5666fef6168c'}})
*** ValueError: index 'UID': option 'not' is not valid
```
😩 

It would have been perfect, since we could use the custom_query override for the querybuilder.

Next thought, override the `path`, but then we will overlap with the path queries, and the override works by completely override the index, so dead end.

Next thought, we could add the feature to `p.a.querystring` querybuilder for exclude the context tapping in the parser, merge with the others path queries.

Other option would be to add `not` capabilities to the `UUIDIndex`, which I don't know what arcane knowledge you need to treasure to unleash that.

Last option would be to continue with the first approach (** the filter approach), then re-wrap the result again in a LazyMap or Batch (if that's possible) or create a custom serializer for plain lists with brains.

/cc @buchi @lukasgraf @tisto @jensens @pbauer any insight?

sneridagh:
Finally fixed the UUIDIndex 😁
https://github.com/zopefoundation/Products.ZCatalog/pull/99

PR #916, opened 2020-04-17
Make the querystring-search endpoint context aware

sneridagh:
The tests will fail until https://github.com/zopefoundation/Products.ZCatalog/pull/99 is merged and released.

### DIFF
--- a/src/plone/restapi/services/querystringsearch/configure.zcml
+++ b/src/plone/restapi/services/querystringsearch/configure.zcml
@@ -10,4 +10,11 @@
     permission="zope2.View"
     />
 
+  <plone:service
+    method="POST"
+    for="Products.CMFCore.interfaces.IContentish"
+    factory=".get.QuerystringSearchPost"
+    name="@querystring-search"
+    permission="zope2.View"
+    />
 </configure>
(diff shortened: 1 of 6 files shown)


sneridagh:
Fixes: https://github.com/plone/plone.restapi/issues/911

sneridagh:

src/plone/restapi/services/querystringsearch/configure.zcml
```diff
@@ -10,4 +10,11 @@
     permission="zope2.View"
     />
 
+  <plone:service
```
tisto, line 13: We need to keep the existing one because the Root does not implement IContentish, right?

src/plone/restapi/services/querystringsearch/get.py
```diff
@@ -38,6 +48,13 @@ def reply(self):
             limit=limit,
         )
 
+        if SUPPORT_NOT_UUID_QUERIES and not (IPloneSiteRoot.providedBy(self.context)):
```
tisto, line 51: Maybe a code comment: "# Exclude "self" content item from the results when ZCatalog supports NOT UUID queries and is called on a content object." Now that I read it, maybe we should do the IPloneSiteRoot check first and amend the comment accordingly. Though that's nitpicking, we can do that after the merge as well.

sneridagh: Done!

sneridagh:

tisto:
@sneridagh we should pin ZCatalog to 5.1 to make the test pass and also see if that version fails for older Plone versions.

sneridagh:
@tisto it seems it passes! Should we set it as a hard dependency (setup.py) then?

sneridagh:

PR #921, opened 2020-04-21
Add dx controlpanel integration test

tisto:


tisto:
@avoinea I created some integration tests for the dx control panel. Can you please have a look? I commented out some test code that would fail right now. I guess you are just serializing what Plone gives you. Though, from an API consumer point of view, this is what I would expect. Not sure how much work it would be to fix this and I guess it is not super important for the functionality we are providing. Just give me a feedback on what you think. :)

avoinea:
@tisto LGTM

avoinea:
I will fix the remaining todos and then we can merge :wink:

PR #922, opened 2020-04-23
Install black

tisto:


ale-rt:
Generally when I had black to my packages I have to add the following options to buildout.cfg to make it live happily with flake8 and code-analysis:
```
[code-analysis]
flake8-ignore = C901,E203,W503
flake8-max-line-length = 88
```
The check I use is:
```
black --fast --check setup.py src/
```

And I have to had the following lines to setup.cfg:
```
[flake8]
max-line-length = 88
inline-quotes = "
ignore = C901,E203,W503
```

I hope this helps :)

tisto:
@ale-rt thanks! I had the ignores and max-line-length configured. No idea why this is not working. Will investigate...

jensens:
While at it, I would also add a Black and Plone `isort` config:
```
[isort]
force_alphabetical_sort = True
force_single_line = True
not_skip = __init__.py
multi_line_output = 3
include_trailing_comma = True
force_grid_wrap = 0
use_parentheses = True
line_length = 88
```

PR #925, opened 2020-04-26
Adapt imagefield serializer tests 

erral:
This fixes are needed to fix the tests with the original image scale handling fixes in plone.namedfile

Those changes are in this PR and both branches should be tested together:
https://github.com/plone/plone.namedfile/pull/92

### DIFF
--- a/news/932.bugfix
+++ b/news/932.bugfix
@@ -1,0 +1,2 @@
+Adapt tests to the new way of handling original image urls
+[erral]
--- a/src/plone/restapi/tests/test_dxfield_serializer.py
+++ b/src/plone/restapi/tests/test_dxfield_serializer.py
@@ -22,12 +22,7 @@ from zope.interface.verify import verifyClass
 
 import os
 import six
-
-
-if PLONE_VERSION.base_version >= "5.1":
-    GIF_SCALE_FORMAT = "png"
-else:
-    GIF_SCALE_FORMAT = "jpeg"
+import unittest
 
 
 class TestDexterityFieldSerializing(TestCase):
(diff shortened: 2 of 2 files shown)


tisto:
@erral ok, the Travis tests for Plone 4.3 and 5.1 fail, which was expected. Shall we amend the tests or just don't run them on 4.3 and 5.1?

erral:
Oh, I see that there are no 4.3 tests in Jenkins here. If the plone.namedfile changes will not be backported to Plone 4.3 version, we will need indeed to amend the tests.

erral:
And the same happens for Plone 5.0 and 5.1.

So I guess we will need 2 different URL checks in the test, one for 5.2 and above and the other for <5.2, right?

erral:
I have reworked the image field serializer tests to check all available different cases:

- Plone < 5.1: original image is a scaled image in JPG format and corrupted images do not create a download url for the original image, because it's created creating a scale of the same size.

- Plone == 5.1: original image is a scaled image in PNG format and corrupted images do not create a download url for the original image, because it's created creating a scale of the same size.

- Plone >= 5.2: original image is the actual original image in its format and corrupted images do create a download url with the actual original image.

To achieve that I have extracted the relevant test cases to a specific TestCase and skipped using unittest.skipUnless when the relevant Plone version is running the tests.

erral:

erral:

erral:
The Travis tests on Plone 5.2 should fail, because we are testing against a released version of Plone 5.2 and not against the buildout.coredev. But the Jenkins tests on 5.2 should be green.

tisto:
@erral thank you Mikel! Will check later...

PR #927, opened 2020-04-27
Fixes #926 - Add type-schema adapters for: Email, URI and Password

### LINKED ISSUE #926: Missing type-schema adapters
We need to add schema adapters to https://plonerestapi.readthedocs.io/en/latest/types-schema.html for the following fields:
* plone.schema.Email
* zope.schema.URI
* zope.schema.Password

as there is almost impossible to distinguish them from TextLine field.

avoinea:
See #926


avoinea:

avoinea:

src/plone/restapi/tests/test_types.py
```diff
@@ -311,6 +312,81 @@ def test_float(self):
             adapter.get_schema(),
         )
 
+    def test_email(self):
+        field = Email(
+            title=u"Email",
+            description=u"Email field",
+            default="foo@bar.com",
+            min_length=10,
```
tisto, line 320: @avoinea if I am not mistaken min/max length does not really work in Volto, right? I mean if you set min/max length on the backend and Volto ignores it the REST API will raise an error if a user does not enter a text within a limit. Did you amend Volto to take min/max length into account on the frontend?

src/plone/restapi/tests/http-examples/controlpanels_get_dexterity_item.resp
```diff
@@ -111,7 +111,6 @@ Content-Type: application/json
       }, 
       "description": {
         "description": "", 
-        "minLength": 0,
```
tisto, line 114: @avoinea ok, it makes sense to remove that if the value is 0. Technically this is a breaking change but I guess nobody uses that in the first place.

src/plone/restapi/types/adapters.py
```diff
@@ -110,38 +113,57 @@ def get_type(self):
...
 @implementer(IJsonSchemaProvider)
 class TextJsonSchemaProvider(TextLineJsonSchemaProvider):
-    def additional(self):
-        info = {}
-        if self.field.min_length is not None:
-            info["minLength"] = self.field.min_length
+    def get_widget(self):
+        return "textarea"
 
-        if self.field.max_length is not None:
```
tisto, line 125: @avoinea doesn't "Text" require a min/max length attribute as well or am I missing something here?

src/plone/restapi/tests/test_types.py
avoinea: @tisto Yes, my colleague @rexalex is working on it on https://github.com/plone/volto/issues/1246

src/plone/restapi/types/adapters.py
avoinea: @tisto This diff is not that accurate, I moved the additional method within the parent class `TextLineJsonSchemaProvider`

avoinea:
> I guess you already checked that the Volto form generation and stuff still works with this right? 

@tisto Yes. Right now the new widgets use the TextWidget as fallback, but we will also add a PR on Volto, to improve rendering and validation for email, url, password and number fields :wink:

avoinea:
@tisto Hmm, I don't think travis is complaining because of this PR changes :thinking:

avoinea:

avoinea:

Issue #937, opened 2020-05-07
Given a translated object, calling the @translations expansion throws 401


sneridagh:
if the object is private (anonymous or the user has no permissions over it).

sneridagh:
@erral 

https://github.com/plone/plone.restapi/blob/b51c942f23babbacce7d8f5536471941528a24bf/src/plone/restapi/services/multilingual/pam.py#L37-L41

It seems that get_translations does a getObject() on unrestricted search. How does it work in PAM for it not to break?

erral:
p.a.m does not render direct URLs of content objects, but creates a link with the translation UID which in the end redirects the user to the relevant page and if the relevant page is private the log-in page appears.

erral:
Here are the relevant bits: https://github.com/plone/plone.app.multilingual/blob/master/src/plone/app/multilingual/browser/selector.py#L104

sneridagh:
Yeah, I see... how would you fix this? Since we don't need the getObject() maybe we can just do the unrestrictedSearch, then return the brain path. If the user goes there, then it will be a 401 in the destination object... However, that does not work well for "not yet published and being translated" sites.

sneridagh:
For that last use case, I would return the site root instead. Although I see that PAM does some crazy "give me the nearest translated thing"...

sneridagh:
Then maybe use the `get_restricted_translations` method, right? What do you think? Sorry for the brain storm :P

erral:
I would go for a `_checkPermission(View)` check. If the user querying the API has no permission to see the translation whatever the reason is. 

I don't know if this would be a penalty performance-wise, but we can check it.

erral:
Mmmm, I see... depending on the translation policy selected we may need to go to the "nearest translated content". We have a setting to check that on language control panel.

We would need to check that setting, if the setting is "nearest translated content", your way would be OK. If not, I would not return the translation

PR #938, opened 2020-05-07
Fixed @translations endpoint to only retrieve the translations that t…

sneridagh:
…he current user can really access


sneridagh:
Fixes #937

sneridagh:
@erral I think that for now that is enough and fixes all the use cases. No adaptation is required in Volto to support this behavior right away, which is good.

Regarding the nearest translation redirect I think it's too much... clients don't want this, since it could lead to the wrong content type and is confusing for the editors and for the users.

 If I could call my 10 years ago me and tell this to myself... 😂

sneridagh:

PR #944, opened 2020-05-12
Layout based serialization, include default_page in serialization

tiberiuichim:
This PR makes it possible to use a different content serializer, based on the ``layout`` property of context.

It also serializes fully serializes the linked ``default_page`` content, where it exists.

### DIFF
--- a/src/plone/restapi/serializer/dxcontent.py
+++ b/src/plone/restapi/serializer/dxcontent.py
@@ -2,7 +2,6 @@
 from AccessControl import getSecurityManager
 from Acquisition import aq_inner
 from Acquisition import aq_parent
-from Products.CMFPlone.utils import base_hasattr
 from plone.autoform.interfaces import READ_PERMISSIONS_KEY
 from plone.dexterity.interfaces import IDexterityContainer
 from plone.dexterity.interfaces import IDexterityContent
@@ -19,6 +18,7 @@ from plone.restapi.serializer.nextprev import NextPrevious
 from plone.rfc822.interfaces import IPrimaryFieldInfo
 from plone.supermodel.utils import mergedTaggedValueDict
 from Products.CMFCore.utils import getToolByName
+from Products.CMFPlone.utils import base_hasattr
 from zope.component import adapter
 from zope.component import getMultiAdapter
 from zope.component import queryMultiAdapter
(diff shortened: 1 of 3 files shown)


tiberiuichim:

src/plone/restapi/serializer/dxcontent.py
```diff
@@ -80,7 +80,8 @@ def __call__(self, version=None, include_items=True):
         # Insert field values
         primary_field_name = self.get_primary_field_name()
         for schema in iterSchemata(self.context):
-            read_permissions = mergedTaggedValueDict(schema, READ_PERMISSIONS_KEY)
+            read_permissions = mergedTaggedValueDict(
```
tisto, line 83: @tiberiuichim is it possible that you have flake8 with an 80 chars limit enabled in your IDE? We currently let black handle the line breaks and increased flake8 max-line-length to 200.

src/plone/restapi/serializer/dxcontent.py
```diff
@@ -185,7 +188,15 @@ def __call__(self, version=None, include_items=True):
                 )(fullobjects=True)["items"]
             else:
                 result["items"] = [
-                    getMultiAdapter((brain, self.request), ISerializeToJsonSummary)()
+                    getMultiAdapter((brain, self.request),
+                                    ISerializeToJsonSummary)()
                     for brain in batch
                 ]
+
+        page = self.context.getDefaultPage()
```
tisto, line 196: Nitpick: I'd rename "page" to "default_page" to be more clear.

src/plone/restapi/serializer/dxcontent.py
```diff
@@ -185,7 +188,15 @@ def __call__(self, version=None, include_items=True):
...
+        if page:
+            child = self.context._getOb(page)
```
tisto, line 198: Just wondering: why do you use _getOb instead of just self.content.get(page)?

src/plone/restapi/serializer/dxcontent.py
```diff
@@ -185,7 +188,15 @@ def __call__(self, version=None, include_items=True):
...
+            result['default_page'] = getMultiAdapter(
```
tisto, line 199: So we serialize the entire page instead of just one providing a link to the actual default page, right? I first thought about the performance penalty. Though, if no default page is set I guess we should be fine. Could you explain the reasoning behind why you choose to serialize the full object instead of providing a link to the default object?

tiberiuichim: This is to avoid an extra async call.

The idea is this: in Volto View.jsx can select the view component to be used based on this algorithm: 
```
    const RenderedView =
      this.getViewByType() || this.getViewByLayout() || this.getViewDefault();
```
If for some reason it finds that it should use the "default page content", then it would have to do an extra fetch for the content at that location. If that content is already in ``default_page`` attribute, then it can use that to render the proper view implementation.

wesleybl:
@tisto Is the idea good for you? If so, I can make the requested changes and implement the tests.

PR #955, opened 2020-06-03
PATCH (editing) in @user endpoint now is able to remove existing valu…

sneridagh:
…es using null


sneridagh:

Issue #956, opened 2020-06-11
plone.restapi 6.13.1 does not start in Plone4


gbastien:
Hi @tisto @sneridagh 

I think #940 breaks fix reported in #694, in Plone4, plone.restapi fails to start with error :
ConfigurationExecutionError: <class 'zope.component.interfaces.ComponentLookupError'>: (<InterfaceClass zope.security.interfaces.IPermission>, 'plone.app.controlpanel.Overview')

This is because now controlpanels service is only loaded in Plone5 and the configure.zcml into controlpanels is doing the job of including permissions.zcml but is no more called.

I think moving this from services/controlpanels/configure.zcml to services/configure.zcml would fix the problem.  Or at least add the same fix to database service.

What do you think?

Thank you for great addon!

Gauthier

tisto:
@gbastien I haven't had time to look into this in-depth but what you write makes sense. What I don't get it why our Plone 4 tests do not catch this. A PR with a fix would be welcome!

gbastien:
@tisto we have the same behavior, tests pass (imio.restapi/plonemeeting.restapi packages) but instance fails to start...  this is related to package order, I do not know exactly difference regarding this between test and instance...

tisto:
@gbastien yeah. I suspected something like this. It would be nice to have a test failing but this is way too much work to figure out, so I am good with a simple fix. :)

gotcha:
Order of packages issues are often due to/hidden by `z3c.autoinclude` (I dunno if it is the case here). This is the reason why I try to not forget to explicitly include stuff in ZCML.

tisto:
@gotcha yeah, thanks for the hint! I suspected something like this but I do not really have time to look into this right now. It bugs me that our test setup was not able to catch this regression.

PR #960, opened 2020-06-12
Include plone.app.controlpanel permissions.zcml in database service

gbastien:
to avoid ConfigurationExecutionError regarding 'plone.app.controlpanel.Overview' permission while starting Plone 4.3.x, see #956

I added the same fix explained in #526 in configure.zcml of services/database

I think yhis could be done in the services/configure.zcml where we already include CMFCore permissions because service "system" that also use the permission would have the same problem if permission was not loaded in database configure.zcml

So if you think it is better to move it to the parent's configure.zcml, I may change it.

Thank you for review and merge!

Gauthier


tisto:
Thank you @gbastien!

Issue #970, opened 2020-06-18
Parameter "metadata_fields" of serializer.catalog.LazyCatalogResultSerializer is not used


gbastien:
Hi @tisto

me again ;-)

In search/handler.SearchHandler.search, the metadata_fields is managed to be passed to ISerializeToJson serializer (LazyCatalogResultSerializer).

The LazyCatalogResultSerializer receives the metadata_fields but does nothing with it, instead the called ISerializeToJson that renders the brain get the metadata_fields again in the request.form in the metadata_fields method.

Where it does again the same work more or less than in SearchHandler.

As the DefaultJSONSummarySerializer is doing it, to me we should just remove metadata_fields as a parameter and let it get it from request.form.

For now, it only work if in request.form.

I can propose a PR to clean this because I was lost into this and I suppose other could be also.

Thank you,

Gauthier

tisto:
@gbastien thanks for working on p.restapi and contributing! Keep it going! :)

PR #972, opened 2020-06-24
Removed useless management of metadata_fields in SearchHandler/LazyCatalogResultSerializer

gbastien:
... since it is handled in DefaultJSONSummarySerializer, see #970

Removing it from search still let tests pass as it was not used, computed in search then computed again in DefaultJSONSummarySerializer...

This cleaning is necessary to avoid using it in search using the parameter as it is not taken into account and done after from request.form

Please review, comment, merge, ... ;-)

Thank you!

Gauthier

### DIFF
--- a/src/plone/restapi/search/handler.py
+++ b/src/plone/restapi/search/handler.py
@@ -74,16 +74,12 @@ class SearchHandler(object):
         else:
             fullobjects = False
 
-        metadata_fields = query.pop("metadata_fields", [])
-        if not isinstance(metadata_fields, list):
-            metadata_fields = [metadata_fields]
-
         self._constrain_query_by_path(query)
         query = self._parse_query(query)
 
         lazy_resultset = self.catalog.searchResults(query)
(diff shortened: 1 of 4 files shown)


gbastien:

src/plone/restapi/search/handler.py
```diff
@@ -74,16 +74,12 @@ def search(self, query=None):
         else:
             fullobjects = False
 
-        metadata_fields = query.pop("metadata_fields", [])
-        if not isinstance(metadata_fields, list):
```
tisto, line 78: @gbastien question: is it safe to remove this isinstance check here? @sunew fixed an issue with this some time ago: https://github.com/plone/plone.restapi/commit/22d610e486aa4cd85ba87c85263cd3e68aea59cc

Since Sune added tests for this I guess it should be safe to remove. Though, maybe we need to double check if that use case is properly solved without this check...

tisto: Ok. I should read the description before I guess. Found the metadata_field method in summary.py. Sorry for the noise and thank you for your PR!

PR #973, opened 2020-06-26
remove the use of plone.api in upgrade code

### LINKED ISSUE #971: plone.api dependency in upgrade code
There is an undeclared plone.api dependency in [upgrade.py](https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/upgrades/ordering.py#L3)

I ran into this when installing plone.restapi in Plone 4.3.x

IIRC it was decided that it was not acceptable to have plone.api dependency except for tests, right?

erral:
Fixes #971

### DIFF
--- a/news/917.bugfix
+++ b/news/917.bugfix
@@ -1,0 +1,2 @@
+Remove the use of plone.api in upgrade code
+[erral]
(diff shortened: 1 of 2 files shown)


erral:

PR #983, opened 2020-08-24
fix #411 serializing default_page item property into something like  …

### LINKED ISSUE #411: Provide default_page and layout properties in content GET
Needed to be able to render the default page or the correct view in the frontend.

Example:
```
{
    "@id": "http://localhost:8080/Plone",
    "@type": "Plone Site",
    "default_page": "http://localhost:8080/Plone/front-page",
    ...
}
```

or make it an expandable component?

```
{
    "@id": "http://localhost:8080/Plone",
    "@type": "Plone Site",
    "@components": {
        "properties": {
            "@id": "http://localhost:8080/Plone/@properties",
            "items": [
                 "default_page": "http://localhost:8080/Plone/front-page"
            ]
        },
    ...
}
```

See also #126

@tisto @sneridagh @ebrehault opinions?

sauzher:
…`"default_page": {"@id": "http://localhost:8080/Plone/front-page"}`

### DIFF
--- a/news/411.feature
+++ b/news/411.feature
@@ -1,0 +1,5 @@
+New feature:
+
+`default_page` property is now serialized in GET requests. Ex:
+
+``"default_page": {"@id": "http://localhost:8080/Plone/front-page"}``
--- a/src/plone/restapi/serializer/atcontent.py
+++ b/src/plone/restapi/serializer/atcontent.py
@@ -57,7 +57,12 @@ class SerializeToJson(object):
         result.update(
             {"previous_item": nextprevious.previous, "next_item": nextprevious.next}
         )
-
+        
+        # Insert default_page information
+        default_page = self.context.getDefaultPage()
+        if default_page is not None:
+            result.update({'default_page': {'@id': obj.absolute_url() + '/' + default_page}})
+     
         # Insert expandable elements
         result.update(expandable_elements(self.context, self.request))
 
--- a/src/plone/restapi/serializer/dxcontent.py
+++ b/src/plone/restapi/serializer/dxcontent.py
@@ -74,7 +74,12 @@ class SerializeToJson(object):
         result.update(
             {"previous_item": nextprevious.previous, "next_item": nextprevious.next}
         )
-
+        
+        # Insert default_page information
+        default_page = self.context.getDefaultPage()
+        if default_page is not None:
+            result.update({'default_page': {'@id': obj.absolute_url() + '/' + default_page}})
+            
         # Insert expandable elements
         result.update(expandable_elements(self.context, self.request))
 
--- a/src/plone/restapi/serializer/site.py
+++ b/src/plone/restapi/serializer/site.py
@@ -57,6 +57,11 @@ class SerializeSiteRootToJson(object):
 
         # Insert expandable elements
         result.update(expandable_elements(self.context, self.request))
+        
+        # Insert default_page information
+        default_page = self.context.getDefaultPage()
+        if default_page is not None:
+            result.update({'default_page': {'@id': self.context.absolute_url() + '/' + default_page}})        
 
         result["items_total"] = batch.items_total
         if batch.links:


sauzher:
I really do not know how to solve this https://travis-ci.org/github/plone/plone.restapi/jobs/720903264#L346.
(actually I'm not getting what is the problem)
any hint?

tisto:
@sauzher please run black on your code base:

```
bin/black src/
```

Improving the CI error message has been on my list for quite a while. Travis is limited though.

sauzher:

docs/source/conf.py
```diff
@@ -18,18 +18,18 @@
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
-#sys.path.insert(0, os.path.abspath('.'))
+# sys.path.insert(0, os.path.abspath('.'))
```
tiberiuichim, line 21: No need to run black on this file, I think it's autogenerated and hasn't been changed since.

tiberiuichim (approved):
I think most useful is that when you include the default page you also serialize it in the response. This would avoid needing to fetch it again. Maybe implement it as an expander?

sauzher:
@tiberiuichim thanks for the pull review.

> I think most useful is that when you include the default page you also serialize it in the response. This would avoid needing to fetch it again. Maybe implement it as an expander?

I don't know the desiderata behind `expander` section.
But, IMHO the default_page must be considered as a simple object property. actually it is only that. My point is that it is a task of the frontend engine choosing how to handle this information i.e. fetching the default_page object if needed just like any other component of the context.

tiberiuichim:
@sauzher Indeed, it is the job of the frontend to decide if it wants this information. With an expander, it can do that:

```
curl http://localhost:3000/somefolder?expand=default_page
```

So it would get the "context", but also whatever content it needs to use to properly show the page. What should be included in the `default_page` expander? Just fully serialize that page, so that it can be used to show the proper content.

docs/source/conf.py
tisto: We restrict black to run on the /src directory so far to avoid this. Therefore I agree with Tiberiu here.

sauzher: gotcha. reverted in 7b40aa94c01d58b4dc27dc7b2cec55c15a1c1b7d

sauzher:
> @sauzher Indeed, it is the job of the frontend to decide if it wants this information. With an expander, it can do that [...]

@tiberiuichim Thanks,  I was unaware of this possiblity. 

Dunno what to say... honestly, at this time, I personally would prefer a two step approach on fetching a container and its default_page.  They are still two separate context with theirs own components information (workflows, actions, etc.).  Having ID (in the default_page property) and its title, description, type properties available in the `items` section I think it might be enough. 

Perhaps dealing with containers with thousands items (or more) it may be better having  default_page's metadatas directly into the new `default_page` section?

Fetching the `container` first and then `container?expand=default_page`  having all the default_page information nested somewhere in the json i think is in someway confusing. But I admit it's quite subjective, though.

tiberiuichim:
Just throwing https://github.com/plone/plone.restapi/pull/944 into discuss. I'm ok either way

sauzher:

sauzher:

thomasmassmann:
For the deserializer, we currently have a PATCH service for a customer project:

```python
class SetDefaultPage(Service):
    """Set the default page for the given content object."""

    def reply(self):
        data = json_body(self.request)
        default_page = data.get("default_page", None)
        try:
            self.context.setDefaultPage(default_page)
        except Unauthorized as exc:
            self.request.response.setStatus(403)
            return dict(error=dict(type="Forbidden", message=str(exc)))
        except BadRequest as exc:
            self.request.response.setStatus(400)
            return dict(error=dict(type="Bad Request", message=str(exc)))

        return self.reply_no_content()
```

PR #992, opened 2020-09-23
Pin pyrsistent to 0.15.7 in buildout configs.

mauritsvanrees:
This is the last pyrsistent version that is python 2 compatible.
On 5.2 it is pinned in the core versions.
It is a dependency of jsonschema.

Just a Travis test fix, so no changelog entry, and no Jenkins jobs needed.


mauritsvanrees:
Note that on master, the 4.3 and 5.1 jobs are failing:
https://travis-ci.org/github/plone/plone.restapi/builds/727355540
I noticed it on a PR.

From the looks of it, this worked fine before, but `pyrsistent` recently had some releases that were for Python 3 only, and started containing actual SyntaxErrors on Python 2. See https://pypi.org/project/pyrsistent/0.15.7/#history

So no one is to blame, it is just a missed version pin.

PR #993, opened 2020-09-25
Fix resolveuid

tisto:


sneridagh:
@tisto I've added new functional tests, and fixed the failing ones. 

The use case for the image bloc type (url attribute for the image itself, href for the target link) is also covered.

It's due to add a changelog, remove the print statements.

cekk:
lgtm except for the prints ;)

btw i don't remember why i looked into the wrong field.

sneridagh:

tisto:

sneridagh:
@tisto weird, why that Jenkins job fail in such places where the others don't

tisto:
@sneridagh the robot tests on Jenkins are very fragile and do not run reliably.

tisto:

sneridagh:

mauritsvanrees:
Jenkins seems reasonable to me currently. I remember both better times and worse times. Most unrelated failures seem to occur under high load, so when lots of jobs are run together, which easily happens when you run PR jobs for 5.2 and 6.0.

The PR jobs seem more prone to random failures than the main jobs, probably because the main jobs are split between robot tests and other tests.

Most random failures are in robot tests. Some are because no browser is open, which may be some race condition in the browser setup. For others some content is not found on the page, which may have any number of reasons.

It would be nice if we could catch failures and automatically retry those once or twice. I did some work on that in the beginning of this year, but it never really worked well. See:

- PR for retrying all failed robot tests in coredev: https://github.com/plone/buildout.coredev/pull/636
- PR for retrying failed robot tests specifically in plone.app.multilingual: https://github.com/plone/plone.app.multilingual/pull/369

Better would be if the test runner (I think `zope.testrunner` and/or `collective.xmltestreport`) would support retrying failed tests.

sneridagh:

sneridagh:

plone-5.2.x.cfg
```diff
@@ -9,6 +9,7 @@ versions=versions
 plone.restapi =
 plone.namedfile = 5.3.1
 astunparse = 1.6.2
+pep517 = 0.9.1
```
wesleybl, line 12: This was duplicated after the master merge. See:

https://github.com/plone/plone.restapi/blob/64efe43ea48695902e8579fc5ec9d1d375f03238/plone-5.2.x.cfg#L26

sneridagh: oh ok, thanks for the heads up

sneridagh:

sneridagh (approved):
LGTM let's ship it!

src/plone/restapi/deserializer/blocks.py
```diff
@@ -104,20 +109,22 @@ def __call__(self, block):
...
         # Convert absolute links to resolveuid
-        # Assumes in-place mutations
-
+        #   http://localhost:55001/plone/link-target
+        #   ->
+        #   ../resolveuid/023c61b44e194652804d05a15dc126f4
         entity_map = block.get("text", {}).get("entityMap", {})
         for entity in entity_map.values():
             if entity.get("type") == "LINK":
-                href = entity.get("data", {}).get("href", "")
```
tisto, line 119: @sneridagh we really have to properly document when to user href and when to use URL. This won't be the last time we mix things up.

src/plone/restapi/deserializer/blocks.py
```diff
@@ -126,6 +133,7 @@ def __call__(self, block):
 class HTMLBlockDeserializer(object):
     order = 100
     block_type = "html"
+    disabled = os.environ.get("disable_transform_html", False)
```
tisto, line 136: @sneridagh that's also something we need to document.

Issue #995, opened 2020-09-28
@layout endpoint to expose default blocks / blocks_layout


avoinea:
Now that `blocks` and `blocks_layout` can have default values we need a way to expose these values as we do with [workflow](https://plonerestapi.readthedocs.io/en/latest/workflow.html), [actions](https://plonerestapi.readthedocs.io/en/latest/actions.html), etc.

For this, I propose adding expandable `@layout` component.

jensens:
Closing as outdated (5 years old). Default blocks/blocks_layout exposed via @types endpoint. Volto handles defaults on the frontend.
If you think this issue is still relevant please reopen it. Thank you for the original report!

PR #996, opened 2020-09-28
Add @layout endpoint

avoinea:
See #995


tiberiuichim (approved):

avoinea:
@tisto @sneridagh I need your thoughts on this one.

The idea is that we need a read-only endpoint with `permission="zope2.View"` to be able to access Content-Type's default values for `blocks` and `blocks_layout`.

This will allow us to upgrade DX CType Layout, for example, read-only blocks and have these changes reflected on existing instances of that DX CType.

Issue #999, opened 2020-09-29
Calls to `@workflow` fail when executing user has no permissions to access `review_history` in target state.


deiferni:
We have stumbled upon an edge-case when executing workflow transitions via the `@workflow` service endpoint when:
- the state variable `review_history` in the workflow is protected with a `guard-permission`
- executing a workflow-transition removes that permission for the executing user in the target state of the transition
- retrieving the workflow history in the endpoint after the transition has been executed will fail in https://github.com/plone/plone.restapi/blob/f01ec6b8896a76d2877bb0ef3a29fb7964fac3f3/src/plone/restapi/services/workflow/transition.py#L80 with the following traceback
   ```
   ...
  File "plone.restapi/src/plone/restapi/services/workflow/transition.py", line 80, in reply
    history = self.wftool.getInfoFor(self.context, "review_history")
  File "eggs/Products.CMFCore-2.4.8-py3.8.egg/Products/CMFCore/WorkflowTool.py", line 288, in getInfoFor
    raise WorkflowException(msg)
  Products.CMFCore.WorkflowCore.WorkflowException: Could not get info: ${name}
   ```
- so the issue surfaces after successful execution but while attempting to serialize the JSON-response

PR #1000, opened 2020-09-29
Fix @workflow when executing user has no permissions to access `review_history` in target state.

deiferni:
We have stumbled upon an edge-case when executing workflow transitions via the `@workflow` service endpoint when:
- the state variable `review_history` in the workflow is protected with a `guard-permission`
- executing a workflow-transition removes that permission for the executing user in the target state of the transition

An example workflow to illustrate the issue is provided in this PR. It has the following properties:
- two states `restricted` and `unrestricted` (default)
- two transitions `restrict` and `show` to change between the two states
- the `restricted` state removes the `View` permission for everyone but a `Manager` or a `Site Administrator`
- the variable `review_history` is protected with a `guard-permission` using the `View` permission

There is a test to illustrate the problem:
- when a user with none of the above roles executes the `restrict` transition from the `unrestricted` into the `restricted` state the transition is executed successfully, but
- retrieving the workflow history in the endpoint after the transition has been executed will fail in https://github.com/plone/plone.restapi/blob/f01ec6b8896a76d2877bb0ef3a29fb7964fac3f3/src/plone/restapi/services/workflow/transition.py#L80 with the following traceback
   ```
   ...
  File "plone.restapi/src/plone/restapi/services/workflow/transition.py", line 80, in reply
    history = self.wftool.getInfoFor(self.context, "review_history")
  File "eggs/Products.CMFCore-2.4.8-py3.8.egg/Products/CMFCore/WorkflowTool.py", line 288, in getInfoFor
    raise WorkflowException(msg)
  Products.CMFCore.WorkflowCore.WorkflowException: Could not get info: ${name}
   ```
- so the issue surfaces after successful execution but while attempting to serialize the JSON-response 

We provide a solution to this issue by accessing `review_history` with elevated privileges once the transition has been done using https://docs.plone.org/develop/plone/security/permissions.html#bypassing-permission-checks. This is safe because:
- only the last entry in the `review_history` is returned
- this is the workflow transition that the user just executed, they can know about it
- executing the transition itself is of course still done with normal privileges and will fail first if the user does not have the required permissions

Fixes https://github.com/plone/plone.restapi/issues/999


deiferni:

deiferni:

deiferni:

lukasgraf (approved):
LGTM 👍

src/plone/restapi/configure.zcml
```diff
@@ -38,6 +38,15 @@
...
 
+  <genericsetup:registerProfile
+      name="testing-workflows"
+      title="plone.restapi testing-workflows"
+      directory="profiles/testing-workflows"
+      description="Adds sample workflows for testing"
+      provides="Products.GenericSetup.interfaces.EXTENSION"
+      for="Products.CMFPlone.interfaces.ITestCasePloneSiteRoot"
+      />
+
```
jensens, line 49: the profile must be excluded in https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/setuphandlers.py#L21

jensens (changes requested):

deiferni:

deiferni:

src/plone/restapi/configure.zcml
jensens: @deiferni this need to be resolved before merge.

deiferni:

src/plone/restapi/configure.zcml
deiferni: @jensens i have amended the desired changes in https://github.com/plone/plone.restapi/blob/aa99141eeaeb742aa7c1852d48d373c09d1a26b1/src/plone/restapi/setuphandlers.py#L21 yesterday, but i am currently fighting with what to me seem to be unrelated test-failures. will request re-review when it is ready!

deiferni:
@jensens thanks for your feedback! tests are now green, ready for re-review. i've amended my last commit so that the profile is excluded.

jensens (approved):

tisto:
Please do not merge for now. I have to clean up the mess in master / 6.x.x branch first and make a release before we can merge and release this.

src/plone/restapi/configure.zcml
```diff
@@ -38,6 +38,15 @@
       for="Products.CMFPlone.interfaces.ITestCasePloneSiteRoot"
       />
 
+  <genericsetup:registerProfile
+      name="testing-workflows"
+      title="plone.restapi testing-workflows"
+      directory="profiles/testing-workflows"
+      description="Adds sample workflows for testing"
+      provides="Products.GenericSetup.interfaces.EXTENSION"
+      for="Products.CMFPlone.interfaces.ITestCasePloneSiteRoot"
```
ale-rt, line 47: IMHO if this profile is needed for the tests, it should be moved inside the test folder and the zcml should be included by the test setup.
In that way you will not even need to bother with the `getNonInstallableProfiles`

src/plone/restapi/services/workflow/transition.py
```diff
@@ -77,20 +90,40 @@ def reply(self):
...
             return dict(error=dict(type="Bad Request", message=str(e)))
 
-        history = self.wftool.getInfoFor(self.context, "review_history")
-        action = history[-1]
-        if six.PY2:
-            action["title"] = self.context.translate(
-                self.wftool.getTitleForStateOnType(
-                    action["review_state"], self.context.portal_type
-                ).decode("utf8")
+        sm = getSecurityManager()
```
ale-rt, line 93: I would use a context manager here

src/plone/restapi/configure.zcml
deiferni: I have followed the existing convention of where the test profiles are put, there is already a `testing` profile in the same directory. In principle i agree but i'd rather not change too much for just this hotfix, so if you don't object i'd prefer to leave it where it is.

src/plone/restapi/services/workflow/transition.py
deiferni: Ok, i have amended a context manager in my last commit and moved the whole security manager management into it.

deiferni:

deiferni:

deiferni:

deiferni:
@ale-rt thanks for your feedback, i have implemented or responded, in case you want to have another look. the tests are now also green again.

ale-rt (approved):
I am still of the idea that is a bad idea registering the testing profiles in the main `configure.zcml` and fiddle with `getNonInstallableProfiles` (see https://github.com/plone/plone.restapi/pull/1000/files#r530845175). Anyway it is ok for me to merge this PR in its current state!

ale-rt:
> Please do not merge for now. I have to clean up the mess in master / 6.x.x branch first and make a release before we can merge and release this.

@tisto can this be merged now?

jensens (approved):

deiferni:

deiferni:
Hey @tisto, I dropped the ball here for a while, but i'd like to pick this one up again.

- I have now rebased and made sure tests are green.
- Some things have changed in `black` so i have applied the necessary changes in a separate commit. I can also ship this in a separate PR if desired.

Ist there something still blocking this PR from your side? If not, please let me know what to do to move forward here. Thanks!

deiferni:

deiferni:
@tisto rebases after your black fixes in master and dropped local black fixes again. please let me know if anything else comes up and i will amend.

tisto:
@deiferni we can merge this now. I am just wondering about the testing profile in the main configure.zcml as @ale-rt mentioned previously. I don't have time right now to look into this right now. Though, won't the test profile show up in the available profiles to the user? Wasn't there a way to hide this?

deiferni:
@tisto we currently have the following strategy:
- the testing profile is defined in the same place as the other previously existing testing profile `plone.restapi:testing`
- it is then excluded in https://github.com/plone/plone.restapi/blob/04d3114ad25f5d0678405cb64fb8da4e2f5b2ac4/src/plone/restapi/setuphandlers.py#L21 like the other profiles, so it should not show up.

IIRC we agreed that moving all the testing profiles is out of scope for this bugfix. But i agree that having a separate `zcml` for testing only would by more elegant, so if you and @ale-rt want the profiles moved, i can do that first (separate PR probably). Let me know what you prefer!

deiferni:
thanks!

PR #1014, opened 2020-11-09
make update

tisto:


plone-5.1.x.cfg
```diff
@@ -5,16 +5,14 @@ extends =
...
+plone.testing = 5.0.0
+
 # fixes: SyntaxError: invalid syntax (more.py, line 340)
 zipp = 0.5.2
 
 # plone.restapi specific
 plone.schema = 1.2.1
 
-# Last pyrsistent version that is python 2 compatible:
-pyrsistent = 0.15.7
```
wesleybl, line 16: If this pin is removed in Plone 5.1, a version that is not compatible with Python 2 will be used. See:

https://github.com/tobgu/pyrsistent/blob/47ab1f38bd85a21b92e3df5a03cc81aa76c5884e/setup.py#L81

wesleybl: I run the buildout locally and Plone 5.1 doesn't use this package. So this removal is ok.

Issue #1019, opened 2020-11-12
Test failures with Zope 4.5.2 on Python 2.7


mauritsvanrees:
Yesterday I added Zope to the checkouts in plone coredev 5.2. That causes [test failures](https://jenkins.plone.org/job/plone-5.2-python-2.7/1162/#showFailuresLink) on Python 2.7 only.
Zope 4.5.2 was released today. Using it indeed gives the same problems.
This is caused by a change in headers in Zope. See [my comment](https://github.com/zopefoundation/Zope/pull/905#issuecomment-725992091) where I explain the current situation in more detail.

Note: the tests fail both with `plone.restapi` 6.x.x and master.

mauritsvanrees:
I have [reverted coredev to 4.5.1](https://github.com/plone/buildout.coredev/commit/75ccb20598a19228c07b9eff02991c5397761abd) for the moment.

mauritsvanrees:
Cause found, see [comment](https://github.com/zopefoundation/Zope/pull/905#issuecomment-726042099). We can fix it in `plone.restapi` code by making sure we set the Location header as a string, not unicode.

Warning for clarity: this seems not just a problem in the tests, but should be visible when you try it live with Zope 4.5.2.

PR #1020, opened 2020-11-12
Fix Zope 4.5.2 compatibility: make sure Location header is string [master]

mauritsvanrees:
On Python 2 it could be unicode for the users and groups end points.
Fixes https://github.com/plone/plone.restapi/issues/1019

### DIFF
--- a/news/1019.bugfix
+++ b/news/1019.bugfix
@@ -1,0 +1,4 @@
+Fixed compatibility with Zope 4.5.2 by making sure Location header is string.
+On Python 2 it could be unicode for the users and groups end points.
+Fixes `issue 1019 <https://github.com/plone/plone.restapi/issues/1019>`_.
+[maurits]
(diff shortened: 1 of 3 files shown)


mauritsvanrees:

tisto (approved):

tisto:
@mauritsvanrees merged and cherry-picked into the 6.x.x branch. I will wait for a green Jenkins and then make a 6.x.x release.

mauritsvanrees:
FTR, I started wondering if I should have used `iso-8859-1` instead of `utf-8`, as that is what HTTP requires in headers, according to the [original Zope PR](https://github.com/zopefoundation/Zope/pull/905). Then we would need to handle the case where encoding to `iso-8859-1` fails.
But in standard Plone 5.2 with Python 2 you cannot add users or groups with non-ascii in their id, you get a validation error. On Python 3 a non-ascii user cannot be created, but a group can. But the code in this PR does not change the Python 3 case.
So this should be good.
BTW, the original problem is [fixed in Zope](https://github.com/zopefoundation/Zope/pull/930).

PR #1023, opened 2020-11-12
Change the name of the profile blocks to volto-blocks

### LINKED ISSUE #895: plone.restapi 6.4.1 shows "Volto Blocks" installer in add-on products controlpanel
I have added plone.restapi 6.4.1 to a Plone 4.3.x buildout, and after fighting buildout, I see "Volto Blocks" profile as installable in add-on products control panel

2020-03-30 13-24-15-ko-pantaila-argazkia

I expected to see the default plone.restapi profile.

I guess that in Plone 4.3.x Plone shows the alphabetically first profile, and "blocks" goes before "default".

wesleybl:
Change the name of the profile `blocks` to `volto-blocks`, so that the install profile of `Plone 4.3` is the `default` profile.

In `Plone 4.3`, the default profile must be first in alphabetical order.

I tested it on a portal already had the `blocks` profile  installed and everything went well.

Fix #895


wesleybl:

idgserpro:

news/895.bugfix
```diff
@@ -0,0 +1,3 @@
+- Change the name of the profile blocks to volto-block, so that the install
```
mauritsvanrees, line 1: Typo: it is `volto-blocks` plural.
Could be fixed after merge too.

sneridagh:
Can you state why? I don't get the change. Also it will break a lot of projects that rely on it to be named exactly like it.

mauritsvanrees:
> Can you state why? I don't get the change. Also it will break a lot of projects that rely on it to be named exactly like it.

This solves https://github.com/plone/plone.restapi/issues/895
It is because of [these lines in the quick installer](https://github.com/plone/Products.CMFQuickInstallerTool/blob/3.0.16/Products/CMFQuickInstallerTool/QuickInstallerTool.py#L587-L595). When installing `plone.restapi` it takes all extension profiles, sorts them alphabetically, and picks the first one as the install profile. Without this PR, it picks `blocks` over `default`. So in Plone 4.3, in the `prefs_install_products_form` you do not see `plone.restapi`, but you see "Volto Blocks" in the list of modules.

In Plone 5 the quick installer tool is less and less involved, so it may work fine there, probably already in 5.0.

news/895.bugfix
wesleybl: @mauritsvanrees Done!

tisto:
FYI: I do not plan to merge this anytime soon. This has the potential to break lots of existing Volto projects.

wesleybl:
@tisto maybe break packages in the situation described by @avoinea :  https://github.com/plone/plone.restapi/issues/895#issuecomment-610923052

But the packages could be updated.

mauritsvanrees:
I was going to say "LGTM", but now I am wondering who would be responsible for applying the `blocks` or `volto-blocks` profile. I had expected that the default profile would depend on this one, but I see it is the other way around. (You can see I don't know much about the structure of `plone.restapi`.) In other words, there is probably documentation somewhere that tells people to add `plone.restapi:blocks` to their `metadata.xml`. This documentation will be wrong when this PR is merged, and people would have to change their code.

Traditional workaround is to add an old-style `Extensions/install.py` containing an `install` function that explicitly installs the `default` profile. The quickinstaller tool first tries this function before picking a profile.

wesleybl:
> In other words, there is probably documentation somewhere that tells people to add plone.restapi:blocks to their metadata.xml.

@mauritsvanrees I searched for documentation on this profile and didn't find it.

> Traditional workaround is to add an old-style Extensions/install.py containing an install function that explicitly installs the default profile. The quickinstaller tool first tries this function before picking a profile.

I will take a look at this.

Another alternative that I thought is to keep both profiles and put a zcml condition on them.

wesleybl:
> Traditional workaround is to add an old-style Extensions/install.py containing an install function that explicitly installs the default profile.

@mauritsvanrees the default profile is installed in Plone 4.3 when profile blocks are installed: See:

https://github.com/plone/plone.restapi/blob/28cd92e3d6f5adac738b068124a387089d72c160/src/plone/restapi/profiles/blocks/metadata.xml#L5

The problem is that profile blocks should not be installed.

@tisto we have already changed the name of this profile before:

https://github.com/plone/plone.restapi/commit/28cd92e3d6f5adac738b068124a387089d72c160#diff-397c7ccf64bf715e69e564921ae40d23dd2f2de0a221cf04bb368e977f6c1152L49

tisto:
@wesleybl this is not an issue that is open for discussion and I made my point clear multiple times already. The name change was a big thing back then and I guess back then we were the only company that relied on this, so we could do this move. Today we have lots of projects and multiple companies that rely on this profile and naming. As the maintainer of this package, I am not going to break things without careful consideration and planning.

PR #1027, opened 2020-12-03
Patch Plone 4.3's _install_profile_info method to list the default profile of plone.restapi in the first position.

### LINKED ISSUE #895: plone.restapi 6.4.1 shows "Volto Blocks" installer in add-on products controlpanel

wesleybl:
Plone 4.3 considers that the first profile, after ordering all profiles in alphabetical order, is the profile to be installed when installing
the product. Since plone.restapi has the profile `blocks`, it is installed, instead of the default profile. This patch puts the `default` profile in the first position, so that it is the installed profile.

Fix #895

### DIFF
--- a/src/plone/restapi/patch.py
+++ b/src/plone/restapi/patch.py
@@ -1,0 +1,31 @@
+# -*- coding: utf-8 -*-
+
+from Products.CMFQuickInstallerTool.QuickInstallerTool import QuickInstallerTool
+
+_original_QuickInstallerTool_install_profile_info = (
+    QuickInstallerTool._install_profile_info
+)
+
+
+def _install_profile_info(self, productname):
+    install_profile_info = _original_QuickInstallerTool_install_profile_info(
+        self, productname
+    )
+    if productname == "plone.restapi" and install_profile_info:
+        default_profile = u"plone.restapi:default"
+        first_profile = install_profile_info[0]["id"]
+        if default_profile != first_profile:
+            install_profile_info[0], install_profile_info[1] = (
+                install_profile_info[1],
+                install_profile_info[0],
+            )
+    return install_profile_info
+
+
+def apply_patch():
+    # BBB: In Plone 4, the profile that runs when installing a product is the first in
+    # the list of profiles, ordered alphabetically. Since plone.restapi has profile
+    # blocks, it comes before the default and that is why it is what is listed for
+    # installing of plone.restapi. This patch changes the order, of the profiles blocks
+    # and default, so that the default is the installed profile.
+    QuickInstallerTool._install_profile_info = _install_profile_info
--- a/news/895.bugfix
+++ b/news/895.bugfix
@@ -1,0 +1,4 @@
+Fix error in Plone 4.3 that installed the blocks profile  when installing the package,
+instead of the default profile.
+Fix `#895 <https://github.com/plone/plone.restapi/issues/895>`
+  [wesleybl]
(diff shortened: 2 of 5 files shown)


wesleybl:

avoinea:
Interesting approach :smile:

avoinea (approved):

wesleybl:

jensens:
Since there won't be any further 4.3.x release this should be save. 

OT:
But OTOH we should drop 4.3 support on master in plone.restapi soon and go Python 3 (Plone 5.2/6.0) only. If I look at the code in here its a mess with specialities to keep the suffering patient alive. It is OK to have a branch targetting the old releases and even bugfix it, but having it all in one is really no fun.

tisto:
I am ok with dropping Plone 4.3 and Python2 if that leads to significant additional efforts. Though, I'd like @lukasgraf and @buchi to chime in before we make such a move.

erral:
I'm also +1 on dropping Plone 4.3 support but after doing a last release that supports it correctly, for instance with the patch discussed here, that way Plone 4.3 users would have a supported version with a standard installation path.

lukasgraf:
Yeah, I can understand the headaches Py2/P43 compatibility on `master` causes.

Having a way to maintain a Plone 4.3 compatible line would indeed be important for us, for the time being.
But if we can get a Plone 4.3 compatibility branch to backport important fixes to, and as @erral described, a final release, that would work for us 👍

@tisto thanks for considering our situation! 😉

tisto:
Ok. I guess it is time to move on. I tried really hard to keep the Plone REST API branch free but the amount of work that I/we have to put into it is just too much and the two branches diverged from each other significantly already anyway. Plone REST API 7.x.x will drop Plone 4.3 support and maybe even more (needs to be decided). We will do a final pretty soon.

Though, we should move the 6.x.x branch to gh actions making sure we can continue to make releases and stuff on a stable and well-tested base. I will try to take care of this. 

I do not plan to abandon the 6.x.x branch and I will continue to make releases as long as others need them. I'd also be open to passing the 6.x.x release/review responsibilities to @lukasgraf, @buchi, or @erral if they step up. :)

src/plone/restapi/patch.py
```diff
@@ -0,0 +1,31 @@
...
+)
+
+
+def _install_profile_info(self, productname):
+    install_profile_info = _original_QuickInstallerTool_install_profile_info(
+        self, productname
+    )
+    if productname == "plone.restapi" and install_profile_info:
+        default_profile = u"plone.restapi:default"
+        frist_profile = install_profile_info[0]["id"]
```
tisto, line 16: @wesleybl typo: first_profile

wesleybl: Done.

wesleybl:

tisto:
@wesleybl this should go into the 6.x.x branch. We will drop Plone 4 support with plone.restapi 7.0.0.

wesleybl:

wesleybl:
@tisto since the master branch is no longer supported by Python 2, I pointed this PR to branch 7.x.x and rebased it. Can we do the merge? This change does not affect Plone 5.

mauritsvanrees (approved):
LGTM for 7.xx.

tisto:
Thanks @wesleybl!

PR #1009, opened 2020-10-30
Improve import traceback

jensens:
I got an difficult to interpret traceback on content POST like:
```
2020-10-30 15:50:32,222 ERROR   [Zope.SiteErrorLog:252][waitress-3] 1604069432.22186450.03333581790354878 http://localhost:8080/Plone/reviere/POST_application_json_
Traceback (innermost last):
  Module ZPublisher.WSGIPublisher, line 167, in transaction_pubevents
  Module ZPublisher.WSGIPublisher, line 364, in publish_module
  Module ZPublisher.WSGIPublisher, line 259, in publish
  Module ZPublisher.mapply, line 85, in mapply
  Module ZPublisher.WSGIPublisher, line 68, in call_object
  Module plone.rest.service, line 22, in __call__
  Module plone.restapi.services, line 20, in render
  Module plone.restapi.services.content.add, line 77, in reply
  Module plone.restapi.deserializer.dxcontent, line 45, in __call__
  Module plone.restapi.deserializer.dxcontent, line 151, in get_schema_data
  Module plone.app.relationfield.widget, line 30, in get
AttributeError: 'str' object has no attribute 'isBroken'
```
with this change it points me directly to the problem:
```
2020-10-30 15:50:32,222 ERROR   [Zope.SiteErrorLog:252][waitress-3] 1604069432.22186450.03333581790354878 http://localhost:8080/Plone/reviere/POST_application_json_
Traceback (innermost last):
  Module ZPublisher.WSGIPublisher, line 167, in transaction_pubevents
  Module ZPublisher.WSGIPublisher, line 364, in publish_module
  Module ZPublisher.WSGIPublisher, line 259, in publish
  Module ZPublisher.mapply, line 85, in mapply
  Module ZPublisher.WSGIPublisher, line 68, in call_object
  Module plone.rest.service, line 22, in __call__
  Module plone.restapi.services, line 20, in render
  Module plone.restapi.services.content.add, line 77, in reply
  Module plone.restapi.deserializer.dxcontent, line 45, in __call__
  Module plone.restapi.deserializer.dxcontent, line 151, in get_schema_data
   - __traceback_info__: field=kup.tfv.db.revier.revier.IRevierBehavior.bewirtschafter
  Module plone.app.relationfield.widget, line 30, in get
AttributeError: 'str' object has no attribute 'isBroken'
```

PR #1031, opened 2020-12-14
Revert "Improve import traceback"

tisto:
Reverts plone/plone.restapi#1009

### DIFF
--- a/plone-5.1.x.cfg
+++ b/plone-5.1.x.cfg
@@ -15,6 +15,9 @@ plone.schema = 1.2.1
 # Last pyrsistent version that is python 2 compatible:
 pyrsistent = 0.15.7
 
+# codeanalisys
+typing =3.7.4
+
 # zest.releaser
 zest.releaser = 6.20.1
 twine = 1.11.0
--- a/plone-5.1.x.cfg
+++ b/plone-5.1.x.cfg
@@ -16,12 +16,11 @@ plone.schema = 1.2.1
 pyrsistent = 0.15.7
 
 # codeanalisys
-typing =3.7.4
+typing = 3.7.4
 
 # zest.releaser
 zest.releaser = 6.20.1
 twine = 1.11.0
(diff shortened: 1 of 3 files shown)


tisto:

sneridagh:

plone-5.1.x.cfg
```diff
@@ -15,10 +15,12 @@ plone.schema = 1.2.1
 # Last pyrsistent version that is python 2 compatible:
 pyrsistent = 0.15.7
 
+# codeanalisys
+typing = 3.7.4
```
wesleybl, line 19: @sneridagh I think the correct thing here is to pin `importlib-metadata = 2.1.1`. `importlib-metadata >= 3.0.0` is not compatible with Python 2:

https://github.com/python/importlib_metadata/blob/530f5da8dd3a255f1198f29ea0126b8b25e644d8/setup.cfg#L14

wesleybl:
By the way @tisto, I think the master is not broken because of #1009. I think it's an intermittent error. If I have time I can take a look.

plone-5.1.x.cfg
sneridagh: thanks! I'll try, this build is killing me

sneridagh:
@wesleybl it would be great. We need to track down this heisenbug. 
/cc @jensens It would be great if we can also pinpoint what happened in master and why it's red.

jensens:
Well, a `__traceback_info__`` breaks nothing - except if we have some crazy doctest checking the some traceback outputs word by word.
That said, the PR tests been green before.

sneridagh:
@jensens yeah, I know... but we need to find out why failed on master. Any clue given the traceback?

tisto:
@wesleybl fixed master with his PR.

wesleybl:
In fact, it is an intermittent error. I will still investigate what caused the error. But at least the master is green now.

PR #1034, opened 2020-12-15
Add gh action

tisto:


tisto:
Screw you TravisCI. We won't wait for you any longer. I just added gh actions config anyway. Who cares...

wesleybl:
@tisto out of curiosity, do you know if ga has a cache like  Travis? Do you intend to revomer Travis completely?

tisto:
@wesleybl looking into this is actually on my todo list. Though, gh actions are so fast that I expect that some level of (maybe internal) caching already happens there.

djay:
@wesleybl 
something like this
```
      - name: Cache multiple paths
        uses: actions/cache@v2
        with:
          path: |
            ~/buildout-cache
            ~/eggs
          key: ${{ runner.os }}-${{ hashFiles('**/lockfiles') }}

```

wesleybl:
@tisto to completely remove Travis it is necessary to find an alternative to coverage.

Thank you @djay ! Then we can try that.

PR #1042, opened 2020-12-22
Navportlet as a separate service and expand

tiberiuichim:
I've created the `@navportlet` service endpoint and expand. I've ported the code (and tests) from plone.app.portlets to the navportlet.py file, so that this implementation doesn't depend on any portlet-related code.

I've kept the "portlet" name (although this is not a portlet) because `@navigation` is already taken and I didn't want to mix the functionalities (global navigation and "local" portlet navigation) in the same endpoint, it would have been confusing. There's a ton of edge cases and special functionality in the navigation portlet which makes mixing the two hard.


.flake8
```diff
@@ -0,0 +1,7 @@
+[flake8]
+# Recommend matching the black line length (default 88),
+# rather than using the flake8 default of 79:
+max-line-length = 88
```
tisto, line 4: @tiberiuichim don't we set this in buildout already? I usually set max-line-length to 200. Otherwise, it can interfere with black.

base.cfg
```diff
@@ -205,3 +205,4 @@ plone.schema = git git://github.com/plone/plone.schema.git pushurl=git@github.co
 
 [versions]
 plone.restapi =
+feedparser =
```
tisto, line 208: @tiberiuichim added by accident?

.flake8
tiberiuichim: @tisto It's for the integration with my vim setup, I've installed [flake8-black](https://pypi.org/project/flake8-black/) (I usually use autopep8). They recommend adding this file to the project. The problem with buildout settings is that it's not read by the editor setup.

base.cfg
tiberiuichim: No, it won't build otherwise. There's a pinned version in upstream that breaks this.

.flake8
tisto: I am fine with moving the buildout settings to a flake8 config file. Though, we should use the 200 line-length then.

base.cfg
tisto: Do we have a feedparser dependency in the nav portlet code?!

tiberiuichim: No, the dependency is in Plone.

https://github.com/plone/buildout.coredev/blob/20beb512d2931ddd43558fa1c3086b1e15d35eac/versions.cfg#L89

I no longer have access to my development computer, but looking at the feedparser readme, it [looks rough](https://pypi.org/project/feedparser/5.2.1/) with its python3 compatibility. I run ArchLinux and it always upgrades core packages to their latest released version (so, for example, I get to use python3.9 a lot sooner then other distros). So it might be a combination of those two factors.

tisto: @tiberiuichim ok, this might be the reason why the Python 2.7 builds are all failing. I actually had troubles with Python 3.8 and 3.9 on OS X so I am currently using 3.7, so I can't really check on my machine right now.

tisto:
@tiberiuichim can you rebase master to see if that fixes the CI failure?

jensens:
So,Portlets, really? I had the hope we can get rid of them with Volto. Can't we generalize this code to deal with parametrized navigation overall instead of a specific Portlet?

tisto:
@jensens Volto needs a navigation tree element to allow users to navigate deep content structures. We can call this portlet, slots, or whatever. If you have a better name we are open to suggestions. I personally do not like the portlet naming in that endpoint either but we can not overload the @navigation endpoint because it behaves differently. @navigation also allows you to retrieve a tree, so calling the new thing @navigationtree is also misleading. Naming...

@sneridagh is working on the slots idea and we have no plans to include portlets in Volto. 

This PR is a compromise to get the functionality we need into Volto (or at least into our customer project). Adding something to plone.restapi does not mean the functionality is added to Volto automatically. Volto is just one consumer of the REST API.

tiberiuichim:
@jensens @sneridagh @tisto Outside of a real convention for "slots endpoints", what would be a good name for this component? I can propose:

- `@navtree`
- `@contextnav`
- `@smartnav`
- `@localnav`

What about naming internally? Because I've ported code from plone.app.portlets, I've tried to keep the changes to a minimum, to allow some sort of "tracing back the code" and identify what was what.

jensens:
Thinking further and given that people with less Plone history in mind are users of the API I question why we need two things doing the same? We already have `@navigation` and you in your introduction you say you dont want to confuse with it, I would go for `@localnavigation` or `@contextnavigation`.

- abbreviations are confusing for new users, avoid them 
- the term `tree` is misleading here, since the global navigation is a tree too
- the term `smart` is never smart (same category as easy, simple, ...) - also we had this for Smart Folders in past (today the Collection)

jensens:
Thinking about it, I would go for `@contextnavigation`, because we do things "in context" in Plone.

news/1042.feature
```diff
@@ -0,0 +1,22 @@
+A new service endpoint and expand is available, a Navigation Portlet exposed as
+``@navportlet``. It uses the same semantics as the classic Plone navigation
+portlet, largely through reusing the same code. Instead of storing the
+"portlet" configuration in a portlet assignment storage, you can pass these as
+parameters to the service or expand component.
+
+You can provide these as parameters:
+
+- ``name`` - The title of the navigation tree.
+- ``root_path`` - Root node path, can be "frontend path", derived from router
```
jensens, line 10: What does `can be "frontend path", derived from router`  mean? This does not make any sense to me.

tiberiuichim: Meaning "window location". The "physical path" in Zope requires rooting the path in the ZODB root, so you'd have to know the name of the Plone website. In contrast, the path accepted here can consider its root as the Plone site root

jensens: ok :) well, docs need to be written anyway.

tiberiuichim:

tiberiuichim:

sneridagh (approved):

sneridagh:

tisto:

news/1042.feature
```diff
@@ -0,0 +1,22 @@
+A new service endpoint and expand is available, a Navigation Portlet exposed as
+``@navportlet``. It uses the same semantics as the classic Plone navigation
```
tisto, line 2: @tiberiuichim contextnavigation :)

news/1042.feature
```diff
@@ -0,0 +1,22 @@
...
+portlet, largely through reusing the same code. Instead of storing the
+"portlet" configuration in a portlet assignment storage, you can pass these as
+parameters to the service or expand component.
+
+You can provide these as parameters:
```
tisto, line 7: @tiberiuichim this all belongs to the docs, not into the changelog.

PR #1052, opened 2021-01-25
Unify nav_title and title in navs

sneridagh:


tisto (approved):
After thinking about this for a few days and discussing it with @sneridagh and @tiberiuichim I came to the conclusion that we should build this for the 80% use case and keep things simple with a single title attribute. There are always edge-cases you can come up with (that are still possible to cover via customizations) and the price we have to pay is having the nav_title fallback in many occasions in Volto or any other 3rd party apps. Let's keep things simple and learn from our past. :)

tisto:
@sneridagh you need to amend the tests :)

sneridagh:

PR #1053, opened 2021-01-26
Fix @id attribute to has the expansions query on them when no fullobj…

sneridagh:
…ects call is made.

This is the story: We always called `fullobjects` in Volto until Volto 10, when we stopped doing that. That hide this bug, where the `@id` is the canonical_url of the batch. This url has the query string on it, which breaks the assumption `@id` = `absolute_url` item (which we are doing in Volto).

Question is, why this was like this in the beginning? what `batch.canonical_url` gives in addition? I do not remember using it for pagination or anything, at least in Volto...

Might be a leftover from jsonLD or hydra specs?

Proof that this has poor meaning is that removing it, all tests still passed. I added a new one to keep track of it too.

### DIFF
--- a/src/plone/restapi/tests/http-examples/content_get_folder.resp
+++ b/src/plone/restapi/tests/http-examples/content_get_folder.resp
@@ -22,7 +22,7 @@ Content-Type: application/json
       "@id": "http://localhost:55001/plone/folder/@workflow"
     }
   }, 
-  "@id": "http://localhost:55001/plone/folder?metadata_fields=UID&metadata_fields=Creator", 
+  "@id": "http://localhost:55001/plone/folder", 
   "@type": "Folder", 
   "UID": "SomeUUID000000000000000000000002", 
   "allow_discussion": false, 
--- a/news/837.bugfix
+++ b/news/837.bugfix
@@ -1,0 +1,2 @@
+- Fix ``@id`` when content query has no ``fullbojects``
+  [sneridagh]
--- a/src/plone/restapi/serializer/atcontent.py
+++ b/src/plone/restapi/serializer/atcontent.py
@@ -111,8 +111,6 @@ class SerializeFolderToJson(SerializeToJson):
 
             batch = HypermediaBatch(self.request, brains)
 
-            if not self.request.form.get("fullobjects"):
-                result["@id"] = batch.canonical_url
             result["items_total"] = batch.items_total
             if batch.links:
                 result["batching"] = batch.links
--- a/src/plone/restapi/serializer/dxcontent.py
+++ b/src/plone/restapi/serializer/dxcontent.py
@@ -154,8 +154,6 @@ class SerializeFolderToJson(SerializeToJson):
 
             batch = HypermediaBatch(self.request, brains)
 
-            if "fullobjects" not in self.request.form:
-                result["@id"] = batch.canonical_url
             result["items_total"] = batch.items_total
             if batch.links:
                 result["batching"] = batch.links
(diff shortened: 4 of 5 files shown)


src/plone/restapi/tests/http-examples/content_get_folder.resp
```diff
@@ -22,7 +22,7 @@ Content-Type: application/json
       "@id": "http://localhost:55001/plone/folder/@workflow"
     }
   }, 
-  "@id": "http://localhost:55001/plone/folder?metadata_fields=UID&metadata_fields=Creator",
```
sneridagh, line 25: @plone/volto-team I can see that `@id` was initially thought for keeping the request how this response was created... and maintains the "browserability" promised by (is it hydra or jsonld?)
However, the fact that we've been living without it all this time... makes me wonder if maintaining this REST "purity" is no longer relevant.

sneridagh:

src/plone/restapi/tests/http-examples/content_get_folder.resp
tiberiuichim: I think the intention of having the `@id` reflect the uniqueness of the conditions that generated that response is very good. But I think this intention is not reflected in the whole stack, because I argue: if the `@id` is important here and it makes things "browserable", then the `@id` everywhere should reflect this. Let me point to this bit of code, which is the default DX content serializer:
```
        result = {
            # '@context': 'http://www.w3.org/ns/hydra/context.jsonld',
            "@id": obj.absolute_url(),
            "id": obj.id,
            "@type": obj.portal_type,
            "parent": parent_summary,
            "created": json_compatible(obj.created()),
            "modified": json_compatible(obj.modified()),
            "review_state": self._get_workflow_state(obj),
            "UID": obj.UID(),
            "version": version,
            "layout": self.context.getLayout(),
            "is_folderish": False,
        }

        # Insert next/prev information
        nextprevious = NextPrevious(obj)
        result.update(
            {"previous_item": nextprevious.previous, "next_item": nextprevious.next}
        )

        # Insert expandable elements
        result.update(expandable_elements(self.context, self.request))
```

Notice that the `result['@id']` is the object url, but that results depends on the expands. Without the expand you can't reproduce the resource identified by that `@id`.

tisto: The @id is the node identifier from the JSON-LD (https://www.w3.org/TR/json-ld/#node-identifiers) standard and should reference the resource (without attributes as far as I understand it). We could use the absolute_url attr or come up with a new one instead. I don't see a value in just dropping it and replacing it with something non-standard. Just because it is not used everywhere, does not mean it has no value IMHO. It uniquely identifies the endpoint and indeed allows the human user of the API to browse and discover the API.

sneridagh:

tisto (approved):

src/plone/restapi/tests/http-examples/content_get_folder.resp
tiberiuichim: @tisto the `@id` wouldn't be dropped, it's already inserted in the standard serializer. 

My point was that, if the `@id` here is important and should uniquely identify the Hyperbatch, then the same logic should apply to all nodes that represent themselves with that `@id`, which I don't think is true, though, as we have ids attached to resources with varying amount of information, depending on how that resource is fetched (see my expand example in the previous comment). Basically, we have to accept that we're always dealing with partial graphs when fetching the endpoints and the resources with those ids `@id` are incomplete.

tisto: @tiberiuichim I discussed this with @sneridagh 1:1 in a call and figured that out :)

PR #1063, opened 2021-02-09
Slots

tiberiuichim:


src/plone/restapi/configure.zcml
```diff
@@ -1,101 +1,118 @@
...
     xmlns:zcml="http://namespaces.zope.org/zcml"
-    i18n_domain="plone.restapi">
+    i18n_domain="plone.restapi"
+    >
 
   <i18n:registerTranslations directory="locales" />
 
-  <include package="plone.behavior" file="meta.zcml"/>
-  <include package="plone.rest" file="meta.zcml"/>
+  <include
```
tisto, line 13: @tiberiuichim do you have a special formatter in your IDE? We shouldn't mix code cleanups and complex PRs together. This makes it a lot harder to do a proper review.

tisto:
@tiberiuichim @sneridagh we should follow our documentation first approach for this and first work on the documentation and then move on with the implementation. Can we please create a new PR that contains the docs about how this is supposed to work first?

src/plone/restapi/configure.zcml
tiberiuichim: @tisto I use [collective zpretty](https://github.com/collective/zpretty), which has a special mode for zcml. I guess we could integrate zpretty in a separate PR and run it on all zcml files?

```
zpretty [-h] [--encoding ENCODING] [-i] [-x] [-z] [file ...]

positional arguments:
  file                 The list of files to prettify (defaults to stdin)

optional arguments:
  -h, --help           show this help message and exit
  --encoding ENCODING  The file encoding (defaults to utf8)
  -i, --inplace        Format files in place (overwrite existing file)
  -x, --xml            Threat the input file(s) as XML
  -z, --zcml           Threat the input file(s) as XML. Follow the ZCML
                       styleguide
```

tisto: @tiberiuichim yeah. please create a separate PR for this. Is zpretty adopted in Plone core @ale-rt? Can it run as a pre-commit hook as part of p.r.codeanalysis and in a CI mode to check for violation? If we adopt a tool we should either go all in or not use it at all IMHO. Adopting it a bit won't really help much.

ale-rt: Sorry @tisto I missed the comment before.

zpretty is not adopted in Plone core.

Is was thinking about plugging it with https://pre-commit.com/ which is what I use recently.

Note that I recently worked quite a lot on the zcml part, especially in https://github.com/collective/zpretty/pull/32.
I will probably do a release nowish.

ale-rt: If you are interested zpretty 2.1.0 is out.
The nice thing about that is IMO That the attribute order is in sync with the function signature, see this example for the `CORSPolicy` directive
https://github.com/collective/zpretty/blob/e1ef596cfd1edea4113423ce5ef4467c0506c650/zpretty/zcml.py#L47-L57

tiberiuichim:

tiberiuichim:

tiberiuichim:

jensens:
"slots" is a confusing name given its a Python core concept  https://docs.python.org/3/reference/datamodel.html#object.__slots__

PR #1066, opened 2021-02-11
Respect access inactive content permission in @search

cekk:
In [Products.CMFPlone](https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/CatalogTool.py#L423) there is a method that checks if the current user has that permission in the list of paths in kw.

In @search handler we call _searchResults_ method passing a query and not kw.
This leads to a problem because if you give that permission to some roles (Editor for example) and you give Editor role to an user in a subfolder, the @search endpoint will not show expired contents in that folder even if the user has the right permission.

You can test it easily with Volto because the "/contents" view does exactly the problematic @search call:
- add Editor to "Access inactive portal content" permission
- assign Editor role to an user inside a subfolder
- create an expired content inside it
- go to /contents view
- expired content is missing
- in the backend you can see it correctly in folder_contents

with this pr i am going to change how we pass parameters to searchResults

I'm doing the fix here because it's easier and quicker but if you think that this should be fixed inside CMFPlone, let me know.

### DIFF
--- a/news/1066.bugfix
+++ b/news/1066.bugfix
@@ -1,0 +1,1 @@
+Respect "Access inactive portal content" permission in @search endpoint [cekk]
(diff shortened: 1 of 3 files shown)


cekk:

tisto (approved):
LGTM. We worked around lots of quirks in Plone core in plone.restapi in the past. Though, at some point, we have to start fixing things...

PR #1067, opened 2021-02-12
handle quoted urls in copymove endpoint

cekk:
This pr solves the following problem:

i have a folder with a space in the id (/aaa/b b/)..maybe from an old migration.

With Volto i can't copy/paste items inside it because Volto calls the endpoint with the quoted path (/aaa/b%20b) and restapi don't find the right object to move.

### DIFF
--- a/news/bugfix.1067
+++ b/news/bugfix.1067
@@ -1,0 +1,1 @@
+Handle quoted urls into copymove endpoint [cekk]
--- a/src/plone/restapi/services/copymove/copymove.py
+++ b/src/plone/restapi/services/copymove/copymove.py
@@ -11,6 +11,11 @@ from zope.security import checkPermission
 import plone
 import six
 
+if six.PY2:
+    from urllib2 import unquote
+else:
+    from urllib.parse import unquote
+
 
 class BaseCopyMove(Service):
     """Base service for copy/move operations."""
@@ -25,6 +30,7 @@ class BaseCopyMove(Service):
 
     def get_object(self, key):
         """Get an object by url, path or UID."""
+        key = unquote(key)
         if key.startswith(self.portal_url):
             # Resolve by URL
             key = key[len(self.portal_url) + 1 :]
--- a/src/plone/restapi/tests/test_copymove.py
+++ b/src/plone/restapi/tests/test_copymove.py
@@ -73,7 +73,7 @@ class TestCopyMove(unittest.TestCase):
         Volto returns paths like this
         """
         service = self.traverse("/plone/@copy", method="POST")
-        obj = service.get_object('/doc2-with%20spaces')
+        obj = service.get_object("/doc2-with%20spaces")
         self.assertEqual(self.doc2, obj)
 
 
--- a/src/plone/restapi/tests/test_copymove.py
+++ b/src/plone/restapi/tests/test_copymove.py
@@ -27,7 +27,9 @@ class TestCopyMove(unittest.TestCase):
             self.portal.invokeFactory("Document", id="doc1", title="My Document")
         ]
         self.doc2 = self.portal[
-            self.portal.invokeFactory("Document", id="doc2-with spaces", title="My Document")
+            self.portal.invokeFactory(
+                "Document", id="doc2-with spaces", title="My Document"
+            )
         ]
         self.folder1 = self.portal[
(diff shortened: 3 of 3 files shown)


tisto:
@cekk we had lots of blanks in IDs in a large migration that we did recently. I can tell you that this is something you want to fix because it leads to tons of problems (e.g. ordering does not work in the folder where the object lives).

I am ok with making plone.restapi more resilient to those kinds of errors. Though, you have to get rid of those objects or create PRs for a lot more things in Plone.

cekk:
Damn..stupid old Plone 😁

Ok, so you fixed all the ids in your database?

tisto:
@cekk yeah. we had to put quite some effort into fixing those. Though, the good thing is that p.a.redirector does the trick and the old URLs still work.

We are happy to share our fixing script if you are interested. @robdayz wrote it...

cekk:
yes, could be great if you'll share your work. I was thinking about to create a post-migration script for our [tool](https://github.com/RedTurtle/redturtle.importer.base) ;)

robdayz:
@cekk Sorry, I forget to add the script.
```

"""
Script to normalize ids.
Duplicated item id in the same folder is not replaced, only logged.
"""

from plone import api
from zope.component import getUtility
from plone.i18n.normalizer.interfaces import IIDNormalizer
from Acquisition import aq_base
from Acquisition import aq_inner
from Acquisition import aq_parent
from zope.container.interfaces import INameChooser
from ZODB.POSException import POSKeyError
from time import sleep

import transaction
import sys, os
import time
import datetime

sys.path.append(os.path.abspath(os.path.join("..", "api")))
from scripts import utils

APPLY_CHANGES = False
NEW_ID = 0
DUPLICATED = 0
FILENAME = "normalize_ids"
FILE = datetime.datetime.now().strftime(FILENAME + "_%Y%m%d-%H%M.csv")

# CSV header
csv_file = [[" #PAGE", "OLD_ID", "NEW_ID", "INFO"]]

catalog = api.portal.get_tool("portal_catalog")
# Sort results before start renaming ids.
results = catalog.searchResults(sort_on="path")

utils.print_info("Found {} objects.".format(len(results)))

normalizer = getUtility(IIDNormalizer)
items = []
total = 0

for obj_res in results:
    obj = obj_res.getObject()
    item = "/".join(obj.getPhysicalPath())
    # Check if blob is present to bypass some problems before renaming.
    # This is not needed in production only in local (without blobs).
    try:
        elem = catalog(path={"query": item, "depth": 0})
        blob_ok = True
    except POSKeyError:
        elem = None
        blob_ok = False
        utils.print_error("Missing blob: {}".format(item))

    if elem:
        obj = elem[0].getObject()
        oldid = obj.id
        newid = normalizer.normalize(oldid, max_length=256)
        # We check if after normalizing this id we obtain same or not.
        # If it is different, then we need to apply some renaming tasks here.
        CONTINUE = True
        if oldid != newid:
            NEW_ID += 1
            parent = aq_parent(obj)
            ids = []
            # get folder items of current id to check if they are duplicated
            for k, v in parent.items():
                ids.append(k)
            if newid in ids:
                # When ID is used, I need to generate a unique one.
                # But we only log it, don't do any replacement.
                CONTINUE = False
                chooser = INameChooser(parent)
                newid = chooser.chooseName(newid, parent)
                # Log the duplicated id
                utils.print_error("{} / {} / ({})".format(oldid, newid, item))
                DUPLICATED += 1
                csv_file.append([item, oldid, newid, "DUPLICATED_ID"])

            # Only continues when id is not in folder.items()
            # And it doesn't generates -1 id
            if CONTINUE and APPLY_CHANGES:
                try:
                    blob_ok = True
                    parent.manage_renameObject(oldid, newid)
                except POSKeyError:
                    blob_ok = False
                    utils.print_error("Missing blob: {}".format(item))

                if blob_ok:
                    # Save oldid in custom obj fields
                    obj._original_id = oldid
                    obj._new_id = newid
                    # reindex new value
                    obj.reindexObject(idxs=["id"])
                    csv_file.append([item, oldid, newid, "RENAMED"])
                    transaction.commit()

utils.save_csv(csv_file, FILE)
utils.print_ok("{} IDs should be normalized".format(NEW_ID))
utils.print_ok("Found {} IDs that are duplicated and should use -1 in id".format(DUPLICATED))
utils.print_ok("Check results in '{}'".format(FILE))
```

PR #1069, opened 2021-02-15
Mark restapi 7 with a zcml feature flag

sneridagh:
I want to be able to:

```
<configure zcml:condition="has plonerestapi7">
...
```
Added a couple of comprehensive tests for feature reassurance too.

### DIFF
--- a/news/1068.feature
+++ b/news/1068.feature
@@ -1,4 +1,2 @@
-- Mark restapi 7 with a zcml feature flag
-  [sneridagh]
-- Add a couple of additional tests for resolveuid feature reassurance
-  [sneridagh]
+Mark restapi 7 with a zcml feature flag: ``plonerestapi-7``
+[sneridagh]
--- a/news/1068.feature
+++ b/news/1068.feature
@@ -1,0 +1,4 @@
+- Mark restapi 7 with a zcml feature flag
+  [sneridagh]
+- Add a couple of additional tests for resolveuid feature reassurance
+  [sneridagh]
--- a/src/plone/restapi/configure.zcml
+++ b/src/plone/restapi/configure.zcml
@@ -10,7 +10,7 @@
 
   <i18n:registerTranslations directory="locales" />
 
-  <meta:provides feature="plonerestapi7" />
+  <meta:provides feature="plonerestapi-7" />
 
   <include package="plone.behavior" file="meta.zcml"/>
   <include package="plone.rest" file="meta.zcml"/>
--- a/src/plone/restapi/configure.zcml
+++ b/src/plone/restapi/configure.zcml
@@ -5,9 +5,12 @@
     xmlns:plone="http://namespaces.plone.org/plone"
     xmlns:genericsetup="http://namespaces.zope.org/genericsetup"
     xmlns:zcml="http://namespaces.zope.org/zcml"
+    xmlns:meta="http://namespaces.zope.org/meta"
     i18n_domain="plone.restapi">
 
   <i18n:registerTranslations directory="locales" />
+
+  <meta:provides feature="plonerestapi7" />
 
   <include package="plone.behavior" file="meta.zcml"/>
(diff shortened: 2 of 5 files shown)


jensens:
Given what we have in 
https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/meta.zcml
I would name it `plonerestapi-7` (with dash).
Otherwise LGTM.

sneridagh:

news/1068.feature
```diff
@@ -0,0 +1,4 @@
+- Mark restapi 7 with a zcml feature flag
```
tisto, line 1: @sneridagh as far as I know towncrier does not allow multiple entries in one file. The "-" also need to be removed since towncrier always adds them.

src/plone/restapi/configure.zcml
```diff
@@ -5,10 +5,13 @@
     xmlns:plone="http://namespaces.plone.org/plone"
     xmlns:genericsetup="http://namespaces.zope.org/genericsetup"
     xmlns:zcml="http://namespaces.zope.org/zcml"
+    xmlns:meta="http://namespaces.zope.org/meta"
     i18n_domain="plone.restapi">
 
   <i18n:registerTranslations directory="locales" />
 
+  <meta:provides feature="plonerestapi-7" />
```
tisto, line 13: @sneridagh I am +1 on this. Though, we have to be super careful that we change this when we release p.restapi 8.

news/1068.feature
mauritsvanrees: This should indeed be split into multiple files: `1068.feature.1`, `1068.feature.2`, etc.

Or keep them in one file like this, if you don't mind them ending up in the same line:

```
Mark restapi 7 with a zcml feature flag: ``plonerestapi-7``.
Add a couple of additional tests for resolveuid feature reassurance
[sneridagh]
```

(Or remove the line about additional tests entirely.)

Note that in the example I added the name of the zcml feature `plonerestapi-7` explicitly, so people don't need to look in the code to see what it is called. :-)

src/plone/restapi/configure.zcml
mauritsvanrees: To be sure: we should not _change_ this when we release plone.restapi 8. We should keep it and **add** a second feature`plonerestapi-8`. `CMFPlone` does the same, and still has features `plone-4`, `plone-52`, etc.
But maybe you mean that.

news/1068.feature
sneridagh: @mauritsvanrees @tisto done!

sneridagh:

mauritsvanrees (changes requested):
Wait a minute. This should be in `meta.zcml`.
From what I saw while working on `plone.autoinclude` last week, it *might* work in `configure.zcml`, but not always. I tried just now in its tests: the tests that use `pip install` can find features from `configure.zcml` just fine, but the buildout tests and the coverage tests don't.

This could also be a problem in the `plone.autoinclude` tests, but I have a better reason for putting it in `meta.zcml`: only then is this feature loaded before any zcml that uses it is loaded. In other words: when package `aaa` checks the feature, it is not there, but when package `zzz` checks it, is *is* there (assuming for simplicity that packages are loaded alphabetically).

sneridagh:
@mauritsvanrees interesting indeed! I will make the change.

sneridagh:
@mauritsvanrees does it needs to be called from configure.zcml?

mauritsvanrees:
> does it needs to be called from configure.zcml?

No, the current autoinclude entry point in `setup.py` is enough. Then `CMFPlone/meta.zcml` [includes it](https://github.com/plone/Products.CMFPlone/blob/5.2.3/Products/CMFPlone/meta.zcml#L32-L42).

mauritsvanrees (approved):

sneridagh:

Issue #1070, opened 2021-02-15
IBlockFieldSerializationTransformer and IBlockFieldDeserializationTransformer broken for DX CType with Layout enabled


avoinea:
Right now all **Field Serialization Transformers** and **Field Deserialization Transformers** are registered on `IBlocks` behavior interface. 

But in order for TTW DX Content-types Layout to work we had to disable this behavior and manually add `blocks` and `blocks_layout` fields.

Thus, we need a new marker interface behavior for custom blocks.

davisagli:
Fixed in https://github.com/plone/plone.restapi/pull/1477

PR #1071, opened 2021-02-15
Add volto.blocks.custom (ICustomBlocks) behavior

avoinea:
See #1070


avoinea:
I'll try to go to the root of this issue https://github.com/plone/plone.dexterity/issues/144

Issue #1073, opened 2021-02-18
Avoid duplicate fields from multiple behaviors, TTW schema


avoinea:
See https://github.com/plone/plone.dexterity/issues/144

mauritsvanrees:
Closing because this was apparently fixed already two years ago in PR #1074.

The related change in `plone.dexterity` was never completed, and I closed all related issues and PRs today.  See https://github.com/plone/plone.dexterity/issues/144#issuecomment-1444632183.

I don't know if this mismatch is in any way a problem, but I guess not, since the situation has not changed the past two years.

PR #1074, opened 2021-02-18
Avoid duplicate fields from multiple behaviors

avoinea:
See #1073


avoinea:

tisto (approved):
Good catch!

Issue #817, opened 2019-10-26
 @actions does not include the action URLs


zopyx:
I wonder why @actions does not include the URLs of actions?
When I need to render actions then I need their URLs. 
Otherwise the @actions endpoint has no value. Or?

https://plonerestapi.readthedocs.io/en/latest/actions.html 2

tisto:
Just for the record my comment on PR #1076 (maybe it is better to continue the discussion here):

@libargutxi thank you for your contribution! This issue might be more complex than it looks at first sight because REST API design principles need to be taken into account.

Plone REST API uses the concept of resources that are linked via the "@id" attribute, this is the idea of a hypermedia API with linked resources. We currently do not have a similar concept for "actions". There are concepts in the JSON-LD standard but they are very academic and we never implemented them.

If we just expose the Plone Classic URLs, like in this PR, we only serve the Plone Classic use case and ignore Volto and other REST API consumers. The question is: how is Volto or another consumer suppose to handle this URL field in a generic way. If we start using consumer-specific fields, like this URL field, we risk ending up with an overly complex API and duplicate fields (e.g. how should we call the Volto URL actions field, how the Vue actions field, etc.). I think you get the point.

I don't have the answers to the questions that I raise. I just would like to point out the underlying complexity here and that there is no easy and straight-forward solution IMHO.

erral:
correct my if I am wrong but from my limited Volto experience I see that Volto already handles REST API generated URLs, for instance when showing the breadcrumb, here: https://github.com/plone/volto/blob/master/src/reducers/breadcrumbs/breadcrumbs.js#L40

In this case Volto should handle that too (we are preparing a PR also for the actions reducer on Volto to handle these changes in @actions endpoint).

The point is that if we are exposing the actions, we need to somehow expose the URL of the action, otherwise the endpoint is useless in our point of view.

sneridagh:
@erral @ionlizarazu my 5cents,
@actions is used in Volto like this in the toolbar:
https://github.com/plone/volto/blob/master/src/components/manage/Toolbar/Toolbar.jsx#L279

then we infer them from the action, using Volto's own URLs. That allows also the frontends not to be tied to old naming like "contact_info", etc.

We could have set the Volto URLs in the @actions endpoint, then all the other frontends that want to use this would have been out, so we decided not to bind that to any (including classic Plone).

@erral the breadcrumbs reducer is "normalising" the backend (of content resources, not CMSUI) URL to the Volto ones.

[...]

PR #1076, opened 2021-02-24
add url field to Actions

ionlizarazu:
fixes this issue:
https://github.com/plone/plone.restapi/issues/817


ionlizarazu:

tisto:
@libargutxi thank you for your contribution! This issue might be more complex than it looks at first sight because REST API design principles need to be taken into account. 

Plone REST API uses the concept of resources that are linked via the "@id" attribute, this is the idea of a hypermedia API with linked resources. We currently do not have a similar concept for "actions". There are concepts in the JSON-LD standard but they are very academic and we never implemented them.

If we just expose the Plone Classic URLs, like in this PR, we only serve the Plone Classic use case and ignore Volto and other REST API consumers. The question is: how is Volto or another consumer suppose to handle this URL field in a generic way. If we start using consumer-specific fields, like this URL field, we risk ending up with an overly complex API and duplicate fields (e.g. how should we call the Volto URL actions field, how the Vue actions field, etc.). I think you get the point.

I don't have the answers to the questions that I raise. I just would like to point out the underlying complexity here and that there is no easy and straight-forward solution IMHO.

nileshgulia1:
@ionlizarazu Can you please bring this PR up-to-date?

erral:
Updated

src/plone/restapi/tests/http-examples/expansion_expanded_full.resp
```diff
@@ -9,100 +9,118 @@ Content-Type: application/json
                 {
                     "icon": "",
                     "id": "view",
-                    "title": "View"
+                    "title": "View",
+                    "url": "http://localhost:55001/plone/front-page"
```
jensens, line 13: please add a trailing `,` in order to avoid a two-line diff next time (same as `black` would do).

jensens: just found, `black` check is failing anyway.

erral: added running black. That shoukd be catched by the black checker indeed :/

avoinea (approved):

erral:

sneridagh:
I think we need the plone.volto counterpart (GS config to make sure that no classic URLs are leaked there).

nileshgulia1:

avoinea:
@tisto Please have a look.

nileshgulia1:

avoinea:
@jensens @tisto What's the status here? Can we move on with this one, please?

jensens (approved):
I think we are fine if we merge this one and place an issue in plone.volto

nileshgulia1:
> I think we are fine if we merge this one and place an issue in plone.volto

@jensens  We also have a PR in plone.volto https://github.com/plone/plone.volto/pull/64

Issue #1077, opened 2021-02-28
Support fields/widgets with concrete dynamic vocabularies


alecpm:
Right now fields that have `vocabulary` set to a vocabulary name and vocabularies which are essentially static are well supported by plone.restapi. Vocabularies where the values depend on the context or on a query (e.g. vocabularies providing `z3c.formwidget.query.interfaces.IQuerySource`) are not really supported. This relates to #535, and I will add a PR with basic support for vocabularies using `IQuerySource` shortly.

jensens:
Closing as resolved. IQuerySource support implemented via @querysources endpoint.
Thank you for the original report!

davisagli:
~~I wasn't aware of that endpoint, and it seems to have no tests or documentation, so I would not really call this done.~~

davisagli:
Ah no, I just had a bad setting when I was searching. Still, I'd like to review the status before closing.

PR #1078, opened 2021-03-01
Better support for IQuerySource vocabularies

alecpm:
This PR implements a couple small changes to help facilitate use of Async/AJAX select widgets with fields using a vocabulary providing `IQuerySource`. It makes the following changes.

1) Look for such vocabularies in both the `vocabulary` property of the field as well as the `source`.
2) Handle widgets that set their `vocabulary` attribute to a value other than a vocabulary name string.

This relates to issue #1077 and indirectly to #535.

CC: @pbauer @skleinfeldt

### DIFF
--- a/src/plone/restapi/services/querysources/get.py
+++ b/src/plone/restapi/services/querysources/get.py
@@ -24,6 +24,11 @@ class QuerySourcesGet(SourcesGet):
         bound_field = field.bind(self.context)
 
         source = bound_field.source
+        if not source and IQuerySource.providedBy(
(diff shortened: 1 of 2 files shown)


tisto:
@alecpm LGTM. Thanks for the fix! Would it be possible to write a test for this or is this too much hassle? We have a testing profile where we could set up vocabularies I guess...

tisto (approved):

tisto:
Any takers to write a few tests? Then we could merge this PR.

PR #1081, opened 2021-03-11
Optional search endpoint follows site settings

tiberiuichim:
The `@search` endpoint can now follow Plone's settings if a `use_site_search_settings` parameter is passed in the request.

This means that the types will be filtered, the navigation root will be respected and the default sort order will be used.


tiberiuichim:

sneridagh (approved):

sneridagh:
The search view use case in Plone is a different beast that should be addressed like this in the backend while preserving the independence of the `search` endpoint as main interface for querying the Plone DB.

This is the Volto counterpart:
https://github.com/plone/volto/pull/2264

Thanks @tiberiuichim for taking care.

tiberiuichim:

tisto (approved):

Issue #1084, opened 2021-03-11
Blocks stored inside other blocks


tiberiuichim:
## Motivation

We're now embedding blocks in the value of other blocks. I can enumerate several community addons that do this:

- volto-columns-block
- volto-grid-block
- volto-accordion-block (multiple implementations)
- collective/volto-subblocks does it
- volto-group-block

The problem with all these implementations is that their inner blocks don't have transformer support.

## Solution

We can provide sub-blocks support by inspecting the block value for "subblocks". The only problem is: "how are the subblocks stored"? 

All EEA subblocks implementation rely on one convention: subblocks are stored in the block value either as `blocks` or `data.blocks`. The implementation I'm writing is using this convention. Other implementations can easily be done as a separate transformer, see this gist: https://gist.github.com/tiberiuichim/aa21bb3952e2011ba538212d56dc531c

PR #1085, opened 2021-03-11
Add subblocks transformer support

tiberiuichim:
Fixes https://github.com/plone/plone.restapi/issues/1084

### DIFF
--- a/docs/source/blocks.rst
+++ b/docs/source/blocks.rst
@@ -149,16 +149,20 @@ Then register as a subscription adapter::
 Generic block transformers and smart fields
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-You can create a block transformer that applies to all blocks, by using `None`
-as the value for `block_type`. The `order` field still applies, though. Using
+You can create a block transformer that applies to all blocks, by using ``None``
+as the value for ``block_type``. The ``order`` field still applies, though. Using
 the generic block transfomers enables us to create **smart block fields**,
-which are handled differently. For example, any internal link stored as `url`
-or `href` in a block value is converted (and stored) as a resolveuid-based URL,
+which are handled differently. For example, any internal link stored as ``url``
+or ``href`` in a block value is converted (and stored) as a resolveuid-based URL,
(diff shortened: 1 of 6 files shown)


cekk (approved):

cekk:
LGTM
Just to know: is this will be the standard of managing subblocks? Because we are using them in a different way (\cc @nzambello @giuliaghisini)

tiberiuichim:
@cekk I'm aware that volto-subblocks uses another storage scheme, but I wasn't 100% sure how that looks like, and I made a deal with Nicola that I'll push this implementation and you can complement how the subblocks are found. My proposal is that you push changes to this PR with our storage specifics, should be easy enough. You'll have to add your implementation here: https://github.com/plone/plone.restapi/pull/1085/files#diff-6ad4a25ea7136e36c0d249c9361fff5494a7bf28825b8e7702cc35f189bcef43R79

tiberiuichim:
@cekk I've also checked with @sneridagh and they don't have subblocks, so there's just these 2 cases to be handled.

tiberiuichim:
@cekk: See also related issue, where I'm detailing how to write a transformer-based subblocks implementation: https://github.com/plone/plone.restapi/issues/1084

tiberiuichim:

tiberiuichim:

tisto:
@cekk @tiberiuichim we should talk about our grid/sub-blocks implementations and try to unify our approaches. As far as I know @tiberiuichim and @sneridagh already kind of did this.

cekk:
I like @tiberiuichim implementation and that we should unify our approaches (i ping @nzambello  and @giuliaghisini for that).

I already made a super generic transformer [here](https://github.com/RedTurtle/redturtle.volto/blob/master/src/redturtle/volto/restapi/serializer/blocks.py#L21) because we have a different pattern: we store subblocks that are not real blocks but only a list of "a bunch of data with the same shape".

So: for me it's a big 👍  this approach basically because we can also use proper transformers and not a recursion hell like in our implementation

sneridagh (approved):

tisto (approved):

PR #1086, opened 2021-03-12
Try to fix buildout for all python versions

tiberiuichim:


avoinea:
We should upgrade to 5.2.4

tiberiuichim:

news/1086.bugfix
```diff
@@ -0,0 +1 @@
+Pin some package versions to fix buildout @tiberiuichim
```
tisto, line 1: @mauritsvanrees is it possible that we start using an "internal" categorie for towncrier? It does not make any sense at all IMHO to add such changelog entries to the changelog.

tisto (approved):
+1 to upgrade to 5.2.4.

news/1086.bugfix
mauritsvanrees: @tisto Two options I think:

1. Simply do not upload a news snippet. One GitHub check will fail. In that case I usually add a comment like "I did not add a news snippet, because I did not think it was necessary."
(Perhaps we could teach mr.roboto to react to a comment like "please pass news check" and let the news snippet check succeed then, but that seems overkill.)
It *is* possible to merge a PR even with a failing check, right? Needs admin permission perhaps, but still possible as far as I know.

2. Add a category in [`pyproject.toml`](https://github.com/plone/plone.restapi/blob/master/pyproject.toml), something like:

```
[[tool.towncrier.type]]
directory = "internal"
name = "Internal details:"
showcontent = true
```

Then it will show under an 'Internal details' header.

I don't know of a way to tell towncrier to ignore a snippet
I tried `show content = false`, but that does not do what we want.

You can use an extension `.ignore` or `.unknown` and towncrier will ignore it. I think `zestreleaser.towncrier` complains and proposes to quit to fix it, but you should be able to continue. But the unknown snippet will remain in the news directory after release.

PR #1094, opened 2021-03-22
Fix mr.developer checkout

tisto:


jensens:
Why? `develop = .` is usually sufficient.

tisto:
@jensens only if you use "plone.restapi =" in the buildout config that you call on the command line (doesn't work if you extend it). Buildout just sucks...

PR #1095, opened 2021-03-22
Fixes build was using the released version

sneridagh:


sneridagh:
@tisto once realised that mr.developer was not who to blame, I found out that the build was getting the wrong `versions`. 
I used the same approach than `collective.solr`. We already found some buildouts that was neglecting the `versions` if it was not set in the main .cfg.

sneridagh:
I would favor this one instead of #1094

tisto:
@sneridagh the thing I don't understand here is why we do not have to fix the documentation that I needed to fix in #1094 to make CI pass...

tisto (approved):
LGTM.

PR #1100, opened 2021-03-24
Skip using ci.cfg. Because you know buildout sucks...

tisto:


gotcha:
@tisto I am interested to know how buildout sucked in this case. Does it deserve an issue at https://github.com/buildout/buildout/ ?

tisto:
@gotcha sorry for the inappropriate changelog entry and thanks for asking! We have multiple issues at the same time I guess and tbh it is hard to pinpoint them and figure them all out without putting a lot of effort into trying to reproduce them in isolation. 

The buildout of plone.restapi is complex because it must serve many use cases. Therefore we have multiple buildout cfgs that depend on each other. It seems that version pins are not taken into account when they are not on the top level cfg. E.g. we pin versions in base.cfg or versions.cfg but they are ignored when we use them in plone-5.2.x.cfg or ci.cfg that extend those base cfgs.

I am trying to reduce complexity right now (e.g. with this PR) to being able to figure out what goes on any why.

Another issue, that is most likely related, is that our buildout uses the released version of plone.restapi instead of the one that is declared with "develop = .". You have to set "plone.restapi =" in all buildout cfgs to make sure this is applied.

I spent almost the full day yesterday on this and got a bit frustrated tbh. Will give it another shot today and report back...

gotcha:
@tisto when fighting to understand versions overrides with complex buildouts, this can help:

    buildout -v annotate versions

PR #1105, opened 2021-03-26
Fix search in VHM with PhysicalRoot

tiberiuichim:


tiberiuichim:

tisto (approved):
LGTM. Though, I 'd like to see a second review since I am not too familiar with VHM tbh.

tiberiuichim:
@tisto: There's already VHM-aware code in the search handler: https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/search/handler.py#L61

The problem is like this: the @search endpoint, in the code branch that I've added, that will "apply site search settings", will set the path to be the INavigationRoot. The only problem is that the existing search code will then tweak the path as it's expecting that the path is coming "from the frontend", so it's not something "physical path" that we use in Zope. My code basically translates the "zope path" to "frontend path" before it's then processed again, so that it ends up correct.

sneridagh (approved):
LGTM, I'm curious about why it was working before?
I remember using query with paths in some projects, and it worked well...

sneridagh:
ok, got it, it's about the navigation roots.

tiberiuichim:
there's two code paths, for each case:
- with use_site_search_settings
- withouth (plain old search behavior)

When I've introduced use_site_search_settings, I didn't get to test it in production, and when I've deployed it I found out that it doesn't work properly

So the PR fixes the use_site_search_settings code path, it doesn't touch the old use case

tiberiuichim:

PR #1120, opened 2021-05-08
Add collective.zpretty integration

tiberiuichim:


tiberiuichim:
@tisto what are the targets for this PR? We now have Github workflow that tests code for zpretty-based diffs.

Should I add a precommit hook that automatically formats zcml files?

tisto:
@tiberiuichim a pre-commit hook would be nice to have. I am still somehow experimenting with the CI pipeline on gh actions and I recently started to have doubts about the future of p.r.codeanalysis. Even though I wrote that recipe it does not really fit the gh actions style that we currently start to follow (multiple small jobs). Therefore we could add a zpretty pre-commit hook to the recipe but I am not sure if that's a long term perspective. 

How long does zpretty take on the p.restapi code? It is important that pre-commit hooks run fast...

sneridagh (approved):

tisto (changes requested):
@tiberiuichim the zpretty ci job fails with a diff.

tiberiuichim:
@tisto yes, it was intentional, to showcase how the job would fail. So, good to go? I'll commit the fix.

tisto:
@tiberiuichim cool! A failing CI is important before fixing things! Yeah, go ahead with the fix!

tisto:
@tiberiuichim plone.restapi 8 is released. We could go ahead with this PR when the CI is green. Feel free to rebase with master to drop the py2 builds. Maybe it would make sense to backport this to restapi 7 as well (in case we need to backport other PRs/fixes in the future).

news/1120.feature
```diff
@@ -0,0 +1 @@
+Format zcml files with collective.zpretty. Add zpretty Github workflow. @tiberiuichim
```
tisto, line 1: @tiberiuichim if I recall correctly there is an "internal" category that we can use for this.

tiberiuichim:

requirements.txt
```diff
@@ -2,3 +2,4 @@
 setuptools==42.0.2
 zc.buildout==2.13.3
 wheel
+zpretty
```
tisto, line 5: @tiberiuichim it should be sufficient to add the zpretty dependency in buildout only. The other three dependencies are necessary for the bootstrap.

tisto (approved):

PR #1122, opened 2021-05-12
Drop support for Python 2 and Plone 5.1 and 4.3.

tisto:
https://packaging.python.org/guides/dropping-older-python-versions/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,27 +6,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.8, 3.7, 3.6, 2.7]
-        plone-version: [5.2, 5.1, 4.3]
-        exclude:
-          - python-version: 3.6
-            plone-version: 4.3
-          - python-version: 3.6
-            plone-version: 5.0
-          - python-version: 3.6
-            plone-version: 5.1
-          - python-version: 3.7
-            plone-version: 4.3
-          - python-version: 3.7
-            plone-version: 5.0
-          - python-version: 3.7
-            plone-version: 5.1
-          - python-version: 3.8
-            plone-version: 4.3
-          - python-version: 3.8
-            plone-version: 5.0
-          - python-version: 3.8
-            plone-version: 5.1
+        python-version: [3.8, 3.7, 3.6]
+        plone-version: [5.2]
 
     steps:
       # git checkout
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.8, 3.7, 3.6]
+        python-version: [3.9, 3.8, 3.7, 3.6]
         plone-version: [5.2]
 
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.9, 3.8, 3.7, 3.6]
+        python-version: [3.8, 3.7, 3.6]
         plone-version: [5.2]
 
     steps:
--- a/news/1121.breaking
+++ b/news/1121.breaking
@@ -1,0 +1,2 @@
+- Drop support for Python 2 and Plone 5.1 and 4.3. Plone RESTAPI >= 8 supports Python 3 and Plone 5.2/6.x only.
+ [timo]
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,4 @@
+[bdist_wheel]
+# Do not make the generated wheels have `py3` tag
+# See https://packaging.python.org/guides/dropping-older-python-versions/#dealing-with-the-universal-wheels
+universal = 0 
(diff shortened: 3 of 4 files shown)


tisto:

mauritsvanrees (approved):
LGTM.

I am not sure how wise the `assert` statement in the `setup.py` is, but should be okay.
Note that with this PR merged, you should only run a `fullrelease` command which is installed by Python 3. With Python 2 it would fail at `python setup.py sdist`, due to this assertion.

One improvement could be to [disable universal wheels](https://packaging.python.org/guides/dropping-older-python-versions/#dealing-with-the-universal-wheels). When creating a wheel, you should no longer create a py2/3-compatible wheel (universal), but a Py3-only wheel. But currently I see [no wheels on PyPI](https://pypi.org/project/plone.restapi/#files) anyway.
So a `setup.cfg` with the next contents would be nice to have:

```
[bdist_wheel]
universal = 0 
```

When `zest.releaser` sees `[bdist_wheel]`, it will automatically create and upload a binary wheel to PyPI, next to the standard source distribution. Creating is done with `python setup.py bdist_wheel`. With `universal = 0` this will automatically create a py2 or py3 wheel, depending on which python version is used in the command.

Anyway, I think such a `setup.cfg` would be good to have, but can be done in a different PR. It is orthogonal to the current PR. But for clarity: a `setup.cfg` on 7.x should have `universal = 1` and on master after this PR it must have `universal = 0`, which is why I bring it up here.

ale-rt:
It feels weird to me to drop support for Python2.7 and keep support for Plone 5.2.
Will Plone 5.2 continue to ship plone.restapi 7?

mauritsvanrees:
Ah, right, I did not think it through completely. Thanks @ale-rt.
Dropping 5.1 and earlier makes a lot of sense. From a Plone 5.2 standpoint this can safely be done at any time.
Dropping Python 2.7 at some point makes sense. Question is when exactly.

Technically, the Plone coredev buildout could use the 7.x branch on Python 2 and the master branch on Python 3. But that probably gets very confusing, both for developers and for Jenkins/mr.roboto.

We could do this:

- Let coredev 5.2 use 7.x. Branch 7.x will be tested on Jenkins 5.2 and on GHA 4.3-5.2 (py 2.7, 3.6-3.8).
- Merge this PR to master.
- Let coredev 6.0 use master. Branch master will be tested on Jenkins 6.0 and on GHA 5.2 Py 3. When a 6.0 alpha is out, GHA would also test 6.0.

Put differently:
- Plone coredev 5.2 will use branch 7.x, which should become relatively stable now. The `versions.cfg` of releases will have a 7.x version.
- If you know what you are doing, you can update the pin in an individual buildout to 8.x.

@tisto Does this make sense?
Or did you have something else in mind?

mauritsvanrees:
Meanwhile, the [Jenkins 5.2-2.7 job fails](https://jenkins.plone.org/job/pull-request-5.2-2.7/1999/console) understandably:

    AssertionError: plone.restapi requires Python 3.6.0+

:-)

tisto:
@mauritsvanrees thanks for your swift feedback! Adding setup.cfg with "universal = 0" was indeed one of my open issues/questions. I just added a setup.cfg to this PR. 

This absolutely makes sense. My plan is indeed to use the master branch (8.x.x) of plone.restapi in Plone 6 only. This will make the failing Plone 5.2 / Python 2 job go away if I am not mistaken (mr.roboto should take care of this automatically). Plone 5.2 can continue using the 7.x.x branch.

I added the assert statement in setup.py to have a nice error message when trying to install plone.restapi 8 in Python 2:

```
AssertionError: plone.restapi requires Python 3.6.0+
```

I want to make sure that the first 8.x.x release of plone.restapi fails to install on Python 2 so that devs are aware and do not run into unclear error messages at a later point (or maybe into the regular python_requires message (I don't know how that would look like).

One thing that I am still wondering about is if we want to keep Python 3.6 compatibility. Plone 5.2 supports Python 3.6, right? Python 3.6 still has 6 months of official support. Therefore I kept it for now. The downside of this decision is that it will require another major release to drop Python 3.6 (though, I am not afraid of increasing version numbers). I'd be open to drop it right away as well though. @jensens @mauritsvanrees @ale-rt opinions?

tisto:

tisto:
@mauritsvanrees @ericof I merged this PR after the positive review from Maurits and now I want to check if the Jenkins jobs pass. Though, when I go to the Jenkins job I do not see any details of the job (e.g. failing tests or console log):

https://jenkins.plone.org/job/plone-5.2-python-2.7/1402/

Plone-5-2-Python-2-7-1402-Jenkins-

Either our Jenkins is broken or I lack permissions in Jenkins. I have no privileged access to Jenkins any longer so I can't check anything here or look into it further...

ale-rt:
I think it was a bad idea to merge this one.
Past Timo would be very mad at you :)

Maurits already suggested the PR was failing because of the assert. 

The console can be found in the dropdown that appears if you over the job number in the breadcrumbs:
image

You can even just add `/console` at the end of the job: https://jenkins.plone.org/job/plone-5.2-python-2.7/1402/console

tisto:
@ale-rt I hear him yelling at me. :)

Thanks for the hint. The behavior is odd. I never saw this on any other Jenkins instance so far.

I have a fix already for this:

https://github.com/plone/buildout.coredev/pull/720

Now that I can check the Jenkins results I can fix things again.

mauritsvanrees:
You probably did not see anything because the buildout failed. When a job fails this early, you see much less in the Jenkins user interface.
Anyway, Jenkins is very green after your further fix, so all is good.

Issue #1135, opened 2021-05-24
Create content even without title


ksuess:
Membrane users do not have a title, but lastname and firstname.
This causes an error on 
`def create(container, type_, id_=None, title=None):`

ksuess:
I close this issue as this problem should not be handled in plone.restapi, but in membrane.

Comment for membrane:
A fake setter helps. 
Not elegant and no final solution, but for a custom membrane type in a custom add-on, this fixes the problem.

```
    @title.setter
    def title(self, value):
        # self.context.title = value
        pass
```

PR #1136, opened 2021-05-24
Call factory without title if there is no title

ksuess:
Fixes https://github.com/plone/plone.restapi/issues/1135

### DIFF
--- a/news/1135.bugfix
+++ b/news/1135.bugfix
@@ -1,0 +1,1 @@
+Call factory without title if there is no title. @ksuess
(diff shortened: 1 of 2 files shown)


ksuess:

jensens:
Good catch. Some lines above for old 4.3 style product factories its the same. But since we're on master here, this code is dead anyway.

ksuess:
Closed with https://github.com/plone/plone.restapi/issues/1135

Issue #1107, opened 2021-04-02
Navigation endpoint should sort by object position in parent


thomasmassmann:
The latest implementation of the `navigation` endpoint uses the new navigation strategy as Plone does for the main navigation, but it does not sort the query. Moving items within a folder will affect the rendered Plone Backend, but not the frontend connected through the rest api.

The base query should be:

```diff
        query = {
            "path": {
                "query": self.navtree_path,
                "depth": self.depth,
            },
            "portal_type": {"query": self.settings["displayed_types"]},
            "Language": self.current_language,
            "is_default_page": False,
+           "sort_on": "getObjPositionInParent",
        }
```

Any thoughts? @tisto @sneridagh

sneridagh:
@thomasmassmann yes you are correct. Would you mind to create a PR for it? Thanks!

PR #1137, opened 2021-05-25
Fix navigation sort

valipod:
Navigation endpoint sort by object position in parent
(https://github.com/plone/plone.restapi/issues/1107)
NO LONGER WORKS IN plone.restapi 8.0.0


valipod:
Wrong target.

PR #1139, opened 2021-05-26
Python 3 Cleanup

tisto:


tisto:

src/plone/restapi/testing.py
```diff
@@ -57,20 +56,6 @@
 else:
     PLONE_5 = True  # pragma: no cover
```
thet, line 57: I think we can also drop this, right?

thet: I think the ``PLONE_VERSION`` should also not be necessary, some lines above.

thet (approved):
LGTM.

I love this massive reduction of code.
IMO, this is a necessary step to move forward efficiently without the burden of keeping everything backwards compatible. I hope the FTW people also see it like that and are OK with this change.

thet:
@tisto one minor thing: please add a news/towncryer entry.

src/plone/restapi/testing.py
tisto: @thet Good catch! Removed it.

tisto:

mauritsvanrees (approved):
LGTM. Great to see this code size reduction!

More could be done, but that can also happen in another PR. At least I still see lots of uses of `six`, and especially `six.PY2` which could be removed.

tisto:
@mauritsvanrees @thet thanks for the review! Yeah, I am wondering what's the best approach here. Lots of small PRs that are easy to review or one big bang PR that removes everything. I guess the current PR is big enough already, so I wouldn't add more. Next to do would be indeed to remove all the six imports and stuff.

tisto:
@thet tbh I don't think an official FWT review is necessary here. We decided that Plone 6 will be Python 3 only and this is just a cleanup of a core package that is used only in Plone 6. Though, of course, I am happy for any review and approval (official or not) from the @plone/framework-team or its members.

mauritsvanrees:
Agreed, `plone.restapi` master is for Plone 6 only since a few days or weeks, so dropping code for earlier versions is fine.

And fine to merge as is, and not create an even larger PR.

ale-rt:
I am also fine with this cleanup and welcome similar actions in the future for other packages.

P.s.: And yes, I would also like other @plone/framework-team members to express their ideas and maybe organize a meeting...

plone-4.3.x.cfg
```diff
@@ -1,39 +0,0 @@
-[buildout]
```
wesleybl, line 1: I see that `make update` is used to update the versions of this package. The `Makefile` file still tries to download cfgs from unsupported versions. It is necessary to remove these files from the `Makefile`:

https://github.com/plone/plone.restapi/blob/at-cleanup/Makefile#L27-L28

wesleybl (changes requested):

plone-4.3.x.cfg
tisto: @wesleybl thanks for the review! I am aware of this problem and I plan to put some effort into fixing it. Though, not in this PR.

tisto:
@ale-rt I am open for a fwt call. Do you see this as a blocker for a merge here? I'd prefer to merge this sooner than later because it touches quite a few files and will lead to merge conflicts when this becomes a long-running PR.

plone-4.3.x.cfg
wesleybl: OK @tisto Just for the record, I created #1144

wesleybl (commented):
Okay. Makefile will be handled in another PR

wesleybl (approved):

ale-rt:
> @ale-rt I am open for a fwt call. 

That's a great news, I will ask again in the @plone/framework-team mailing list.

> Do you see this as a blocker for a merge here? 

Absolutely not!

> I'd prefer to merge this sooner than later because it touches quite a few files and will lead to merge conflicts when this becomes a long-running PR.

I would approve the PR but the CI complains.
I will indeed approve it when green :)

tisto:

PR #1142, opened 2021-06-01
reset password: error: add message to response

ksuess:
fix https://github.com/plone/plone.restapi/issues/1143

### DIFF
--- a/src/plone/restapi/services/users/add.py
+++ b/src/plone/restapi/services/users/add.py
@@ -264,9 +264,11 @@ class UsersPost(Service):
         mt = getToolByName(self.context, "portal_membership")
         pwt = getToolByName(self.context, "portal_password_reset")
 
+        # TODO error type and error message depending on login per email or userid
         if target_user is None:
-            self.request.response.setStatus(404)
-            return
+            return self._error(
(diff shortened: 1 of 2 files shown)


ksuess:

src/plone/restapi/services/users/add.py
```diff
@@ -264,9 +264,13 @@ def update_password(self, data):
         mt = getToolByName(self.context, "portal_membership")
         pwt = getToolByName(self.context, "portal_password_reset")
 
+        # TODO error type and error message depending on login per email or userid
         if target_user is None:
-            self.request.response.setStatus(404)
-            return
+            return self._error(
+                403,
```
tisto, line 270: This would be a breaking change I guess. Maybe it makes sense to separate the custom error message from the change of the HTTP response code.

tisto: @ksuess I am currently thinking about if I'd do a breaking change release. Is this something you plan to continue to work on?

ksuess:
Thanks for asking. 
As it it not an urging need, now, for me, I close the pull request and issue.

PR #1158, opened 2021-07-05
Eliminate non-pythonic 'return None' usage

jensens:
Sorry, it hurts my eyes. So, be more pythonic and just return, pass or drop it and end of method/function.


jensens:

tisto (approved):
Fully agree :)

PR #1005, opened 2020-10-21
Handle remoteUrl field for Links

cekk:
This is related to https://github.com/plone/volto/pull/1899

I added a serializer and deserializer for TextLine field for Link objects and:
- if it's an internal link (starts with portal_url), deserialize it like "classic" Plone
- if it's an internal link, serialize it converting into a resolved url

PR #1166, opened 2021-07-08
Fix startup on Plone 4 without plone.app.contenttypes.

mauritsvanrees:
plone.restapi 7.0.0b2 introduced a serializer/deserializer for remoteUrl Link's field, in PR #1005.
This introduced an undeclared dependency on plone.app.contenttypes.
Solution: use conditional imports and use zcml:condition to register the serializer and deserializer.

Use case: collective.exportimport on Plone 4 loads `serializers/dxfields.py` and this fails without plone.app.contenttypes.

### DIFF
--- a/news/1166.bugfix
+++ b/news/1166.bugfix
@@ -1,0 +1,2 @@
+Fix startup on Plone 4 without plone.app.contenttypes.
+[maurits]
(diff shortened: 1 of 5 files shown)


mauritsvanrees:

tisto (approved):

PR #1173, opened 2021-07-10
Set UID of a content during creation

ericof:
Allow setting UID of a content during creation if the user has Manage Portal permission.
If the user has no permission, the content is created but UID is not set.

### DIFF
--- a/src/plone/restapi/services/content/add.py
+++ b/src/plone/restapi/services/content/add.py
@@ -1,3 +1,4 @@
+from AccessControl import getSecurityManager
 from Acquisition import aq_base
 from Acquisition.interfaces import IAcquirer
 from plone.app.multilingual.interfaces import IPloneAppMultilingualInstalled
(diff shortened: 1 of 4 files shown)


ericof:

ericof:

src/plone/restapi/services/content/add.py
```diff
@@ -74,6 +77,11 @@ def reply(self):
         if temporarily_wrapped:
             obj = aq_base(obj)
 
+        # ManagePortal is required to set the uid of an object during creation
+        sm = getSecurityManager()
+        if uid and sm.checkPermission(ManagePortal, self.context):
```
tisto, line 82: @ericof I am wondering if we should fail silently here or issue a message (at least on the console) or even fail hard when someone tries to set a UID without the proper permission?

jensens: Please fail with an `Unauthorized` and a reason.

ericof: 👍🏼

ericof:

tisto (approved):
LGTM. We might want to add a paragraph about this feature in our docs.

ericof:

Issue #1181, opened 2021-07-22
[9.x.x] Refactor @lock endpoint based on CRUD operations and add it to expandable @components


avoinea:
Verb | URL | Action
-- | -- | --
POST | /@lock | Lock an object
GET | /@lock | Current lock information
PATCH | /@lock | Refresh lock
DELETE | /@lock | Remove lock

We should also be able to use `expand=lock` in order to easily get locking information in one Rest API call.

@tisto @sneridagh @tiberiuichim thoughts?!

tiberiuichim:
+1 looks good, seems semantic.

tisto:
@avoinea this makes sense to me. Though, we have to be aware that this is a breaking change and requires a new major release. I saw that you kept the old code to be backward compatible. This might be a way to go to avoid a major release.

Though, the question is if this effort does make sense. I could imagine that not lots of people are using this anyways. Maybe it is cleaner to make a breaking release and get rid of the old code altogether. Just thinking out loud here...

avoinea:
@tisto As we will need to cherry-pick it in 7.x I think is best to do it in 2 steps, first deprecate, then drop it and major release :wink:

tisto:
@avoinea right. We won't be able to drop the functionality in the 7.x branch (at all).

tisto:
@avoinea could you image to work on a PR for master that removes the old lock endpoint behavior for plone.restapi 9? I'd like to have a set of PRs ready once we are ready to move on to version 9 (with Plone 6).

avoinea:
@tisto I think that's an easy one. :beer:

PR #1182, opened 2021-07-22
Refactor `@lock` endpoint based on CRUD operations

avoinea:
See #1181

### DIFF
--- a/docs/source/locking.rst
+++ b/docs/source/locking.rst
@@ -6,6 +6,17 @@ Locking is a mechanism to prevent users from accidentially overriding each other
 When a user edits a content object in Plone, the object is locked until the user hits the save or cancel button.
 If a second user tries to edit the object at the same time, she will see a message that this object is locked.
 
+The API consumer can create, read, update, and delete a content-type lock.
+
+======= =============================================== =====================================================================
+Verb    URL                                             Action
+======= =============================================== =====================================================================
+POST    ``/@lock``                                      Lock an object
+GET     ``/@lock``                                      Get information about the current lock
+PATCH   ``/@lock``                                      Refresh existing lock
+DELETE  ``/@lock``                                      Unlock an object
+======= =============================================== =====================================================================
+
 
 Locking an object
 -----------------
@@ -42,7 +53,7 @@ The server responds with status :term:`200 OK` and returns the lock information.
 Unlocking an object
 -------------------
 
-To unlock an object send a POST request to the ``/@unlock`` endpoint.
+To unlock an object send a DELETE request to the ``/@lock`` endpoint.
 
 ..  http:example:: curl httpie python-requests
     :request: ../../src/plone/restapi/tests/http-examples/unlock.req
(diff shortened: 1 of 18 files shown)


avoinea:

avoinea:

tisto:

docs/source/locking.rst
```diff
@@ -42,7 +53,7 @@ The server responds with status :term:`200 OK` and returns the lock information.
 Unlocking an object
 -------------------
 
-To unlock an object send a POST request to the ``/@unlock`` endpoint.
```
tisto, line 45: @avoinea shall we add a section here for the deprecated methods that we have and that we will drop in plone.restapi 9?

src/plone/restapi/services/locking/configure.zcml
```diff
@@ -1,17 +1,46 @@
 <configure
-    xmlns="http://namespaces.zope.org/zope"
-    xmlns:plone="http://namespaces.plone.org/plone"
-    xmlns:zcml="http://namespaces.zope.org/zcml"
-    >
+  xmlns="http://namespaces.zope.org/zope"
```
tisto, line 2: Just wondering why zpretty is not complaining here? Either before or after. :)

src/plone/restapi/services/locking/configure.zcml
```diff
@@ -1,17 +1,46 @@
...
+      factory=".get.Lock"
+      for="zope.interface.Interface"
+      permission="zope2.View"
+      name="@lock"
+      />
 
   <plone:service
       method="POST"
-      factory=".locking.Lock"
-      for="Products.CMFCore.interfaces.IContentish"
```
tisto, line 10: @avoinea why did we move from Contentish to zope Interface here? Also for the other services. Locking does not work on the SiteRoot anyways, right?

tisto (approved):
@avoinea apart from nitpicking this looks good to me!

src/plone/restapi/services/locking/configure.zcml
avoinea: @tisto True. But in the front-end the Edit form is generic and it sends a `@lock` request which would return 404 if the `@lock` endpoint is not registered on a more generic interface.

src/plone/restapi/services/locking/configure.zcml
avoinea: Fixed.

docs/source/locking.rst
avoinea: Added.

avoinea:
Maybe we should also serialize the `locking info` on dxcontent as we do with `working_copy`

https://github.com/plone/plone.restapi/blob/60543ac4a6f79fe5c6378d42a711229ecb0a0a1a/src/plone/restapi/serializer/dxcontent.py#L78-L80

avoinea:
> Maybe we should also serialize the `locking info` on dxcontent as we do with `working_copy`

I did it in another branch in order to be able to discard it easily if not needed https://github.com/plone/plone.restapi/pull/1201

src/plone/restapi/services/locking/configure.zcml
tisto: @avoinea can you elaborate? I don't get it. :)

avoinea: I think we may not need this if we agree on #1201

avoinea:

avoinea:
@tisto I think this is ready and complete now.

tisto (approved):
LGTM. @avoinea just one question: why do we need an expandable when we serialize the lock information into the regular response?

avoinea:
@tisto As we don't use it in Volto, yes, I think we can drop the expandable, also in order to have only one way of getting the `lock` info.

tisto:
@avoinea I will wait then for this before pushing the merge button. :)

avoinea:

PR #1190, opened 2021-08-02
Skip doActionFor for items without workflow

cekk:
Do not break with a 400 if we are trying to change wf for a content without workflow.

I wanted to fix that on children contents but right now it does not return a 400 also if we try to call this action directly on a content without a workflow. Should we keep the 400 error in this case?


cekk:

tiberiuichim (approved):

PR #1192, opened 2021-08-09
Handle timezone in publication fields (effective and expires)

cekk:
This could fix plone/volto#2597

The failing test is because also if i customize env var, i can't save a DateTime value in that field with the wanted timezone.

Anyone have an idea on how to fix it? @buchi ?


tisto:

avoinea:
Holy mother of... this is here for 2 years and nobody took care of it? :smile: 

I can confirm @cekk 's conclusions from here https://github.com/plone/volto/issues/2597#issuecomment-895114458  

How to test and reproduce:

1. Start plone backend with TZ env

```
docker run -it --rm -p 8080:8080 -e TZ="Europe/Berlin" plone/plone-backend
```

2. In Volto publish a News items with publishing date and time after a half an hour
3. In Volto add a listing block showing all news before today
4. The News should not be listed yet, but it is.
5. Play with the Publishing Date until the News item disappear from the listing block
6. Check the portal_catalog entry and you'll notice the EffectiveDate is actually `UTC - 2` :scream:

src/plone/restapi/deserializer/dxfields.py
```diff
@@ -97,12 +101,17 @@ def __call__(self, value):
...
         # We try to guess the correct deserialization from the current field
         # value.
-        dm = queryMultiAdapter((self.context, self.field), IDataManager)
-        current = dm.get()
-        if current is not None:
-            tzinfo = current.tzinfo
+        is_publication_field = self.field.interface == IPublication
+        if is_publication_field:
+            # because IPublication datamanager strips timezones
+            tzinfo = pytz.timezone(default_timezone())
```
avoinea, line 107: @cekk While this is way better than assuming UTC always, it can still have issues if the server has different timezone than the one from registry, **UTC** for example (the default behavior of `plone/plone-backend` Docker image).

And I think @sneridagh can confirm that their installations doesn't provide the TZ environment variable to docker containers, thus **UTC** by default. While the default timezone registry entry when creating a new Plone instance is `Europe/Berlin`.

How to test:
```
docker run -it --rm -p 8080:8080 -e SITE=Plone plone/plone-backend bash

git clone https://github.com/plone/plone.restapi.git
cd plone.restapi
git checkout handle_tz_in_publication_fields
cd /app
bin/pip install -e ./plone.restapi --no-deps
./docker-entrypoint.sh start
```

Now add a News Item with Effective Date via Volto and check portal_catalog entry.

avoinea (changes requested):

erral:
We have been hit by this too, and I can confirm that with this changes the original datetime as entered in Volto interface is correctly saved in Plone.

We configure our Zope instance passing the `TZ = Europe/Madrid` environment variable.

In all cases here, I have `Europe/Madrid` configured as a default timezone in Plone.

Item saved with the following datetime through Volto: 2023-08-28 10:00:00

Previous scenario (without this branch):

```
>>> news = app.Plone.komunikazioa.albisteak['albistea-2023-08-28-10-00-00']
>>> news
<FolderishNewsItem at /Plone/komunikazioa/albisteak/albistea-2023-08-28-10-00-00>
>>> value = news.effective()
>>> value
DateTime('2023/08/28 08:00:00 GMT+2')
>>> value = value.asdatetime()
>>> value
datetime.datetime(2023, 8, 28, 8, 0, tzinfo=<StaticTzInfo 'GMT+2'>)
```

Current scenario (with this branch)
```
>>> news = app.Plone.komunikazioa.albisteak['albistea-2023-08-28-11-00']
>>> news
<FolderishNewsItem at /Plone/komunikazioa/albisteak/albistea-2023-08-28-11-00>
>>> news.effective()
DateTime('2023/08/28 10:00:00 GMT+2')
>>> news.effective().asdatetime()
datetime.datetime(2023, 8, 28, 10, 0, tzinfo=<StaticTzInfo 'GMT+2'>)
```

I have made a test *without* adding the `TZ` environment variable in the instance configuration and the behavior is the following:

Previous scenario (without this patch)

```
>>> news = app.Plone.komunikazioa.albisteak['albistea-2023-08-28-12-00']
>>> news
<FolderishNewsItem at /Plone/komunikazioa/albisteak/albistea-2023-08-28-12-00>
>>> value = news.effective()
>>> value
DateTime('2023/08/28 08:00:00 GMT+2')
>>> value.asdatetime()
datetime.datetime(2023, 8, 28, 8, 0, tzinfo=<StaticTzInfo 'GMT+2'>)
```

Current scenario (with this branch):

```
>>> news = app.Plone.komunikazioa.albisteak['albistea-2023-08-28-12-00']
>>> newd
<FolderishNewsItem at /Plone/komunikazioa/albisteak/albistea-2023-08-28-12-00>
>>> value = news.effective()
>>> value
DateTime('2023/08/28 10:00:00 GMT+2')
>>> value = value.asdatetime()
>>> value
datetime.datetime(2023, 8, 28, 10, 0, tzinfo=<StaticTzInfo 'GMT+2'>)
```

stevepiercy:
@tisto @cekk @erral @avoinea @sneridagh can we add this to the Volto Team meeting agenda? This bug has been hanging around for almost 3 years now without resolution. The tests are broken, too.

sneridagh:
@davisagli @jensens @thet related with current dates handling. Aside from the new fixes that we are working on already, we probably will need this patch for now. Are you able to take a look?

jensens (changes requested):
I do not think this nails it. Checking specifically for IPublication is wrong, ugly and is a recipe for harmful future side effects.

The dates of effective/expiration/created/modified are all still Zope DateTime instances = best would to get rid of those and convert to python datetime. But this is lots of work and even if it was planned for 6.0 nobody took time and energie to make it happen. Maybe for 7.0? 

One can not rely on it being a Zope Datetime: It could be a Python datetime as well.

To convert Zope DateTime  to a Python datetime we have a function `pydt` in `plone.event.utils.
The right way for the serializer is to check if there is a Zope DateTime returned and if so, use pydt to convert it. The `missing_zone` parameter passes in a pytz zone to be used, if no timezone is present.

I propose to change this to a generic solution to deal with Zope DateTime instead of covering just one special case here.

claytonc:
Same problem volto version 16.32.1 and  plone version 6.0.13, I had to make this correction because the content was not being published on the selected date and time.

davisagli:

davisagli:
@cekk @avoinea @erral @stevepiercy @jensens @stevepiercy I think I (finally) have this in a state where it can be merged now.

I also tested in a docker container per @avoinea's [instructions](https://github.com/plone/plone.restapi/pull/1192#discussion_r1161918740) (thanks, helpful!)

@jensens This is in fact a special case needed for the fields from IPublication, because they use the [DCFieldProperty adapter](https://github.com/plone/plone.app.dexterity/blob/master/plone/app/dexterity/behaviors/metadata.py#L266) which expects timezone-naive datetimes. It would be cleaner to fix it there, and maybe we can in Plone 6.2 or 7, but I don't think it can be done in a backwards-compatible way. So I'll accept this as a pragmatic workaround in the meantime.

I plan to merge this tomorrow unless someone finds that it is not working correctly.

PR #1194, opened 2021-08-12
Fixes #1193: deprecated JWT decode.

### LINKED ISSUE #1193: Does not work with a recent PyJWT
Test are failing with `pyjwt = 2.1.0`. We are using an outdated PyJWT 1.7.1 (as of the year 2018).

jensens:
- `verify` as kwarg of `decode` was already deprecated in PyJWT and was removed in 2.0. Uses now `options` style to pass this on.
- `encode` returns a str (Py3!) which can not be decoded again.
- Update to use and require PyJWT 2.1.0.

Note:
After merge, buildout.coredev needs to be updated. 
I prepared an separate PR for this: plone/buildout.coredev#735
This is difficult to test with our current Jenkins setup.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,6 +55,7 @@ setup(
         "Environment :: Web Environment",
         "Framework :: Plone",
         "Framework :: Plone :: 5.2",
+        "Framework :: Plone :: 6.0",
         "Framework :: Plone :: Core",
         "Intended Audience :: Developers",
         "Operating System :: OS Independent",
@@ -62,6 +63,7 @@ setup(
         "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
+        "Programming Language :: Python :: 3.9",
         "Programming Language :: Python :: 3 :: Only",
         "Topic :: Software Development :: Libraries :: Python Modules",
     ],
@@ -82,7 +84,7 @@ setup(
         "plone.behavior>=1.1",  # adds name to behavior directive
         "plone.rest >= 1.0a6",  # json renderer moved to plone.restapi
         "plone.schema >= 1.2.1",  # new/fixed json field
-        "PyJWT",
+        "PyJWT>=2",
         "pytz",
     ],
     extras_require={"test": TEST_REQUIRES},
--- a/versions.cfg
+++ b/versions.cfg
@@ -26,3 +26,6 @@ cffi = 1.14.4
 # requirement for json widget tests to pass
 plone.schema = 1.3.0
 plone.dexterity = 2.9.8
+
+# recent pyjwt 
+pyjwt = 2.1.0
--- a/README.rst
+++ b/README.rst
@@ -63,13 +63,6 @@ Install plone.restapi by adding it to your buildout::
 
 and then running ``bin/buildout``
 
-Usage in Plone 5.2:
-
-- Older plone.restapi 7.x. is part of the Plone 5.2.x release series and works on both Python 2.7 and 3.6 to 3.8.
-- plone.restapi 8.x or later works with Plone 5.2:
-  - if Python 3.6 or later is used and
-  - since plone.restapi 8.9.0 if PyJWT 2.1.0 is used. Set a pin `pyjwt = 2.1.0` in the `[versions]` section of the buildout file. 
-
 
 Contribute
 ==========
--- a/README.rst
+++ b/README.rst
@@ -63,6 +63,13 @@ Install plone.restapi by adding it to your buildout::
 
 and then running ``bin/buildout``
 
+Usage in Plone 5.2:
+
+- Older plone.restapi 7.x. is part of the Plone 5.2.x release series and works on both Python 2.7 and 3.6 to 3.8.
+- plone.restapi 8.x or later works with Plone 5.2:
+  - if Python 3.6 or later is used and
+  - since plone.restapi 8.9.0 if PyJWT 2.1.0 is used. Set a pin `pyjwt = 2.1.0` in the `[versions]` section of the buildout file. 
+
 
 Contribute
 ==========
--- a/news/1193.bugfix
+++ b/news/1193.bugfix
@@ -1,0 +1,3 @@
+Fixes deprecated JWT `decode`usage.
+Uses and requires latest PyJWT 2.1.0 now. 
+[jensens] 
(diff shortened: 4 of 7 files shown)


versions.cfg
```diff
@@ -25,4 +25,7 @@ cffi = 1.14.4
 
 # requirement for json widget tests to pass
 plone.schema = 1.3.0
-plone.dexterity = 2.9.8
\ No newline at end of file
+plone.dexterity = 2.9.8
+
+# recent pyjwt 
+pyjwt = 2.1.0
```
wesleybl, line 31: I would put a comment here to remove this pin after release Plone version with coredev PR.

jensens: It is needed for Plone 5.2 - which is still supported with Py3 here at master.

wesleybl: You won't update the version in the 5.2 branch of coredev? At least for Python 3?

jensens: Well, with conditional version sections this would be technically possible. 

Since the 8.x series is not part of the release of Plone 5.2 (see https://github.com/plone/buildout.coredev/blob/5.2/versions.cfg#L197 and https://github.com/plone/buildout.coredev/blob/5.2/sources.cfg#L118, it uses the 7.x) I do not think this is possible. 

We could add a note to the README which versions are preconditions are needed for use of the 8 series in Plone 5.2.

wesleybl: Anyway, it's weird to say that version 8.x is compatible with Plone 5.2 and it will not be tested on coredev. If any unreleased changes in any package, break the version 8.x, we won't know about it until the Plone version is released.

jensens: Therefore there are the GH Actions test I think. It works against the latest release, I think this is good enough?

tisto: @mauritsvanrees would it be possible to pin pyjwt to 2.1.0 in the next Plone 5.2.6 release?

tisto:

versions.cfg
mauritsvanrees: @tisto We could pin PyJWT to 2.1.0 for Python 3 in Plone 5.2.
But I think plone.restapi 7.x.x needs a small change then.
It uses the deprecated `verify` parameter, which [is now dropped](https://pyjwt.readthedocs.io/en/stable/changelog.html#dropped-deprecated-verify-param-in-jwt-decode).

mauritsvanrees: Ah, that is exactly what this PR fixes. Except that it should also happen on 7.x.x, but not the `setup.py` change (or that part only for Python 3).

jensens: PyJWT 2.x is Python 3 only, so not an option for 5.2.

Keep it as is for Plone 5.2 in the release. Old PyJWT 1.x plus plone.restapi 7.x - it works.

If one wants to use next plone.restapi 8 in Plone 5.2 buildout fails with a message about PyJWT 2 version conflict. I'll add this to the README->installation docs to pin pyJWT to >=2.1.0.

jensens: Done

jensens:

jensens:

jensens:

jensens:
Jenkins will always fail here. There no easy way to test a branch on buildout.coredev together with branch on a package. Except on an own branch of the branch of buildout.coredev - I did this here https://github.com/plone/buildout.coredev/pull/737

versions.cfg
tisto: @jensens @mauritsvanrees I am wondering if we should defer this to the next major release of plone.restapi (9.x). Once plone.volto has seen a first release, we will need to move some things around (e.g. the block behavior needs to move from restapi to plone.volto). The plan is to have stuff ready for the conference, so it won't take so long. What do you think?

mauritsvanrees: It looks like the new code from this PR also works with the old PyJWT. At least the tests pass.
So I think we can include this code even in 7.x.x. Just don't touch `setup.py`.
On PY3, `token.decode("utf-8")` should still be done, but only if the token is bytes, which PyJWT 1.7.1 gives us.

So that is one option: be a bit smarter in the code so we can support both versions of PyJWT.

Other option is to only merge the current PR when `plone.restapi` master only targets PY 3, which seems to be what Timo is saying.

I don't really mind either way.

jensens:
> Except on an own branch of the branch of buildout.coredev

Nope, won't work as well.

versions.cfg
jensens: I rebased the branch.

jensens:
Summary:
- all done
- this works only with newer PyJWT
  - *without* Python 2 support
  - but then also in Plone 5.2 
- tests are failing as unless buildout.coredev PyJWT pin is upgraded, it pins the old version.

mauritsvanrees:
For me, it would be fine to let `plone.restapi` master use latest pyjwt, as the branch is already for Python 3 only. New minor version. Maybe new major version, but for me it would not be needed.

You changed the `versions.cfg` here to pin `pyjwt = 2.1.0`, but this file is currently not used. Maybe `base.cfg` should extend it. @tisto may know how this is intended.

setup.py
```diff
@@ -55,13 +55,15 @@ def read(filename):
...
         "Framework :: Plone :: 5.2",
+        "Framework :: Plone :: 6.0",
         "Framework :: Plone :: Core",
         "Intended Audience :: Developers",
         "Operating System :: OS Independent",
         "Programming Language :: Python",
         "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
+        "Programming Language :: Python :: 3.9",
```
tisto, line 66: @jensens Python 3.9 is not officially supported or tested with plone.restapi yet: https://github.com/plone/plone.restapi/blob/master/.github/workflows/tests.yml#L9

mauritsvanrees: It is tested in coredev.

jensens: Yes, Python 3.9 must work for Plone 6, so running tests with it here is important. I added issue #1345 to not forget about it.

mauritsvanrees:
Green now.
(Except plone-coredev 6, which can only pass when we update the PyJWT pin there, which we can only do after we merge this PR. Chicken versus egg.)

mauritsvanrees:
I am tired of waiting on someone to either approve and merge this, or say what needs to change.
So I have made a new PR #1377. It takes the main change of the current PR, but makes it conditional: the code works on both `PyJWT` version 1 and 2. That change should be easier to make a decision on and merge soon, instead of being in limbo since August.
The cost is an extra dependency (`importlib-metadata`) on Python 3.7 and lower, but this is likely already pulled in by other packages, and it is already pinned by Plone 5.2 and 6.0. (I could have used `pkg_resources`, but that is apparently deprecated for this use case.)

jensens:
solved with bbb #1377

PR #1195, opened 2021-08-14
Fix tests that involving dates in Python 3.8 on machines that don't use TZ = UTC

wesleybl:
In Python 3.8, for the TZ environment variable to be recognized, you need to call `time.tzset()`.

Ref: https://github.com/plone/buildout.coredev/pull/701


base.cfg
```diff
@@ -63,6 +63,8 @@ recipe = zc.recipe.testrunner
 eggs = ${instance:eggs}
 initialization =
     os.environ['TZ'] = 'UTC'
+    import time
```
tisto, line 66: @wesleybl thanks for your PR! It would help to add a line or two of explanation here why we added this.

wesleybl: Done.

jensens (approved):
LGTM

Issue #1197, opened 2021-08-16
Variable interpolation for internal links is not operative, even in classic


sneridagh:
In classic, it claims that variable interpolation still works:

<img width="776" alt="image" src="https://user-images.githubusercontent.com/486927/129552230-12c683ca-bbc4-4be0-9b44-9ab19a0cf559.png">

but there's no way to set the link "manually" because the widget is in the middle, and forces you to choose it.

Bug in the navigation portlet (in classic, that revolves into the ContextNavigation component in Volto):

1. If you set up an internal link (using the widget)
2. make it public
3. Enable the navigation portlet to see the link content type listed
4. logout, check the navigation portlet

then it shows:
<img width="636" alt="image" src="https://user-images.githubusercontent.com/486927/129552793-13be0b4e-db07-48eb-b6dd-5fc07be3b9b3.png">

A resolveuid URL pointing to: https://volto.kitconcept.dev/Plone/resolveuid/d368b2a5931a454f8a1340035519b0a9

rooted in /Plone (even with a proper domain and VHM in place (the name of the Plone instance in ZMI).

So it kind of works, because the resolveuid and acquisition :) so the bug is hidden in classic but the internals are completely broken.

Back in p.restapi and Volto, we tried to comply with the classic spec, using the variable interpolation, but since it's broken, the result query to the catalog returns:

<img width="854" alt="image" src="https://user-images.githubusercontent.com/486927/129553390-088ff2e4-0a76-49cc-bedd-07fc082035ec.png">

href (coming from https://github.com/plone/plone.app.contenttypes/blob/master/plone/app/contenttypes/indexers.py#L156) and getRemoteUrl index contents, index "/name_of_the_Plone_instance/whatever" then the ContextNavigation can't show correctly the remoteURL

sneridagh:
My 5c, let's rip this out, even from Plone 6 at some point, since it's both unusable and broken.

sneridagh:
Any reason to treat an internal link differently than an external one?

tiberiuichim:
> but there's no way to set the link "manually" because the widget is in the middle

Would it work if you use the External tab?

tisto:
My 2c: let's rip it out. Both in Classic and Volto things should be hidden behind a proper widget.

sneridagh:
> > but there's no way to set the link "manually" because the widget is in the middle
> 
> Would it work if you use the External tab?

Didn't have time to check :(

mauritsvanrees:
[...]

PR #1198, opened 2021-08-16
Add resolveuid to Link content type `remoteUrl` field. Remove the variable interpolation support since it was broken anyways.

sneridagh:


ericof (approved):
LGTM

sneridagh:

src/plone/restapi/tests/http-examples/link.resp
```diff
@@ -1,66 +1,38 @@
-HTTP/1.1 200 OK
+HTTP/1.1 500 Internal Server Error
```
avoinea, line 1: Is this expected?

sneridagh: nope! :P

sneridagh: @avoinea Fixed

sneridagh:

avoinea (approved):

cekk:
i'm ok with these changes, but does this broke "classic Plone" compatibility?
I mean: if i create a Link in Volto, then in Plone everything works fine (the url resolver and the widget)? vice versa should works

sneridagh:
@cekk I left the serializer in place, so if it has a "{portal_url}" it gets decoded. However, the compatibility was not complete either (navigation_root was not covered). I think it's better to rip it off from both sides.

sneridagh:

src/plone/restapi/deserializer/blocks.py
```diff
@@ -32,42 +29,6 @@ def iterate_children(value):
             queue.extend(child["children"] or [])
 
 
-def path2uid(context, link):
```
tisto, line 35: @sneridagh nitpicking: I usually prefer to keep util functions in the location where they are used. As long as a util function is only used in one location, moving it to its own file just makes it harder to understand what goes on.

tisto (approved):

src/plone/restapi/deserializer/blocks.py
sneridagh: For the record, I moved them to its own module due to circular import issues.

mauritsvanrees (approved):
After careful consideration (see comments in issue #1197) this LGTM.

Perhaps change the news snippet, because variable interpolation support is not completely removed: restapi can still read such a value in both serialiser and deserialiser. It just saves it slightly differently.

sneridagh:

src/plone/restapi/deserializer/blocks.py
tiberiuichim: I'd argue that a Python library also has to consider the API it exposes as public. Would an import such as `from plone.restapi.blocks import path2uid` make sense? Not much. If the `path2uid` would be an internal util used only in the blocks.py module, then it should also be marked as internal, like `_path2uid`.

PR #1201, opened 2021-08-18
Serialize lock information on dx content

avoinea:


tisto:
@avoinea is this PR ready for review? Could you add a changelog entry?

avoinea:
Yes. Changelog not needed as it is already within the base branch.

tiberiuichim:
What about "can_unlock" instead of "stealable" ?

tisto:
@avoinea I don't get it. Don't you want me to merge this PR first and then the older one? Then we need a changelog entry.

tiberiuichim:
@tisto I think @avoinea is looking for feedback on these particular set of changes, as a separate part of the main lock PR. He can do the merge into the main Lock PR once he's satisfied, I guess.

tisto:
@tiberiuichim @avoinea the code in this PR looks good to me. If we have a changelog entry I can merge it right away. I am also ok with having one big PR with everything.

tiberiuichim:
@tisto @avoinea is now in vacation, that's why I was answering for him. I think we should wait until he returns?

PR #1205, opened 2021-08-20
Fix @vocabularies endpoint to search in translated term titles

sneridagh:


tisto (approved):
LGTM. @sneridagh maybe we should add another test with an "Umlaut" in the term, to make sure this works as well. What do you think?

sneridagh:
@tisto it was already there! 
https://github.com/plone/plone.restapi/blob/66fee76c953b7b8aa3aad4cd0adea334ad9c6b18/src/plone/restapi/tests/test_services_vocabularies.py#L158

I didn't add a proper translation test, because it might be tricky (I think we don't have them anywhere anyways).

sneridagh:

sneridagh:
@tisto what it's true is that Volto is not encoding it... let's see if it works anyways...

sneridagh:

tisto (approved):

PR #1209, opened 2021-08-30
Escape some chars in search parameters

cekk:
This is related to [this post](https://community.plone.org/t/parseerror-when-search-term-contains-parenthesis-in-zcatalog-search/13424).

Is a common problem with ZTCTextIndex and this pr will try to fix the problem with searches made with plone.restapi only.

The regexp right now only remove parenthesis from the value. Do we need to manage also other characters?


tiberiuichim (commented):
I don't think it's safe to do this, see https://owasp.org/www-community/attacks/Regular_expression_Denial_of_Service_-_ReDoS

tiberiuichim:
Oh, sorry, I misread your code. I'll remove my comments

cekk:

src/plone/restapi/tests/test_query_parsers.py
```diff
@@ -42,6 +42,10 @@ def test_casts_complex_query_values_to_string(self):
         qp = ZCTextIndexQueryParser()
         self.assertEqual({"query": "42"}, qp.parse({"query": 42}))
 
+    def test_strip_parenthesis_from_value(self):
+        qp = ZCTextIndexQueryParser()
+        self.assertEqual("foo bar", qp.parse("foo (bar)"))
```
tisto, line 47: Nitpicking: maybe add the other parenthesis here as well. Just to document that it works for all of them. A dev who reads this test might wonder about it. Just two more lines :)

tisto:
@cekk LGTM. I am just wondering what other characters we would have to strip that would break zcatalog. Would it be possible to write a set of unit tests to just try what works and what not?

tisto:

cekk:
@tisto i improved tests but i've found some other edge cases that generate exceptions and i don't know if we want to handle or not:

- **not** (already fixed in my pr and tested)
- single chars £ $ * &, etc: ZCTextIndex returns an error if you set SearchableText with one of the special characters

cekk:

tisto (approved):
LGTM. A second review would be nice. @cekk sorry it took me so long to check this PR. Busy days... :)

cekk:

cekk:
@tisto @tiberiuichim any chance to have this pr merged?

davisagli:
I think this is no longer needed since https://github.com/plone/plone.app.querystring/issues/111 was fixed in plone.app.querystring

Issue #1096, opened 2021-03-23
Upgrade to Plone 5.2.4 blocked (because of plone.dexterity 2.10.0)


tisto:
plone.dexterity 2.10.0 introduced a dynamic schema change that breaks our documentation test. See
https://github.com/plone/plone.dexterity/pull/137 for details. A workaround is to pin plone.dexterity to 2.9.8 but that isn't a long-term solution.

avoinea:
@tisto Can you please point me to the failing running commands as running tests on my python3.8 sandbox doesn't seem to fail. 
I can see a hardcoded python 3.7 condition within github actions (it would help if we can switch to python 3.8)

sneridagh:
@avoinea I recall yesterday building it on a 3.8 and see the dynamic behaviour names too... Maybe you didn't build again and you were using an old build?

PR #1214, opened 2021-09-02
Fix testing of a checkout instead of a released package.

mauritsvanrees:


wesleybl (changes requested):
@mauritsvanrees It keeps getting the released version, because in `plone-5.2.x.cfg` `https://dist.plone.org/release/5.2.5/versions.cfg` takes precedence over `base.cfg`:

https://github.com/plone/plone.restapi/blob/60543ac4a6f79fe5c6378d42a711229ecb0a0a1a/plone-5.2.x.cfg#L2-L4

It is necessary to change there too.

Maybe it's good to change in

https://github.com/kitconcept/buildout/blob/5.2/plone-5.2.x.cfg

also, since `plone.restapi` uses this repository as base.

tisto:
@mauritsvanrees we ran into this problem quite a few times already. If I am not mistaken buildout only takes the empty version pin into account on the buildout file that is called via command line. Not in any of the included buildout files. Therefore we need to add this to plone-5.2.x.cfg. I think @wesleybl is correct here.

wesleybl:
@tisto empty pin works in `base.cfg`. It isn't working because `https://dist.plone.org/release/5.2.5/versions.cfg` takes precedence over `base.cfg`. It's necessary to change the order in extends:

```diff
 extends = 
-     base.cfg 
-     https://dist.plone.org/release/5.2.5/versions.cfg 
+     https://dist.plone.org/release/5.2.5/versions.cfg 
+     base.cfg 
```

mauritsvanrees:
Sorry, I should have checked.
I have added the pin to `plone-5.2.x.cfg` now. This command shows that it is working now:

```
$ bin/buildout -c plone-5.2.x.cfg annotate | grep -A1 restapi
...
plone.restapi=
    plone-5.2.x.cfg
```

Wesley is right though: a possibly better alternative  would be to let `plone-5.2.x.cfg` first extend the 5.2 dist.plone.org versions, and then `base.cfg`, and add versions in `base.cfg`. But that might interfere with the standard kitconcept setup and the Makefile, I'm not sure.

wesleybl (changes requested):
Please change the order in extends and put the empty pin in base.cfg. Leaving in the order it is now may cause more pin problems in the future.

mauritsvanrees:
@wesleybl 

> Please change the order in extends and put the empty pin in base.cfg. Leaving in the order it is now may cause more pin problems in the future.

Done.
I have squashed the commits and force pushed.

Looks like the test for uncommitted changes will fail though, at least [this line for the lock token](https://github.com/plone/plone.restapi/runs/3497597204?check_suite_focus=true#step:12:594).

wesleybl (changes requested):
> Looks like the test for uncommitted changes will fail though, at least this line for the lock token.

@mauritsvanrees Take a look at:

https://github.com/plone/plone.restapi/issues/1096

You can pin `plone.dexterity = 2.9.8` in `plone-5.2.x.cfg`.

But there may be other errors that were hidden because of the wrong pin.

mauritsvanrees:
Ah, thanks. I have re-added the `plone.dexterity` pin, with a comment on why we have it.

Only remaining problem that I see, is the [random lock token](https://github.com/plone/plone.restapi/runs/3498322161?check_suite_focus=true#step:12:45) in the http examples.
We might need to do some normalisation of `response.text` in the `save_request_and_response_for_docs` function. That could help for random lock tokens and random schema names.

wesleybl:
I recommend commenting out the documentation testing in this PR so we can merge. The version problem is more serious than the documentation problem. Fix of documentation can be done in another PR.

mauritsvanrees:
> I recommend commenting out the documentation testing in this PR so we can merge. The version problem is more serious than the documentation problem. Fix of documentation can be done in another PR.

Done. I will start on a separate PR for the documentation.

wesleybl (approved):

wesleybl:
The 7.x.x branch has the same problem.

wesleybl:
@tisto consider making these changes to

https://github.com/kitconcept/buildout/blob/5.2/plone-5.2.x.cfg

All repositories that use it will have the same problem.

.github/workflows/ci.yml
```diff
@@ -52,11 +52,14 @@ jobs:
         run: bin/test
 
       # test no uncommited changes
-      - name: test no uncommited changes
-        run: bin/test-no-uncommitted-doc-changes
-        env:
-          PYTHON_VERSION: ${{ matrix.python-version }}
-          PLONE_VERSION: ${{ matrix.plone-version }}
+      # TODO Temporarily commented out.
```
tisto, line 55: @mauritsvanrees I am somehow reluctant to merge this PR because we comment out this check here. I know there are quirks but this works reliably for quite a while now. Can we try to fix things step by step? First the checkout, then the normalization?

tisto:
@mauritsvanrees how did you build plone.restapi? I just did a checkout, ran "make", then amended a test to fail, ran "bin/test" and the test fails. To make sure this is not just test runner magic, I broke code in restapi and ran "bin/test". Tests fail, as expected.

wesleybl:
@tisto GA uses `plone-5.2.x.cfg`:

https://github.com/plone/plone.restapi/blob/3550562b1ad0fedec1538bbfe565befd8110f11c/.github/workflows/ci.yml#L38

You have to use it too to see the problem:

```
./bin/buildout -c plone-5.2.x.cfg
```

.github/workflows/ci.yml
wesleybl: @tisto as the wrong version was being used, there were errors in the documentation that were "hidden". I think the best thing to do is to merge this PR with the documentation tests commented on and then fix the tests. Using the wrong version is a more serious mistake than not running the documentation tests

tisto:
I figured out what's wrong. I am always running "make" which runs "bin/buildout". You folks run "bin/buildout -c plone-5.2.x.cfg".

.github/workflows/ci.yml
mauritsvanrees: > Can we try to fix things step by step? First the checkout, then the normalization?

No. If we only fix the checkout, we run into test failures for the part that checks if there are differences in the http-examples.

I have already fixed the documentation tests in PR https://github.com/plone/plone.restapi/pull/1215 which builds on the current PR.

Problem is that the follow-up PR changes 146 files, making it virtually impossible to review. That is why I split it into two PRs. But that is optional. You can also ignore the current PR and merge the other, as it includes all commits.

tisto:
@mauritsvanrees I merged https://github.com/plone/plone.restapi/pull/1215. Will close this one.

PR #1215, opened 2021-09-02
Normalize http examples

mauritsvanrees:
Note: this builds on my approved PR #1214 so that one should be merged first.

This normalises the requests and responses that are written to the `http-examples` directory:

- Use a stable behavior name for generated schemas. Fixes https://github.com/plone/plone.restapi/issues/1096 so this PR also upgrades us to the newer `plone.dexterity` from Plone 5.2.5.
- It strips away whitespace from the right of the line, leading to much smaller diffs - except for the current PR of course. Note that on GitHub you can add `?w=1` to the url to ignore whitespace changes, making this easier to review.
- It normalises the port number to the standard 55001. This helps when you run the restapi tests in a Plone core development buildout, which uses a random port.


mauritsvanrees:

tisto (approved):

src/plone/restapi/tests/http-examples/workingcopy_baseline_get.resp
```diff
@@ -2,83 +2,83 @@ HTTP/1.1 200 OK
...
+    "description": "",
+    "effective": null,
+    "exclude_from_nav": false,
+    "expires": null,
+    "id": "document",
+    "is_folderish": false,
+    "language": "",
+    "layout": "document_view",
+    "lock": {
+        "created": "1995-07-31T17:30:00",
```
wesleybl, line 43: Funny thing is that locally for me, only this line got different. She stayed:

```
        "created": "1995-07-31T14:30:00",
```

It seems that only in this part the TZ environment variable had no effect.

wesleybl: humm this is explained here:

https://github.com/plone/plone.restapi/pull/1215/commits/3a4e8f3a421accd18e5eb2927e441289bdced5bc

@mauritsvanrees wouldn't it be the case to normalize this too, so as not to show differences locally?

src/plone/restapi/tests/test_documentation.py
```diff
@@ -72,63 +72,61 @@
...
+# How do we open files?
+open_kw = {"newline": "\n"}
 
-def pretty_json(data):
-    return json.dumps(data, sort_keys=True, indent=4, separators=(",", ": "))
 
+def normalize_test_port(value):
+    # When you run these tests in the Plone core development buildout,
+    # the port number is random.  Normalize this to the default port.
+    return re.sub(r"localhost:\d{5}", "localhost:55001", value)
```
wesleybl, line 82: Now we can remove this line:

https://github.com/plone/plone.restapi/blob/c284b560d1a5cbb75a6bfb780c39fdfd139c4fb9/base.cfg#L74

src/plone/restapi/tests/test_documentation.py
```diff
@@ -515,15 +514,41 @@ def test_documentation_types_document_crud(self):
...
             },
         )
-        save_request_and_response_for_docs("types_document_post_field", response)
+        # With plone.dexterity 2.10+ we get an unstable behavior name like this:
+        # plone.dexterity.schema.generated.plone_5_1630611587_2_523689_0_Document
+        # Replace it.
+        document_schema_re = re.compile(
+            r"^plone.dexterity.schema.generated.plone_.*_Document$"
+        )
+        stable_behavior = "plone.dexterity.schema.generated.plone_0_Document"
```
wesleybl, line 523: I think stable_behavior should look like what it is in plone.dexterity 2.10+. That is:

plone.dexterity.schema.generated.plone_5_1630600102_2_680259_0_Document

src/plone/restapi/tests/http-examples/workingcopy_baseline_get.resp
mauritsvanrees: Possibly, but I wonder why the timezone has no effect here.
@avoinea This is the created date of a lock. Can you run the tests locally and then check if this results in a local change in git?

src/plone/restapi/tests/test_documentation.py
mauritsvanrees: We could remove that yes, if we want to.
@sneridagh You made commit https://github.com/plone/plone.restapi/commit/799e11e9c8d8df1040a7e60fe44ea40af028ce6a two years ago to always use port 55001. Was that to make the http-examples always have the same port? In that case we can now remove this setting, if we want.

src/plone/restapi/tests/test_documentation.py
mauritsvanrees: Technically that would be closer to what it looks like. But it would be uglier, so I prefer this clean one.

src/plone/restapi/tests/test_documentation.py
sneridagh: As long as the documentation `resp` files always get the same value and do not change over builds, it's ok.

src/plone/restapi/tests/test_documentation.py
wesleybl: Even being “uglier”, I prefer it closer to reality. Each part of this code has a meaning. When someone unsuspecting consults the documentation, they can get confused.

src/plone/restapi/tests/http-examples/workingcopy_baseline_get.resp
wesleybl: It's really weird that all dates respect the `TZ` environment variable and only this date doesn't.

I saw that who sets this date is the function:

https://github.com/plone/plone.restapi/blob/df3dba9b2eff44dc6349175ffdd9e97157ac58a9/src/plone/restapi/services/locking/__init__.py#L18-L19

Maybe this can help:

https://stackoverflow.com/questions/31977563/python-how-do-you-convert-a-datetime-timestamp-from-one-timezone-to-another-tim

@avoinea FYI

wesleybl: @avelino @mauritsvanrees 

@ericof fell in this hole too:

https://github.com/plone/plone.restapi/commit/2bc6910f3cf8bc856c6138d8e1828e7d87b4c765

Issue #1018, opened 2020-11-11
Tests on test_services_types.py fail randomly on Jenkins


wesleybl:
I noticed that the tests at [test_services_types.py](https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/tests/test_services_types.py) are failing with considerable frequency in Jenkins. Errors like:

Error message:
```
500 != 204
```

Traceback
```
500 != 204

  File "/srv/python2.7/lib/python2.7/unittest/case.py", line 329, in run
    testMethod()
  File "/home/jenkins/workspace/pull-request-5.2/src/plone.restapi/src/plone/restapi/tests/test_services_types.py", line 216, in test_types_document_patch_fieldsets
    self.assertEqual(response.status_code, 204)
  File "/srv/python2.7/lib/python2.7/unittest/case.py", line 513, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/srv/python2.7/lib/python2.7/unittest/case.py", line 506, in _baseAssertEqual
    raise self.failureException(msg)
```

For example:

https://jenkins.plone.org/job/pull-request-5.2/1712/testReport/junit/plone.restapi.tests.test_services_types/TestServicesTypes/test_types_document_get_field/

https://jenkins.plone.org/job/pull-request-5.2/1714/testReport/junit/plone.restapi.tests.test_services_types/TestServicesTypes/test_types_document_get_fieldset/

https://jenkins.plone.org/job/pull-request-5.2/1729/testReport/junit/plone.restapi.tests.test_services_types/TestServicesTypes/test_types_document_patch_fieldsets/

https://jenkins.plone.org/job/pull-request-5.2/1730/testReport/junit/plone.restapi.tests.test_services_types/TestServicesTypes/test_types_document_patch_create_missing/

https://jenkins.plone.org/job/pull-request-5.2/1736/testReport/junit/plone.restapi.tests.test_services_types/TestServicesTypes/test_types_document_get_field/

https://jenkins.plone.org/job/pull-request-5.2/1737/testReport/junit/plone.restapi.tests.test_services_types/TestServicesTypes/test_types_document_patch_create_missing/

These errors are likely due to high load on the Jenkins server. But why do only tests on `test_services_types.py` fail? Are type services heavier than others? Messages like `500 != 204` do not help to understand these errors. Perhaps it is good to print the body of the response when the return code is `500`, to see the backend traceback.

tisto:
[...]

PR #1218, opened 2021-09-06
Use checkout on 5.2, normalise http-examples [7xx]

mauritsvanrees:
This is a backport of my PR #1215 to branch 7.x.x.
It also updates 5.2.x to use Plone 5.2.5, including using the newer `plone.dexterity`. We normalise its unpredictable behavior schema names in the http examples.

This PR is best viewed with `?w=1` at the end of the url, to ignore whitespace changes, especially with the commit that updates the `http-examples`, which is actually a whitespace-only change.


tisto (approved):

mauritsvanrees:

avoinea (approved):

buildout.cfg
```diff
@@ -1,12 +1,2 @@
 [buildout]
 extends = plone-5.2.x.cfg
-versions=versions
-
-[versions]
-plone.restapi =
-
-plone.schema = 1.3.0
-plone.dexterity = 2.9.8
```
wesleybl, line 9: I found that issue #1018 occurs because of `plone.dexterity 2.9.10+`. More specifically after https://github.com/plone/plone.dexterity/pull/137

Strangely, it only occurs in Python 2.

The error occurred in Jenkins, but it did not occur in `plone.restapi` GA, because the `plone.restapi` buildout was pinned to `plone.dexterity 2.9.8`.

After the merge of this PR, the error can happen in GA as well.

wesleybl (commented):
I left some comments.

plone-5.2.x.cfg
```diff
@@ -11,19 +11,5 @@ black = 20.8b1
...
-# cryptography 3.4 requires a rust compiler installed on the system:
-# https://github.com/pyca/cryptography/blob/master/CHANGELOG.rst#34---2021-02-07
-cryptography = 3.3.2
-
-# cffi 1.14.3 fails on apple m1
-cffi = 1.14.4
-
-# requirement for json widget tests to pass
-plone.schema = 1.3.0
-plone.dexterity = 2.9.8
```
wesleybl, line 29: I found that issue #1018 occurs because of plone.dexterity 2.9.10+. More specifically after plone/plone.dexterity#137

Strangely, it only occurs in Python 2.

The error occurred in Jenkins, but it did not occur in plone.restapi GA, because the plone.restapi buildout was pinned to plone.dexterity 2.9.8.

After the merge of this PR, the error can happen in GA as well.

src/plone/restapi/tests/test_documentation.py
```diff
@@ -87,72 +87,64 @@
...
-            if content_type == "application/json":
-                json_body = json.loads(response.request.body)
-                body = pretty_json(json_body)
-                # Make sure Content-Length gets updated, just in case we
-                # ever decide to dump that header
-                response.request.prepare_body(data=body, files=None)
+def normalize_test_port(value):
+    # When you run these tests in the Plone core development buildout,
+    # the port number is random.  Normalize this to the default port.
+    return re.sub(r"localhost:\d{5}", "localhost:55001", value)
```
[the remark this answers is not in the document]
mauritsvanrees: I have now removed the hardcoded ZSERVER_PORT in the PR.

A good thing is, that this makes it possible to introduce tox and run it in parallel, if we want. (That is outside the scope of this PR though.)

src/plone/restapi/tests/test_documentation.py
```diff
@@ -545,15 +536,41 @@ def test_documentation_types_document_crud(self):
...
             },
         )
-        save_request_and_response_for_docs("types_document_post_field", response)
+        # With plone.dexterity 2.10+ we get an unstable behavior name like this:
+        # plone.dexterity.schema.generated.plone_5_1630611587_2_523689_0_Document
+        # Replace it.
+        document_schema_re = re.compile(
+            r"^plone.dexterity.schema.generated.plone_.*_Document$"
+        )
+        stable_behavior = "plone.dexterity.schema.generated.plone_0_Document"
```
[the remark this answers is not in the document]
mauritsvanrees: I have updated the PR to use a name closer to what newer dexterity gives us.

plone-5.2.x.cfg
mauritsvanrees: I guess with newer dexterity more test failures could indeed happen on GH actions. If so, this is good: it should bring us closer to fixing the underlying problem, whether that is in 'real' code or test code.

src/plone/restapi/tests/test_documentation.py
```diff
@@ -545,15 +536,45 @@ def test_documentation_types_document_crud(self):
...
         )
-        save_request_and_response_for_docs("types_document_post_field", response)
+        # With plone.dexterity 2.10+ we get an unstable behavior name like this:
+        # plone.dexterity.schema.generated.plone_5_1630611587_2_523689_0_Document
+        # In older versions, we got this:
+        # plone.dexterity.schema.generated.plone_0_Document
+        # Normalize this to look like the new name
+        document_schema_re = re.compile(
+            r"^plone.dexterity.schema.generated.plone_.*_Document$"
+        )
```
wesleybl, line 546: I think we can be less invasive here. That is, replacing only the really dynamic part. I suggest:

```suggestion
        document_schema_re = re.compile(
            r"^plone.dexterity.schema.generated.plone_5_\d*_2_\d*_0_Document$"
        )
```
So, if we have changes like what happened in `plone.dexterity 2.10+`, the tests will fail.

mauritsvanrees: Why would we do this?
I have no idea if the 5, 2 and 0 are always here or if they can differ as well. Seems pretty random to me.
This is an internal dexterity detail. If it changes, I do *not* want these tests to fail.

wesleybl: They will only change if we have structural changes like the ones that occurred not `plone.dexterity 2.10`. They don't change "via parameter". In this case, I find it valid for the test to fail, as failed. If we had this regular expression, before the change from `plone.dexterity 2.10`, the tests would not fail, even with the name pattern changing. And I think that in this case it should fail.

`schemaName` after this line:

https://github.com/plone/plone.dexterity/blob/35a12ab726b76ee5527ee467974917bc1fbdca0c/plone/dexterity/fti.py#L262

it's always something like:

`plone_5_1631640310_2_6213937_0_Document`

mauritsvanrees: I see the 5 and 0 come from the [`SchemaNameEncoder`](https://github.com/plone/plone.dexterity/blob/35a12ab726b76ee5527ee467974917bc1fbdca0c/plone/dexterity/schema.py#L295-L321).
I see no value in having a test in `plone.restapi` fail when this class changes.

wesleybl: We can have more similar values to the real ones in the documentation. I think that's why we have testing to generate documentation, rather than just writing the documentation.

mauritsvanrees: Fine, since no one stepped up to argue in my favour, I have changed it: commit 8921becee228c0fa2c43f31ed5d4731119a53bc7.

I think it is fine to merge now, right? Well, once everything is green.

mauritsvanrees:

wesleybl (approved):
It would be nice if the changes only made on this branch are also made on the master.

mauritsvanrees:
> It would be nice if the changes only made on this branch are also made on the master.

Will do.

mauritsvanrees:
Done: https://github.com/plone/plone.restapi/pull/1226

PR #1220, opened 2021-09-12
DX site root support for Plone 6 (backwards compat)

sneridagh:


sneridagh:

sneridagh:

tisto:

tisto:
@sneridagh this is a non-breaking change, correct? We can merge this as soon as we are certain that this works well with Plone 6. The only risk that I see is that we might have to change things in plone.restapi in the future (as long as Plone is in alpha phase) and then we might need a breaking release.

We could mark this feature as experimental and get around a possible breaking change I guess though.

sneridagh:
You are right, it's not a breaking change. +1 for marking it experimental, until he have the chance to test it properly.

No clue why Jenkins broke though.

sneridagh:

sneridagh:
@tisto it would be great to have this also soon!

I will do more tests later.

sneridagh:

src/plone/restapi/services/locking/configure.zcml
```diff
@@ -11,6 +11,14 @@
       name="@lock"
       />
 
+  <plone:service
```
tisto, line 14: @sneridagh how did the lock features made it in here?

src/plone/restapi/tests/test_serializer.py
```diff
@@ -395,10 +404,18 @@ def test_serialize_returns_site_root_common(self):
...
 
+    @unittest.skipIf(
+        PLONE_6,
+        "This test is only intended to run for Plone 6 and DX site root enabled",
+    )
     def test_serialize_returns_site_root_opt_in_blocks_not_present(self):
         self.assertEqual(self.serialize(self.portal)["blocks"], {})
         self.assertEqual(self.serialize(self.portal)["blocks_layout"], {})
 
+    @unittest.skipIf(
```
tisto, line 415: @sneridagh just wondering. Don't we need a test that checks the Plone 6 root serialization? Or are the basics covered in another test already?

tisto:
@sneridagh I had a first look and overall this looks good to me. Though, seems we have buildout issues. :/

src/plone/restapi/services/locking/configure.zcml
sneridagh: They are needed, otherwise Plone Site does not have lock support.

No clue why "Plone Site" content type does not implement `IContentish`, but that's another fight.

sneridagh:
@tisto @jensens @mauritsvanrees @pbauer @ericof without this, Plone 6 serialization/deserialization is broken. I marked it already as blocker. 

It would be great to have a unified way to deal with 5.2/6.0 environments. Is there already a good known one out there? 

I created https://github.com/plone/plone.restapi/pull/1313 to have an isolated branch with the buildout errors.

sneridagh:

src/plone/restapi/deserializer/site.py
```diff
@@ -10,68 +10,85 @@
 
 import json
 
+try:
+    from Products.CMFPlone.factory import PLONE60MARKER
+
+    PLONE60MARKER  # pyflakes
```
jensens, line 16: Would this work? An extra line to make pyflakes happy is so ugly...

```suggestion
    from Products.CMFPlone.factory import PLONE60MARKER  # flake8: noqa
```

ale-rt: I usually do:
```
from Products.CMFPlone.factory import PLONE60MARKER  # noqa: F401
```
but IMO trying to import stuff is not the correct way to get the Plone, I will come up with a suggestion

src/plone/restapi/deserializer/site.py
```diff
@@ -10,68 +10,85 @@
...
+except ImportError:
+    PLONE_6 = False
+else:
+    PLONE_6 = True
```
ale-rt, line 20: This is code already in `plone.restapi`:
```suggestion
from pkg_resources import get_distribution
from pkg_resources import parse_version


HAS_PLONE_6 = parse_version(
    get_distribution("Products.CMFPlone").version
) >= parse_version("6.0.0a1")
```

See https://github.com/plone/plone.restapi/search?q=HAS_PLONE_6

sneridagh: @ale-rt I know, that was before we had PLONE60MARKER in place (which I took care of).

Maybe we should change the old one?

src/plone/restapi/deserializer/site.py
sneridagh: I think I took it from somewhere else that was doing the same thing (with the PLONE52MARKER or so... I can change it indeed.

src/plone/restapi/deserializer/site.py
mauritsvanrees: For me, using the marker is good. I approved adding it. :-).
It is similar to the `have plone-6` condition that we have in zcml.

sneridagh: :) It seems you and @jensens wrote it. I pushed for a programmatic PLONE60MARKER because we were already using it in ZCML and we were using this way with `_IMREALLYPLONE5` in a number of tests already. I thought that it would be great to have, instead of dealing with parsing versions. Both are ugly enough and require a copy/paste (not easy recallable) tbh...

```
try:
    from Products.CMFPlone.factory import PLONE60MARKER  # noqa
except ImportError:
    PLONE_6 = False
else:
    PLONE_6 = True
```

ale-rt: For me using that marker like this is a bad practice, if you really want to use it I suggest to use something like:
```
from importlib import import_module

HAS_PLONE_6 = getattr(import_module("Products.CMFPlone.factory"), "PLONE60MARKER", False)
```
which does not pollute the locals...

sneridagh:

src/plone/restapi/deserializer/site.py
sneridagh: @ale-rt Looks good to me! Far less ugly :) Let me try!

sneridagh:
Green! How about now? :)

ericof (approved):
LGTM

tisto (approved):

PR #1221, opened 2021-09-13
@users endpoint should use acl_users.searchUsers.

### LINKED ISSUE #1199: /@users endpoint does not return the same results that the User control panel does
The current implementation of the User control panel uses PlonePAS to search for users (and list them) as [this code shows it](https://github.com/plone/Products.CMFPlone/blob/master/Products/CMFPlone/controlpanel/browser/usergroups_usersoverview.py#L49)

The endpoint for plone.restapi [has a different approach](https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/services/users/get.py#L38).

This distinct behaviour becomes more clear when you use a pas plugin (like pas.plugins.authomatic) and the users created there are not visible via restapi.

ericof:
Fixes #1199

### DIFF
--- a/news/1199.bugfix
+++ b/news/1199.bugfix
@@ -1,0 +1,2 @@
+Fix @users endpoint to use acl_users.searchResults instead of portal_membership.listMembers
+[ericof]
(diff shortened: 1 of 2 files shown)


ericof:

sneridagh (approved):

PR #1224, opened 2021-09-14
Rename ci.yml -> tests.yml. Enable zpretty check on gh actions

tisto:


base.cfg
```diff
@@ -195,6 +196,14 @@ mode = 755
 recipe = collective.recipe.omelette
 eggs = ${instance:eggs}
 
+[zpretty]
+recipe = collective.recipe.template
+input = inline:
+    #!/bin/bash
+    find src -name '*.zcml' | xargs bin/zpretty -i
```
jensens, line 203: For ZCML, argument `-z` - according to help output - "Follow the ZCML styleguide" - I think this is what we want? (cc @ale-rt)
```suggestion
    find src -name '*.zcml' | xargs bin/zpretty -iz
```

ale-rt: There is no need to pass the `-z` parameter if the file extension is already `.zcml`, see https://github.com/collective/zpretty/blob/6f69fc270dbe91f2fec4104e5d03a8317988010f/zpretty/cli.py#L55-L66

PR #1225, opened 2021-09-14
7.x.x Backport: fix(auth): Unify ZMI, HTML form, and API login

avoinea:
Fix some behavior in the PAS plugin that treated API login requests differently than
classic HTML form login requests, namely that the POST'ed login credentials weren't
extracted and authenticated like they are when the classic HTML login form is POST'ed.
IMO, the root cause here was a minor case of not actually fulfilling the PAS contract.
In the case of the classic HTML login form, it's another PAS plugin that extracts the
POST'ed credentials, but since the API login POST submits JSON, it has to be the JWT
plugin that extracts them.

Also reproduce the same handling that the classic Plone HTML login form does after login
has succeeded.  This includes setting the classic HTML login form cookie so that
authenticating to the API also authenticates the user to the classic HTML UI (as well as
the ZMI when the user has the `Manager` role).  But it also includes other core Plone
login behavior.

In the case of Volto, there's one more edge case remaining.  Namely, if the browser is
already authenticated such as by having previously submitted the classic HTML login
form, a fresh load of the Volto UI will still show them as logged out even though their
API requests succeed as authenticated.  I assume this is a problem in the React
component state that doesn't do an actual API check to see if the user is already
authenticated.


tisto:
@avoinea ok, this one needs to get merged as well, right? Can you look into the CI failures?

Issue #1231, opened 2021-09-21
When using @seearch, when a KeyError is raised in the ISerializeToJson serializer, previous result is appended twice


gbastien:
https://github.com/plone/plone.restapi/blob/6d3f27afceaf7355ae8152d9b28ece815945b7ef/src/plone/restapi/serializer/catalog.py#L55

This is because there would need a "continue" after the log.warning at line 55.

In fact the "except KeyError"  expect that the KeyError is made by the call to getObject() but if it is done in the serializer implementation this is somewhat hidden.

I would try/except just around the brain.getObject() if it is what we try to check then call the serializer outside the try/except to avoid hidding errors...

Does that make sense?

I will try to propose a test and a fix.

Gauthier

PS: actually getObject orphan brain will raise an AttributeError and not a KeyError, so I do not get the code...

gbastien:
Tested and fixed

PR #1233, opened 2021-09-24
Avoid `UnboundLocalError` or duplicates in results

gbastien:
when using `@search` endpoint and a brain is orphan or a `KeyError` occurs during result serialization

See issue #1231

PS: if this is backported to version working with Zope2, there will need to change as getObject an orphan brain raises KeyError in Zope4 but an AttributeError in Zope2...

### DIFF
--- a/news/1231.bugfix
+++ b/news/1231.bugfix
@@ -1,0 +1,2 @@
+Avoid `UnboundLocalError` or duplicates in results when using `@search` endpoint and a brain is orphan or a `KeyError` occurs during result serialization
+[gbastien]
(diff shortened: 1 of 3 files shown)


gbastien:
Hi @sneridagh @tisto 

when you have time to look at this?  We overrided in our code so that is not urgent, but I think this could happen to other people as well.

Thank you for great job and support!

Gauthier

tisto (approved):
@gbastien thanks for your fix and PR! LGTM! I'd like to get a second review, then I can merge and release it right away.

tisto:

PR #1237, opened 2021-09-29
fix the parameter name to get the current context item when using the @search endpoint

erral:
The documentation says to use `depth` but this does not work, the correct parameter is `path.depth`


erral:

tiberiuichim (approved):

PR #1241, opened 2021-10-08
Fix exclude context in querystringsearch

cekk:
Adding the "not UID" parameter to every call, really slow down catalog queries.
I've found that in a big website (~300000 contents) a @querystringsearch call with "not UID" parameter took 6-7 seconds instead ~3 without it.

I've also added a new allowed parameter in the endpoint (exclude_context) like the one in collection portlets, to allow excluding current context when needed (false by default).

I've also update the documentation: should i need to rebuild it or there are some hooks?

### DIFF
--- a/docs/source/querystringsearch.rst
+++ b/docs/source/querystringsearch.rst
@@ -21,4 +21,7 @@ The endpoint accepts the same parameters than the querystring search accept:
   - ``sort_order`` : ``"ascending"``, ``"descending"``
   - ``limit``
 
-and also the ``fullobjects`` parameter for return the whole objects instead of the summary serialization of them.
+and also some extra parameters:
+
+  - ``fullobjects`` (parameter for return the whole objects instead of the summary serialization of them)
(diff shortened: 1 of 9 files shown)


cekk:

jensens:
I would use repoze.profiler (set `profile=on` in buildout see https://github.com/plone/plone.recipe.zope2instance#development-options), prepare a slow query and run it with the profiler turned on. This should bring something interesting up.

jensens:
A `not` returns internally a really large `IISet` - means: an IISet containing the catalog ids of all items in the catalog except that one item. This IISet is then intersected with another IISet from another index (which one is defined by the query optimizer). 

My rough guess is, in a large site ZCatalog is here far from efficient. But this needs proof, for now it is a rough theory.

jensens:
Best would be to fix the catalog and not to work around here.

cekk:
I think that right now this pr could be closed: optimizations works well and listing blocks does not ask for fullobjects anymore, so queries are faster

PR #1240, opened 2021-10-07
Add missing backend logout actions for the @logout endpoint (delete c…

sneridagh:
…ookie, etc)

PR #1242, opened 2021-10-11
7.x Backport - Add missing backend logout actions for the @logout endpoint (delete c…

avoinea:
…… (#1240)

* Add missing backend logout actions for the @logout endpoint (delete cookie, etc)

* Fix the use case, removing the 501 branch, refactor old test


ichim-david:
@tisto  this backport isn't done yet, so if it's not too much trouble then yes you can make a new release now and we will see about another release later.

PR #1258, opened 2021-10-30
Lower restriction of vocabularies endpoint

ksuess:
- No check of field permissions: that's not the responsibility of the vocabulary, but of the schema.

### DIFF
--- a/src/plone/restapi/services/vocabularies/configure.zcml
+++ b/src/plone/restapi/services/vocabularies/configure.zcml
@@ -8,7 +8,7 @@
       accept="application/json"
       factory=".get.VocabulariesGet"
       for="Products.CMFPlone.interfaces.IPloneSiteRoot"
-      permission="plone.restapi.vocabularies"
+      permission="zope2.View"
       name="@vocabularies"
       />
 
@@ -17,7 +17,7 @@
       accept="application/json"
       factory=".get.VocabulariesGet"
       for="Products.CMFCore.interfaces.IContentish"
-      permission="plone.restapi.vocabularies"
+      permission="zope2.View"
       name="@vocabularies"
       />
 
--- a/src/plone/restapi/services/vocabularies/get.py
+++ b/src/plone/restapi/services/vocabularies/get.py
@@ -1,12 +1,8 @@
 from AccessControl import getSecurityManager
 
 
-try:
-    from plone.app.vocabularies.security import DEFAULT_PERMISSION
-    from plone.app.vocabularies.security import PERMISSIONS
-except ImportError:
-    from plone.app.content.browser.vocabulary import DEFAULT_PERMISSION
-    from plone.app.content.browser.vocabulary import PERMISSIONS
+from plone.app.content.browser.vocabulary import DEFAULT_PERMISSION
+from plone.app.content.browser.vocabulary import PERMISSIONS
 
 from plone.restapi.interfaces import ISerializeToJson
 from plone.restapi.services import Service
--- a/src/plone/restapi/services/vocabularies/get.py
+++ b/src/plone/restapi/services/vocabularies/get.py
@@ -2,13 +2,11 @@ from AccessControl import getSecurityManager
 
 
 try:
-    from plone.app.vocabularies import DEFAULT_PERMISSION
-    from plone.app.vocabularies import PERMISSIONS
+    from plone.app.vocabularies.security import DEFAULT_PERMISSION
+    from plone.app.vocabularies.security import PERMISSIONS
 except ImportError:
-    from plone.app.content.browser.vocabulary import (
-        DEFAULT_PERMISSION,
-        PERMISSIONS,
-    )
+    from plone.app.content.browser.vocabulary import DEFAULT_PERMISSION
+    from plone.app.content.browser.vocabulary import PERMISSIONS
 
 from plone.restapi.interfaces import ISerializeToJson
 from plone.restapi.services import Service
--- a/src/plone/restapi/services/vocabularies/get.py
+++ b/src/plone/restapi/services/vocabularies/get.py
@@ -35,20 +35,19 @@ class VocabulariesGet(Service):
         return {"error": {"type": type, "message": message}}
 
     def _has_permission_to_access_vocabulary(self, vocabulary_name):
-        """Check if user is authorized to access built-in vocabulary
+        """Check if user is authorized to access the vocabulary.
 
-        default permission for all vocabularies, built-in and others, was
-          <permission
-            id="plone.restapi.vocabularies"
-            title="plone.restapi: Access Plone vocabularies"
-            />
+        The endpoint using this method is supposed to have no further protection (`zope.2Public` permission).
+        A vocabulary with no further protection follows the `plone.app.vocabularies.DEFAULT_PERMISSION` (usually `zope2.View`).
+        For further protection the dictionary `plone.app.vocabularies.PERMISSION` is used.
+        It is a mapping from vocabulary name to permission.
+        If a vocabulary is mapped there, the permission from the map is taken.
+        Thus vocabularies can be protected stronger or weaker than the default.
         """
-        if vocabulary_name in PERMISSIONS:
-            sm = getSecurityManager()
-            return sm.checkPermission(
-                PERMISSIONS.get(vocabulary_name, DEFAULT_PERMISSION), self.context
-            )
-        return True
+        sm = getSecurityManager()
+        return sm.checkPermission(
+            PERMISSIONS.get(vocabulary_name, DEFAULT_PERMISSION), self.context
+        )
 
     def reply(self):
         # return list of all vocabularies
--- a/src/plone/restapi/services/vocabularies/get.py
+++ b/src/plone/restapi/services/vocabularies/get.py
@@ -1,16 +1,21 @@
 from AccessControl import getSecurityManager
+
+
 try:
-    from plone.app.vocabularies import DEFAULT_PERMISSION, DEFAULT_PERMISSION_SECURE, PERMISSIONS
+    from plone.app.vocabularies import DEFAULT_PERMISSION
+    from plone.app.vocabularies import PERMISSIONS
 except ImportError:
-    from plone.app.content.browser.vocabulary import DEFAULT_PERMISSION, DEFAULT_PERMISSION_SECURE, PERMISSIONS
-from plone.app.z3cform.interfaces import IFieldPermissionChecker
+    from plone.app.content.browser.vocabulary import (
+        DEFAULT_PERMISSION,
+        PERMISSIONS,
+    )
+
 from plone.restapi.interfaces import ISerializeToJson
 from plone.restapi.services import Service
 from zope.component import ComponentLookupError
 from zope.component import getMultiAdapter
 from zope.component import getUtilitiesFor
 from zope.component import getUtility
-from zope.component import queryAdapter
 from zope.interface import implementer
 from zope.publisher.interfaces import IPublishTraverse
 from zope.schema.interfaces import IVocabularyFactory
@@ -69,16 +74,14 @@ class VocabulariesGet(Service):
                 (
                     f"You are not authorized to access "
                     f"the vocabulary '{vocabulary_name}'."
-                )
+                ),
             )
 
         try:
             factory = getUtility(IVocabularyFactory, name=vocabulary_name)
         except ComponentLookupError:
             return self._error(
-                404,
-                "Not Found",
-                f"The vocabulary '{vocabulary_name}' does not exist"
+                404, "Not Found", f"The vocabulary '{vocabulary_name}' does not exist"
             )
 
         vocabulary = factory(self.context)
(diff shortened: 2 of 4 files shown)


ksuess:

ksuess:

src/plone/restapi/services/vocabularies/get.py
```diff
@@ -24,7 +36,24 @@ def _error(self, status, type, message):
...
+    def _has_permission_to_access_vocabulary(self, vocabulary_name):
+        """Check if user is authorized to access built-in vocabulary
+
+        default permission for all vocabularies, built-in and others, was
+          <permission
+            id="plone.restapi.vocabularies"
+            title="plone.restapi: Access Plone vocabularies"
+            />
+        """
+        if vocabulary_name in PERMISSIONS:
```
jensens, line 48: We need to always check the `DEFAULT_PERMISSION` in order to get the same defined behavior as here:
https://github.com/plone/plone.app.content/blob/7a1bd2b60136b1fd380d4fa04bd052498548f272/plone/app/content/browser/vocabulary.py#L304-L324

src/plone/restapi/services/vocabularies/get.py
```diff
@@ -1,3 +1,15 @@
...
+
+
+try:
+    from plone.app.vocabularies import DEFAULT_PERMISSION
+    from plone.app.vocabularies import PERMISSIONS
+except ImportError:
+    from plone.app.content.browser.vocabulary import (
+        DEFAULT_PERMISSION,
+        PERMISSIONS,
+    )
```
jensens, line 11: Do you plan to create a separate PR to move the default permissions to p.a.vocabularies? It's where I would expect those, not in the depth of p.a.content?

jensens (changes requested):

jensens:
I added comments - even if this one is a draft. I hope that's OK.

src/plone/restapi/services/vocabularies/get.py
ksuess: I skip the field permission check as this is not the responisbility of the vocabulary, but that of the schema and its consumer.

ksuess:
**Independent** related  PRs:
https://github.com/plone/plone.app.vocabularies/pull/67
https://github.com/plone/plone.app.content/pull/237

ksuess:

src/plone/restapi/services/vocabularies/get.py
ksuess: Independent related PRs:
plone/plone.app.vocabularies#67
plone/plone.app.content#237

ksuess:

src/plone/restapi/services/vocabularies/get.py
jensens: I agree here.

Anyway,  `if vocabulary_name in PERMISSION` is wrong/ superfluous. `DEFAULT_PERMISSION` must always be checked to match the logic in the above referenced current code.

src/plone/restapi/services/vocabularies/get.py
```diff
@@ -24,7 +34,24 @@ def _error(self, status, type, message):
         self.request.response.setStatus(status)
         return {"error": {"type": type, "message": message}}
 
+    def _has_permission_to_access_vocabulary(self, vocabulary_name):
+        """Check if user is authorized to access built-in vocabulary
```
jensens, line 38:
```suggestion
        """Check if user is authorized to access the vocabulary
```

src/plone/restapi/services/vocabularies/get.py
jensens: One could argue, if it needs check if it is different from the permission the view is registered with. 

Thinking further, the endpoint can be registered with `zope.Public` in ZCML. The default goes with `zope.View` as is. Just, if anyone in an add-on needs it, a vocabulary could be registered with `zope.Public`. This way we do not disallow this (probably rare) use case by design.

src/plone/restapi/services/vocabularies/configure.zcml
```diff
@@ -17,7 +17,7 @@
       accept="application/json"
       factory=".get.VocabulariesGet"
       for="Products.CMFCore.interfaces.IContentish"
-      permission="plone.restapi.vocabularies"
+      permission="zope2.View"
```
jensens, line 20: See below.
```suggestion
      permission="zope2.Public"
```

src/plone/restapi/services/vocabularies/configure.zcml
```diff
@@ -8,7 +8,7 @@
       accept="application/json"
       factory=".get.VocabulariesGet"
       for="Products.CMFPlone.interfaces.IPloneSiteRoot"
-      permission="plone.restapi.vocabularies"
+      permission="zope2.View"
```
jensens, line 11: See below.
```suggestion
      permission="zope2.Public"
```

jensens:
The permission here is now no longer needed, right?
https://github.com/plone/plone.restapi/blob/94876588ea3437322d8f9c265d96fd86fe42fb30/src/plone/restapi/permissions.zcml#L11-L14

jensens:
Forget my above comment, the `@sources` endpoint uses it. TBH, I guess It needs some love too.

src/plone/restapi/services/vocabularies/get.py
tisto: @jensens @ksuess if we move the permissions to p.a.vocabularies and plone.app.content, we introduce a tight coupling to those packages, right? This means we would have to backport those enhancements to Plone 5.2 compatible versions and add lots of version pins for the different Plone versions plone.restapi supports (Plone 5.2 and Plone 6). In the past (e.g. for a simple enhancement in plone.schema) this approach caused lots of trouble and confusion. Even in the best possible scenario, we might end up with a long list of KGS version pins to make Plone 5.2 work with plone.restapi. I don't have a silver bullet solution here but maybe it is better to defer those moves until Plone 6 is stable and we can start moving things around in Plone 7 anyway.

jensens: plone.app.content and plone.app.vocabularies are anyway coupled. I tend to move those basic things as an (empty) permission map for vocabularies to a plone.base package at some point and fill it with values where vocabularies are created. But this is for future. For now plone.app.content and plone.app.vocabularies are anyway coupled over CMFPlone in both directions.

I would not open that box here in plone.restapi discussion and go with pragmatic small steps.

I do not think, if done right, we run into trouble in a 5.2 usage. So, IMO @ksuess did it right by importing from the old place in plone.restapi, so no problems here. We may want to add a comment at least why the alternative import is here and when we can remove it.

Regarding the move itself, pa.vocabularies would need a breaking change/major version bump and buildout.coredev 5.2 needs to use the pa.vocabularies 4.x branch then while master is 5.

src/plone/restapi/services/vocabularies/get.py
ksuess: I understand that DEFAULT_PERMISSION needs to be checked if the same behavior as VocabularyView.get_vocabulary of https://github.com/plone/plone.app.content/blob/7a1bd2b60136b1fd380d4fa04bd052498548f272/plone/app/content/browser/vocabulary.py#L304-L324 should be followed in vocabulary endpoint.
It's not in scope of this PR to think about why a vocabulary is a utility which cannot be registered with a permission. Which means that the DEFAULT_PERMISSION is supposed to fit for custom vocabularies.

I will take just the one step to follow the classic behavior in vocabulary endpoint and check DEFAULT_PERMISSION for all vocabularies.

src/plone/restapi/services/vocabularies/configure.zcml
jensens: keep as is - lets follow the old way at https://github.com/plone/plone.app.content/blob/2b5445f9c34c5a897471a3fe2823d12527c6f95a/plone/app/content/browser/configure.zcml#L166

ksuess:

jensens (approved):

sneridagh:
@jensens So we have a messy situation here. This is what happened:

Yesterday in the same moment I was cutting a release of p.restapi, you merged this PR. So, when I pushed the result, then it conflicted, then the history got messed up. So to solve it without rewriting the history, then I re-released. So we have 8.15.1 and 8.15.2.

I tested it then in our project, and it's throwing errors in some tests that we have in custom vocabularies.

So questions:

1. Is this PR a breaking change?
2. Could we drop a minimum amount of documentation about how they work now?
3. I see from the comments that this relies in other PRs in other packages, that are still not merged, let alone released. 
4. Do the latter change the local dependencies of p.restapi, and if so, should we pin some minimum versions?

If after answering this, we conclude it's a brown bag release, we should revert the changes, then re-release again.

Please I need a swift answer on this one.

ksuess:
> So questions:
> 
>     1. Is this PR a breaking change?

It's a bug fix.
IMHO it's not a breaking change, as it 
- decreases the protection
- but does not open a security hole (DEFAULT_PERMISSION `zope2.view` is checked)

The former endpoints permission="plone.restapi.vocabularies" with 
```
  <permission name="plone.restapi: Access Plone vocabularies" acquire="True">
      <role name="Manager"/>
      <role name="Site Administrator"/>
    </permission>
```
led to missing options in select fields for Editors, e.a.. This is patched in multiple forms out there.
> 
>     2. Could we drop a minimum amount of documentation about how they work now?

No patch for users without 'Manager' and 'Site Administrator' role needed anymore.

'_has_permission_to_access_vocabulary' got a sum up of what it does in docstring
https://github.com/plone/plone.restapi/blob/8f1d624727901224ff3eef122cb404221550d492/src/plone/restapi/services/vocabularies/get.py#L33-L42
As this is a bug fix, I do not see the urge to add an explanation to an upgrade guide.

> 
>     3. I see from the comments that this relies in other PRs in other packages, that are still not merged, let alone released.

This PR is independent.
It does not rely on the mentioned PRs on moving permissions for the  built-in vocabularies
    PERMISSIONS = {
}
        "plone.app.vocabularies.Catalog": "View",
        "plone.app.vocabularies.Keywords": "Modify portal content",
        "plone.app.vocabularies.SyndicatableFeedItems": "Modify portal content",
        "plone.app.vocabularies.Users": "Modify portal content",
        "plone.app.multilingual.RootCatalog": "View",
    }
We even removed the alternative import (p.a.vocabulary instead of p.a.content) and postponed it to a following PR that is explicitly addressed to the move of built-in vocabulary permissions.
> 
>     4. Do the latter change the local dependencies of p.restapi, and if so, should we pin some minimum versions?

No, there is no dependency.

jensens:
>    Is this PR a breaking change?

No. It is a bufix.

>    Could we drop a minimum amount of documentation about how they work now?

At the moment there is not any word about access permission of the endpoints here https://plonerestapi.readthedocs.io/en/latest/vocabularies.html 
As a  bugfix, vocabularies are working now the same way as they used to before plone.restapi emerged.
But overall we should document the PERMISSION global of (currently) plone.app.content (see below), which is valid for restapi and classic..

    I see from the comments that this relies in other PRs in other packages, that are still not merged, let alone released.
    Do the latter change the local dependencies of p.restapi, and if so, should we pin some minimum versions?

No, in a call @ksuess and I decided to not cover the changes over there in this PR. It would enhance the scope too much. So there is no dependency to the mentioned PRs anymore in the merged code.

sneridagh:
Hey @ksuess @jensens I have to protect one vocabulary, how do you protect just one in a sane way?

patch the p.a.content doesn't seem an option to me. How about if your code and several add-ons want to do the same?

jensens:
Patching is the way it was in the `getVocabulary` endpoint of `plone.app.content` all the time. Its not perfect, but that's how it works and now concequently rstapi follow the same permissions. I am open for ideas how to improve here. One idea I have is to facilitate the existing `permission` attribute of the `<utility ../> directive if possible (`overides.zcml` for overriding) or another one to create an own directive `<vocabulary ../>` with permissions. 

Given your case of patching it now: I would not expect several add-ons to patch a vocabulary permission. In fact, in past it was rarely done.

```
from plone.app.content.browser.vocabulary import PERMISSIONS

# patch an existing
PERMISSIONS["plone.app.vocabularies.Users"] = "Manage Users"

# add custom
PERMISSIONS["myproject.productcategeories"] = "Myproject: Mypermission"
```

cekk:
I really like this implementation!

Just one thing: these changes basically are exposing all vocabularies to everyone that can View the context where the endpoint is called.

@vocabularies will expose all the available vocabularies to anonymous users and then when we want to get one of them, there is the permission check (now with the mapping with PERMISSIONS variable and later with the #1287).

Is it ok to expose to everyone some vocabularies like the ones for permissions, roles, groups, etc. ? Only Users are protected somehow because are defined in PERMISSIONS mapping.

cekk:
you're right but i didn't care on old plone versions because you needed to know the view and the vocabulary name. With restapi, the endpoint returns all the infos and a list of vocabularies.

Where is the right place to open that PLIP? Here or in a different repo? For me this could be only a restapi-related problem

jensens:
PLIPs are always oig to Products.CMFPlone - theres a template for this.

PR #1259, opened 2021-10-31
Fix GHA test action to just run tests on Plone 5.2. Plone 6 CI is bes…

ale-rt:
…t done in Jenkins. Update to the tested Plone version to 5.2.6

Refs. #1255

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,4 +1,4 @@
-name: Tests
+name: Check Plone 5.2 compatibility, code analysis and documentation
 on: [push]
 jobs:
   build:
@@ -7,7 +7,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [3.8, 3.7, 3.6]
-        plone-version: ["6.0", "5.2"]
+        plone-version: ["5.2"]
 
     steps:
       # git checkout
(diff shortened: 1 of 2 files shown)


ale-rt:
@tisto I hope this PR makes my point more clear than what I wrote in https://github.com/plone/plone.restapi/pull/1255#issuecomment-955715001

Disclaimer I might have not have a full vision of the big picture but I am pretty convinced I nailed it.
If I am wrong, please be patient and explain!

.github/workflows/tests.yml
```diff
@@ -7,7 +7,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [3.8, 3.7, 3.6]
-        plone-version: ["6.0", "5.2"]
```
tisto, line 10: @ale-rt as outlined in my comment I'd prefer to keep things as they are in tests.yml. The rest of the PR is fine. :)

tisto (changes requested):

.github/workflows/tests.yml
ale-rt: I know what you outlined, but I do not understand it :)
jenkins tests are already doing everything is needed and testing Plone 6.0 here is just adding complications.
In addition the tested Plone 6 version is already old, see:
- https://github.com/plone/plone.restapi/blob/52b2bcfbffa4559216843c37a19d088fd0e722f8/plone-6.0.x.cfg#L4

I am still convinced this PR is doing the right thing!

ale-rt:
See the discussion that happened in #1255

PR #1261, opened 2021-11-01
Implement IJSONSummarySerializerMetadata

ericof:
This allows addons to extend the metadata returned by the Summary serializer

To always return **image_field** metadata on summaries, in an addon you will implement a new file called `summary,py` with content
```
from plone.restapi.interfaces import IJSONSummarySerializerMetadata
from zope.interface import implementer


@implementer(IJSONSummarySerializerMetadata)
class JSONSummarySerializerMetadata:
    def default_metadata_fields(self):
        return {
            "image_field",
        }

```

and register it with:
```
<utility factory=".summary.JSONSummarySerializerMetadata" name="plone.volto.summary_serializer_metadata" />
```

### DIFF
--- a/news/1250.feature
+++ b/news/1250.feature
@@ -1,2 +1,2 @@
-Return image_field metadata on SummarySerializer for brains
+Implement IJSONSummarySerializerMetadata allowing addons to extend the metadata returned by Summary serializer.
 [ericof]
--- a/news/1250.feature
+++ b/news/1250.feature
@@ -1,0 +1,2 @@
+Return image_field metadata on SummarySerializer for brains
+[ericof]
--- a/src/plone/restapi/interfaces.py
+++ b/src/plone/restapi/interfaces.py
@@ -208,3 +208,19 @@ class IBlockSearchableText(Interface):
 
     def __call__(value):
         """Extract text from the block value. Returns text"""
+
+
+class IJSONSummarySerializerMetadata(Interface):
+    """Configure JSONSummary serializer."""
+
+    def default_metadata():
+        """Returns a set with default metadata to be serialized."""
+
+    def field_accessors():
+        """Returns a dictionary with field accessors to be used during serialization."""
+
+    def non_metadata_attributes():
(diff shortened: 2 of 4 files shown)


tiberiuichim:
A Plone addon for an image gallery (as an example), wouldn't know which fields it should return in default_metadata_fields. We either need named utilities (so that we can use `getUtilitiesFor()` ) with one such utility per field, or use a subscriber system (and let them write into a list). 

If we go named utility route, we can also do some plone.directive integration, so that we can mark a field as "included in json summary".

ericof:

ericof:
> A Plone addon for an image gallery (as an example), wouldn't know which fields it should return in default_metadata_fields. We either need named utilities (so that we can use `getUtilitiesFor()` ) with one such utility per field, or use a subscriber system (and let them write into a list).

To be honest I was mostly concerned with the use case of a policy package controlling everything (instead of having each package register its own metadata).

@sneridagh, @tisto, opinions?

tiberiuichim:
@erico sorry, I misread your comments and didn't look at your implementation. Good to go, from my part

tiberiuichim (approved):

ericof:
@tiberiuichim, @sneridagh We need to worry about the errors on Jenkins. It looks like @ale-rt merged his "Plone site root on the catalog", and this will start breaking tests here

tisto (approved):
LGTM. I'd love to have a short section in the docs and maybe a short how-to about how to override this. Maybe @ThomasKindermann @steffenri @iFlameing @jackahl can work on this with a few pointers.

ericof:
I will merge this, then open a new PR for the documentation
Also, we need to decide how to deal with the Plone 6 master tests failing

ale-rt:
I saw that coming... 
This was the reason for https://github.com/plone/plone.restapi/pull/1259 but after a long discussion with Timo it was decided to close it.
I still think it was a bad idea not to merge it.

See also https://github.com/plone/plone.restapi/pull/1251/files on how to adapt tests for Plone 6.

PR #1267, opened 2021-11-20
Remove all traces of Products.CMFQuickInstaller

jensens:
For some reason I got an `Products.CMFQuickInstaller` import error in Plone 6. Maybe because its gone? Question is why it worked in past? Did it?

Anyway, this fixes the problem.

### DIFF
--- a/src/plone/restapi/services/addons/addons.py
+++ b/src/plone/restapi/services/addons/addons.py
@@ -2,9 +2,6 @@ from plone.memoize import view
 from Products.CMFCore.utils import getToolByName
 from Products.CMFPlone import PloneMessageFactory as _
 from Products.CMFPlone.interfaces import INonInstallable
-from Products.CMFQuickInstallerTool.interfaces import (
-    INonInstallable as QINonInstallable,
-)
 from Products.GenericSetup import EXTENSION
 from Products.GenericSetup.tool import UNKNOWN
 from zope.component import getAllUtilitiesRegisteredFor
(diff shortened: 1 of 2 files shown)


jensens:

src/plone/restapi/services/addons/addons.py
```diff
@@ -2,16 +2,22 @@
 from Products.CMFCore.utils import getToolByName
 from Products.CMFPlone import PloneMessageFactory as _
 from Products.CMFPlone.interfaces import INonInstallable
-from Products.CMFQuickInstallerTool.interfaces import (
```
ale-rt, line 5: Uhm, this is strange, the iface was moved to:
```
from Products.CMFPlone.interfaces import INonInstallable
```
And we should have aliases for that:

https://github.com/plone/plone.app.upgrade/blob/ef4aef834c561f7a634e819571b7c3bfb48cf314/plone/app/upgrade/__init__.py#L185

Don't you have plone.app.upgrade?

I think the proper fix is to replace the import given that the alias is there since 5.1:
https://github.com/plone/Products.CMFPlone/blob/5.1.x/Products/CMFPlone/interfaces/__init__.py#L38
and this package supports Plone 5.2 only.

Also the towncrier entry should be updated accordingly

ale-rt (changes requested):

src/plone/restapi/services/addons/addons.py
jensens: No I do only depend on `Products.CMFPlone`. Which is totally fine. `plone.app.upgrade` in only installed in a single service instance. The cluster runs usually without.

jensens: > I think the proper fix is to replace the import given that the alias is there since 5.1:
Right, this the current master does not support 5.1 anyway. Removal is probably the best.

jensens:

ale-rt (approved):

PR #1269, opened 2021-11-21
fix: pas plugin was missing after install

jensens:
For some reason i did not get the JWT plugin installed with the default profile. While looking at it I modernized the GS calls and now it works.


jensens:

jensens:

mauritsvanrees (approved):

PR #1276, opened 2021-11-24
Remove sleeps in test_services_types in Python 2

wesleybl:
`plone.dexterity` 2.10+ uses the `_p_mtime` variable in the schema name. These sleeps were added because I thought `_p_mtime` was more precise in Python 3, but it was actually the `str` function, used to convert `_p_mtime`
from float to string, that was rounding off `_p_mtime`. Now `plone.dexterity` uses the `repr` function, which doesn't round float,
instead of `str`.

See: https://github.com/plone/plone.dexterity/pull/159

This revert: https://github.com/plone/plone.restapi/commit/39e4f20e41acad9fcd25cf850a42bbcdd47b7cd2


wesleybl:

mauritsvanrees (approved):

Issue #1273, opened 2021-11-24
Drop support for Plone 5.0 on 7.x.x


wesleybl:
In the 7.x.x setup.py, we still have:

https://github.com/plone/plone.restapi/blob/244cea5d4e545f48139d61c49bbd38efabb9360b/setup.py#L61

Tests do not run on Plone 5.0:

https://github.com/plone/plone.restapi/blob/2c859f649ead38f2d64bcf69c4cf3320fa9b4388/.github/workflows/ci.yml#L10

So we should remove that line from setup.py, right?

PR #1280, opened 2021-11-25
Drop support to Plone 5.0

wesleybl:
CI tests do not run with Plone 5.0: https://github.com/plone/plone.restapi/blob/2c859f649ead38f2d64bcf69c4cf3320fa9b4388/.github/workflows/ci.yml#L10

Locally Plone 5.0 tests are broken.

Fixes: #1273


tisto:
@wesleybl crap. Thanks for pointing that out. I was under the assumption that Plone 5 has been removed from the classifiers. We indeed claim to support that version even though we actually don't (might work still). We can not introduce a breaking change into the 7.x.x. This would lead to a new major version which is 8 (and already occupied).

tisto:
@wesleybl just checked. This is only an issue of the 7.x.x branch:

https://github.com/plone/plone.restapi/blob/master/setup.py#L57

Then we have to make a feature release within the 7.x.x. I don't see any other way. Plone 5.0 is out of security releases already so I guess we can live with that.

wesleybl:
> Then we have to make a feature release within the 7.x.x

This PR is for the 7.x.x branch. Should I change the name of the news/1273.breaking file to news/1273.feature?

tisto:
@wesleybl I will take care of the changelog. Sorry it took so long to merge your PR.

Issue #1282, opened 2021-11-28
Improve support and meaning for `default` and `missing_value` in serializers/deserializers


sneridagh:
While trying to improve the select widget story in Volto, I found out one issue. Since Volto only sends "changed" field data in forms, `missing_value` are not being set correctly (since the part of the code that sets it, assumes that the field is present in the sent data (as null)), and validation fails (especially for array-ish fields). This provokes to "force" you to use `default` in them, which  is wrong. It's wrong, since `default` serves for another purpose.

Also, the combined use of `default` and `missing_value` is also kind of confusing, or I am missing something? Might be opening a Pandora's box here but... won't be more pragmatic to force internally a "missing_value" given the type of the field being used? I'm sure there's a reason why...

More background:

https://github.com/collective/example.contenttype/blob/volto/src/example/contenttype/example.py#L203

```
    list_field_checkbox = schema.List(
        title=u'List field with checkboxes',
        description=u"zope.schema.List",
        value_type=schema.Choice(
            values=[u'Beginner', u'Advanced', u'Professional'],
            ),
        required=False,
        missing_value=[],
        )
```

A list field declared like this, does not work on Volto now. On save, if the field value is untouched, it breaks.

/cc @tisto @ericof @tiberiuichim

tiberiuichim:
As long as we're using z3c.form we have to abide to its API. I think your solution, to adopt the incoming value to the API of z3c.form is the proper one, given what we have.

PR #1283, opened 2021-11-28
Improve support for `missing_value` and `default` story

sneridagh:
See #1282 .

Could be a conflictive piece of code, so please triple check this. Tricky thing here is that I guess that `missing_value` is a `zc3.form`/Classic UI thing, and we are transferring those responsibilities to the RESTAPI code here. One could argue that the frontend should take care of that instead than the API.

### DIFF
--- a/src/plone/restapi/deserializer/dxcontent.py
+++ b/src/plone/restapi/deserializer/dxcontent.py
@@ -159,14 +159,17 @@ class DeserializeFromJson(OrderingMixin):
                     if name == "changeNote":
                         continue
                     dm = queryMultiAdapter((self.context, field), IDataManager)
-                    bound = field.bind(self.context)
-                    try:
-                        if name not in data and not dm.get() and not field.required:
+
+                    # Covering here missing_value/default edge cases
+                    if name not in data:
+                        dm = queryMultiAdapter((self.context, field), IDataManager)
+
+                        if field.missing_value and not dm.get() and not field.required:
                             # If the field is not set in data, it has no value set either,
                             # and missing_value is defined, it sets the missing value
                             dm.set(field.missing_value)
                         if (
-                            name not in data
+                            field.missing_value
                             and dm.get() == field.default
                             and not field.required
                         ):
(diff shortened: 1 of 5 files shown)


tiberiuichim (approved):

sneridagh:

jensens:
missing_value is a `zope.schema` thing and not `z3c.form` related:
https://zopeschema.readthedocs.io/en/latest/api.html#zope.schema.interfaces.IField.missing_value

src/plone/restapi/deserializer/dxcontent.py
```diff
@@ -159,6 +159,29 @@ def get_schema_data(self, data, validate_all, create=False):
...
+                        if (
+                            field.missing_value is not None
+                            and dm.get() == field.default
+                            and not field.required
+                        ):
+                            # If the field is not set in data, it has a default,
+                            # and missing_value is defined, it sets the missing value
+                            # (so it prevails instead of default)
+                            dm.set(field.missing_value)
+
```
jensens, line 184: Sorry, I think the code/condition logic is unnecessary complex and difficult to read.
`dm` is already define some lines before.
Since this mght be called often (?) a dm.get() before save some cycles.
What do you think about this:

```suggestion
                    # Covering here missing_value/default edge cases
                    if (
                        name not in data 
                        and field.missing_value is not None
                        and not field.required
                    ):
                        # precondition:
                        # - name was not handled before 
                        # - not required, otherwise there has to be a value
                        # - missing_value is defined, so it can be set
                        dm_value = dm.get()
                        if not dm_value or dm_value == field.default:
                            # If theres no value at all
                            # or if value is the default 
                            # (so it prevails instead of default)
                            # then it sets the missing value
                            dm.set(field.missing_value)

```

jensens (changes requested):

sneridagh:

sneridagh:

jensens (approved):

tisto (approved):

Issue #1286, opened 2021-11-30
Add check for protected vocabularies in endpoint 'vocabularies'.


ksuess:
Then a vocabulary can be registered with permission:

```
  <utility
      provides="zope.schema.interfaces.IVocabularyFactory"
      name="plone.restapi.testing.protected_vocabulary"
      component=".tests.test_services_vocabularies.test_vocabulary_factory"
      permission="cmf.ManagePortal"
      />
```

PR #1287, opened 2021-11-30
Add check for protected vocabulary in endpoint 'vocabularies'.

ksuess:
Fix [1286](https://github.com/plone/plone.restapi/issues/1286)

### DIFF
--- a/news/1286.feature
+++ b/news/1286.feature
@@ -1,0 +1,1 @@
+Add check for protected vocabulary ind endpoint 'vocabularies' [ksuess]
--- a/src/plone/restapi/services/vocabularies/get.py
+++ b/src/plone/restapi/services/vocabularies/get.py
@@ -92,7 +92,7 @@ class VocabulariesGet(Service):
 
         factory = ProxyFactory(factory)
         checker = getChecker(factory)
-        permission = dict(checker.get_permissions.items()).get('__call__')
+        permission = dict(checker.get_permissions.items()).get("__call__")
         if permission and not permission == CheckerPublic:
             sm = getSecurityManager()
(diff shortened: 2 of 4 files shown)


ksuess:

jensens:
@ksuess Genius! That's how it should work. I love it. 

I would just merge this!

Then we can go further:

- We need the same over in `plone.app.content.browser.vocabulary` 
- then protect all the plone.app.vocabularies utility registrations with permissions
- import the PERMISSION_MAP conditional for restapi (to not break older Plones), see below
- drop PERMISSION_MAP in Plone 6 at all, including its permission check.
- document it (which is just pointing to utility basically).

Finally a clean solution.

jensens (approved):

sneridagh (approved):
LGTM! Awesome! 

Sorry for being picky, I checked quickly and in the cell phone, but shouldn't we also take care of the 401 use case from a formal point of view?

ksuess:
May I propose to close this PR. 
A registration like below can fool developers into believe that this vocabulary is protected.
But the vocabulary is protected by requesting via REST API endpoint, but not in general.

```
  <utility
      provides="zope.schema.interfaces.IVocabularyFactory"
      name="plone.restapi.testing.protected_vocabulary"
      component=".tests.test_services_vocabularies.test_vocabulary_factory"
      permission="cmf.ManagePortal"
      />
```
      
Thanks for the kind feedback and sorry for the noise.

ksuess:
In midterm we can maybe think of a Plone Vocabulary Registry that can deal with permission protection.

A first try: https://github.com/plone/plone.app.vocabularies/pull/69

PR #1290, opened 2021-12-04
Revisit improve support for missing value if present

sneridagh:


sneridagh:
Proof that we are not handling defaults ^^^

jensens:
Ok, I drop here the the definition of both first - I need to wrap my head around the whole...
https://zopeschema.readthedocs.io/en/latest/narr.html#data-modeling-concepts

**Whether or not the field is required**    
If a field is designated as required, assigned field values must always be non-missing. See the next section for a description of missing values.

**A value designated as missing** 
Missing values, when assigned to an object, indicate that there is ‘no data’ for that field. Missing values are analogous to null values in relational databases. For example, a boolean value can be True, False, or missing, in which case its value is unknown.

While Python’s None is the most likely value to signify ‘missing’, some fields may use different values. For example, it is common for text fields to use the empty string (‘’) to signify that a value is missing. Numeric fields may use 0 or -1 instead of None as their missing value.

A field that is ‘required’ signifies that missing values are invalid and should not be assigned

**A default value**
Default field values are assigned to objects when they are first created. A default factory can be specified to dynamically compute default values.

tiberiuichim:
missing = "I'm empty but here's my value". 

I can understand the case where you're creating an object, it comes with a default value, but then the user decides that the default value is wrong and instead "empty" should be the real value. But what is actually the use case for missing? It seems to me that "missing = default" would seem like the sane value all the time.

PR #1291, opened 2021-12-06
Do not break if publish recursive folder children that are already published

cekk:
The problem is that if we have a private folder (A) with some children (AA AB AC) and one of these is already published (AB), the endpoint call returns a 400 because AB can't apply "publish" transition. The publication process will fail with some folders published (the ones before the published one) and other not (the ones after).

With this pr i will check if the current obj state is the same of the transition's destination and skip the exception. Otherwise, raise the error.

### DIFF
--- a/news/1291.bugfix
+++ b/news/1291.bugfix
@@ -1,0 +1,1 @@
+Do not break in recursive transition when children already are in destination state. [cekk]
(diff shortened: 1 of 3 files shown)


cekk:

tisto (approved):
@cekk good catch! It would be interesting to have a performance test for a very large folder that is published (just sayin).

cekk:
@tisto I don't remember how do you create performance tests and if it's something that everyone can see.

cekk:

tisto:
@cekk I started to write new tests with locust:

https://github.com/plone/plone.restapi/blob/master/performance/querystring-search.py

So feel free to add more. :)

We would have to set them up to run regularly on our CI infrastructure (so far I run them on our company Jenkins).

PR #1293, opened 2021-12-07
Keep order of blocks when indexing SearchableText

### LINKED ISSUE #1292: SearchableText indexer should preserve order of blocks

ericof:
Closes #1292

### DIFF
--- a/src/plone/restapi/indexers.py
+++ b/src/plone/restapi/indexers.py
@@ -16,14 +16,6 @@ from zope.interface import implementer
 from zope.publisher.interfaces.browser import IBrowserRequest
 
 
-def _extract_text(block):
-    result = ""
-    for paragraph in block.get("text", {}).get("blocks", {}):
-        text = paragraph["text"]
-        result = " ".join((result, text))
-    return result
-
-
 @implementer(IBlockSearchableText)
 @adapter(IBlocks, IBrowserRequest)
 class TextBlockSearchableText:
@@ -34,7 +26,11 @@ class TextBlockSearchableText:
         self.request = request
 
     def __call__(self, value):
-        return _extract_text(value)
+        result = ""
+        for paragraph in value.get("text", {}).get("blocks", {}):
+            text = paragraph["text"]
+            result = " ".join((result, text))
+        return result
 
 
 @implementer(IBlockSearchableText)
--- a/src/plone/restapi/indexers.py
+++ b/src/plone/restapi/indexers.py
@@ -104,7 +104,6 @@ def extract_text(block, obj, request):
     """
     result = ""
     block_type = block.get("@type", "")
-    subblocks = extract_subblocks(block)
     # searchableText is the conventional way of storing
     # searchable info in a block
     searchableText = block.get("searchableText", "")
@@ -115,7 +114,8 @@ def extract_text(block, obj, request):
     # Use server side adapters to extract the text data
     adapter = queryMultiAdapter((obj, request), IBlockSearchableText, name=block_type)
     result = adapter(block) if adapter is not None else ""
-    if not result and subblocks:
+    if not result:
+        subblocks = extract_subblocks(block)
         for subblock in subblocks:
             tmp_result = extract_text(subblock, obj, request)
             result = f"{result}\n{tmp_result}"
--- a/src/plone/restapi/indexers.py
+++ b/src/plone/restapi/indexers.py
@@ -2,7 +2,7 @@
 # This is an experimental feature meant for use in Volto only!
 # This code is likely to change in the future, even within minor releases.
 # We will make sure plone.restapi latest always works with the latest Volto release.
-# This code is planned to being refactored into CMFPlone 6.0 as soon as Volto 4 final is out.
+# This code is planned to being refactored into plone.volto before CMFPlone 6.0 is out.
 # <tisto@plone.org>
 
 from plone.app.contenttypes.indexers import SearchableText
@@ -27,6 +27,8 @@ def _extract_text(block):
 @implementer(IBlockSearchableText)
 @adapter(IBlocks, IBrowserRequest)
 class TextBlockSearchableText:
+    """Searchable Text indexer for Text (DraftJS) blocks."""
+
     def __init__(self, context, request):
         self.context = context
         self.request = request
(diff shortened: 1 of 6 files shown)


src/plone/restapi/indexers.py
```diff
@@ -2,7 +2,7 @@
 # This is an experimental feature meant for use in Volto only!
 # This code is likely to change in the future, even within minor releases.
 # We will make sure plone.restapi latest always works with the latest Volto release.
-# This code is planned to being refactored into CMFPlone 6.0 as soon as Volto 4 final is out.
```
tisto, line 5: @ericof haha. That comment did not age well...

src/plone/restapi/tests/test_blocks_searchable_text.py
```diff
@@ -231,10 +234,12 @@ def test_search_slate_text(self):
                 "value": [],
             },
         }
-        self.doc.blocks_layout = [
-            "38541872-06c2-41c9-8709-37107e597b18",
-            "4fcfeb9b-f73e-427c-9e06-2e4d53b06865",
-        ]
+        self.doc.blocks_layout = {
```
tisto, line 237: @ericof jikes. We tested the wrong thing. Good catch.

tisto (approved):
LGTM! I'd like @tiberiuichim @avoinea @nzambello @cekk to have a second look before merging.

tisto:

src/plone/restapi/indexers.py
```diff
@@ -55,59 +59,48 @@ def __call__(self, value):
...
+class SlateTextIndexer:
+    """Searchable Text indexer for Slate blocks."""
+
+    def __init__(self, context, request):
+        self.context = context
+        self.request = request
+
+    def __call__(self, block):
+        block = block or {}
+        # BBB compatibility with slate blocks that used the "plaintext" field
```
ksuess, line 73: plaintext is not current practice?

ericof: I was surprised as well, but it seems it was agreed searchableText should be the way to go.

src/plone/restapi/indexers.py
ericof: That is why I "edited" it.

src/plone/restapi/indexers.py
```diff
@@ -55,59 +59,48 @@ def __call__(self, value):
...
 def SearchableText_blocks(obj):
+    """Extract text to be used by the SearchableText index in the Catalog."""
     request = getRequest()
     blocks = obj.blocks
+    blocks_layout = obj.blocks_layout
     blocks_text = []
-
-    for block in blocks.values():
-
+    for block_id in blocks_layout.get("items", []):
```
tiberiuichim, line 84: I think we need to accept the reality of "blocks in blocks". I can mention already two scenarios that will surely affect you: the teasers grid (ok, maybe not the end of the world if that content is not searched) and the accordion block.

See https://github.com/plone/plone.restapi/blob/9c04a3c421300a5f7bfb6732cc0177fa302446a6/src/plone/restapi/deserializer/blocks.py#L54 which has support for deserializers.

I think we should have a separate "block traversing utility". If needed, I can provide some examples, but it would be the equivalent of this: https://github.com/plone/volto/blob/616c19f420aa3117ffd94b2dc99fe4c1da1f03bf/src/helpers/Blocks/Blocks.js#L348

ericof: I do believe we should trust the "block container" to provide us with the **searchableText** attribute or force a standard way for blocks telling us they have sub-blocks (afaik, we do not have that yet as there are implementations with sub-blocks in blocks or data attributes).

tiberiuichim: no... it's gonna be unwieldy to implement this Volto side. What do you do with a columns block inside an accordion block inside a tabs block? All of that extra weight in the json will start piling up, but it's really simple to handle this in the backend by following the convention on how block data would look like.

tiberiuichim: > force a standard way for blocks telling us they have sub-blocks

we have a convention that's followed by the volto-columns-block, volto-accordion-block, volto-tabs-block (any many other EEA addons). The convention is `blocks+blocks_layout` either directly in the block data or in a `data` field of the block data. The volto-blocks-grid unfortunately doesn't follow the convention https://github.com/kitconcept/volto-blocks-grid/blob/da622c3b1635051be0d53476c784e379145db7de/src/components/Grid/Edit.jsx#L180 although I remember that at some point there was discussion to get that migrated, see Volto Team Meeting notes from 2021-10-12

tiberiuichim: We might also do "typed values", see https://github.com/plone/volto/issues/2252

tiberiuichim: So, we've actually had a initiative about a year ago to introduce a new interface function (something declared in a block's config) to allow a block to declare its searchableText, so that it can be lifted to its parent container block. But it wouldn't completely solve the problem, as the parent container block needs to know if it's inside another block. Or otherwise have an "onChange" callback built into the Edit and Add forms, which could recursively go through the blocks and call their api methods.

ericof: I will add recursion here and follow the convention of "blocks" or "data" with "blocks", and search for searchableText attributes there.
Do you think this is a good compromise @tiberiuichim ?

tiberiuichim: @ericof yes, that would be perfect.

ericof: @tiberiuichim Implemented :-)

src/plone/restapi/indexers.py
```diff
@@ -74,6 +70,58 @@ def __call__(self, block):
...
+    more granular control of the indexing.
+
+    :param block: Dictionary with block information.
+    :param obj: Context to be used to get a IBlockSearchableText.
+    :param request: Current request.
+    :returns: A string with text found in the block.
+    """
+    result = ""
+    block_type = block.get("@type", "")
+    subblocks = extract_subblocks(block)
```
tiberiuichim, line 107: Maybe move this line lower, to avoid unnecessary processing?

tiberiuichim (approved):

Issue #1252, opened 2021-10-27
Querystring Search Performance


tisto:
# Possible Solutions

## Use GET instead of POST

* Send GET request instead of POST (for @querystring-search)
* Send query payload as body (query param contains arrays and nested structures)
* The URL should contain a subtraversal path: (e.g. "@querystring-search/1669e4ae9c5258f973ea00a6c31578a9)

````
GET /plone/folder/@querystring-search/1669e4ae9c5258f973ea00a6c31578a9 HTTP/1.1
Accept: application/json
Authorization: Basic YWRtaW46c2VjcmV0
Content-Type: application/json

{
    "body": {
        "b_size": "2",
        "b_start": "0",
        "fullobjects": "False",
        "limit": "10",
        "query": [
            {
                "i": "Title",
                "o": "plone.app.querystring.operation.string.is",
                "v": "Welcome to Plone"
            },
            {
                "i": "path",
                "o": "plone.app.querystring.operation.string.path",
                "v": "/news"
            }
        ],
        "sort_on": "sortable_title",
        "sort_order": "reverse",
    },
    "id": "doc1",
    "method": "GET",
}
````

## Problems to solve

* [ ] plone.rest needs to be able to send a GET request with payload (see https://github.com/plone/plone.rest/pull/125)
* [ ] plone.rest needs to be able to encode/decode payload that contains arrays and nested structures (see https://github.com/plone/plone.rest/pull/125)
* [ ] plone.rest needs to be able to get/ignore the hash in the subtraversal (might just work because of acquisition?)
* [ ] the frontend needs to be able to generate a unique hash value for a given query
* [ ] the frontend needs to be able to send a GET request with payload in the body (that works with arrays and nested structures)
* [ ] Varnish needs to be able to cache the URL "GET /plone/folder/@querystring-search/1669e4ae9c5258f973ea00a6c31578a9"
* [ ] Cloudflare and other services need to be able to cache the URL "GET /plone/folder/@querystring-search/1669e4ae9c5258f973ea00a6c31578a9"

tisto:
@jensens on our way back home from Sorrento, @jackahl and I came up with an idea of how to solve the querystring-search caching issue. We could, as discussed, use a GET request instead of POST and use subtraversal for a unique URL. Though, instead of either encoding the complex query in the URL as GET params (which has lots of possible side effects and the character limit), we could just send the query in the body and generate a unique hash in the URL to make it unique. I outlined the idea above:

[...]

PR #1295, opened 2021-12-18
Enhance the vocabularies serializer to accept a list of tokens

sneridagh:
We will need this next week.

### DIFF
--- a/docs/source/vocabularies.rst
+++ b/docs/source/vocabularies.rst
@@ -90,9 +90,9 @@ Filter Vocabularies
 
 .. http:get:: (context)/@vocabularies/(vocab_name)?title=(filter_query)
 .. http:get:: (context)/@vocabularies/(vocab_name)?token=(filter_query)
-.. http:get:: (context)/@vocabularies/(vocab_name)?token_list=(filter_query1)&token_list=(filter_query2)
+.. http:get:: (context)/@vocabularies/(vocab_name)?tokens=(term1,term2,...)
 
-Vocabulary terms can be filtered using the ``title``, ``token`` or ``token_list`` (array) parameter.
+Vocabulary terms can be filtered using the ``title``, ``token`` or ``tokens`` (array) parameter.
 
 Use the ``title`` parameter to filter vocabulary terms by title.
 E.g. search for all terms that contain the string ``doc`` in the title:
@@ -117,7 +117,7 @@ E.g. search the term ``doc`` in the token:
     You must not filter by title and token at the same time.
     The API returns a 400 response code if you do so.
 
-Use the ``token_list`` parameter to filter vocabulary terms by a list of tokens.
+Use the ``tokens`` parameter to filter vocabulary terms by a list of tokens.
 
 ..  http:example:: curl httpie python-requests
     :request: ../../src/plone/restapi/tests/http-examples/vocabularies_get_filtered_by_token_list.req
--- a/docs/source/vocabularies.rst
+++ b/docs/source/vocabularies.rst
@@ -90,7 +90,7 @@ Filter Vocabularies
 
 .. http:get:: (context)/@vocabularies/(vocab_name)?title=(filter_query)
 .. http:get:: (context)/@vocabularies/(vocab_name)?token=(filter_query)
-.. http:get:: (context)/@vocabularies/(vocab_name)?tokens=(term1,term2,...)
+.. http:get:: (context)/@vocabularies/(vocab_name)?tokens=(filter_term1)&tokens=(filter_term2)&...
 
 Vocabulary terms can be filtered using the ``title``, ``token`` or ``tokens`` (array) parameter.
 
--- a/docs/source/vocabularies.rst
+++ b/docs/source/vocabularies.rst
@@ -90,10 +90,11 @@ Filter Vocabularies
 
 .. http:get:: (context)/@vocabularies/(vocab_name)?title=(filter_query)
 .. http:get:: (context)/@vocabularies/(vocab_name)?token=(filter_query)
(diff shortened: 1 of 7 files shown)


tiberiuichim:
May I suggest `tokens` instead of `token_list` as parameter? Also, I think we'd need to have a POST service here: https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/services/vocabularies/configure.zcml

sneridagh:
I test drive it, and it needs adjustments. Do not merge yet.

@tiberiuichim Better not POST for caching, right? Also token and tokens, wouldn't it be far too similar and error prone?

sneridagh:

tisto:
We should make a decision regarding tokens vs token_list and add it to our naming convention:

https://github.com/plone/plone.restapi/blob/master/docs/source/conventions.rst

I will do some research...

sneridagh:

sneridagh:
@tisto @tiberiuichim ready. Used `tokens` as parameter. Sticked to the GET request for caching purposes.

tiberiuichim (approved):
As far as I understand, we're not gonna be able to have tokens with commas in them, right? I don't know how common this use case is.

sneridagh:
@tiberiuichim you have a point, ideally not... but who knows... could be a vector of bugs. 

The whole "let's decide how we encode lists in parameters" story is also important, I just used what we already have in there.

ksuess:
Same topic, how to handle list parameters, in search endpoint. 
? Is it a default for a JSON API to handle commas as list separators?
search endpoint gets 
`@@search?metadata_fields:list=subject,modified&portal_type:list=News Item,Event`
then searches for
`query {'metadata_fields': ['subject,modified'], 'portal_type': ['News Item,Event']}`
which returns an empty result.

tiberiuichim:
that list argument conversion is not correct. https://zope.readthedocs.io/en/latest/zdgbook/ObjectPublishing.html#argument-conversion. 

> list – collect all values with this name into a list. 

It should be `@search?metadata_fields:list=subject&metadata_fields:list=modified`

sneridagh:
@ksuess in the search endpoint we are constrained by how Zope Querystring parser search parameters are encoded so the action is taking already care of passing Zope the right way (see the action code) in jsx, when you call the action, just pass a JS array in the action.

sneridagh:
hehe, @tiberiuichim was faster than me :)

tisto:
This PR raises a number of REST API design principles questions:

1. Do we want to use multiple parameters dependent on if we pass a single parameter (e.g. "token=foo") or a list of parameters (e.g. "tokens=foo,bar"). Some APIs use a single parameter, no matter if you pass a single parameter or multiple parameters (e.g. "token=foo" and "token=foo,bar"). I think I personally prefer to use a param that accepts both a single term and a list.
2. Do we want to include the type of the passed param into our param name (token vs token_list). I'd be ok with the plural solution that we currently have in this PR. Including the type might be a slippery slope.
3. URL encoding lists. There are a bazillion ways of handling this. You can just urlencode stuff (like "foo,bar" -> "foo%82bar"). You can pass a parameter multiple times (e.g. token=foo&token=bar). You can encode stuff like "[foo,bar]", ...

I started to dip my toe into urlencoding arrays and nested structures for the querysting-search endpoint (https://github.com/plone/plone.restapi/issues/1252#issuecomment-956095050) and I have to admit it is just a horrible mess and there is no standard whatsoever as far as I can see.

tiberiuichim:
@tisto There's also the option of using Zope's argument parsing, with `tokens:list=foo&tokens:list=bar`

ale-rt:
My 2 c., do what we are used to do! 

Either use:
- `tokens=foo&tokens=bar`
or:
- `tokens:list=foo&tokens:list=bar`
if you want to be explicit.

Notes:
1. while there is no difference when the parameter appears once in the query string, id does make a difference when it is present just once, i.e. `token=foo` will be give you a `self.request.form` equal to `{"token": "foo"}`, while `token:list=foo` will give you a `self.request.form` equal to `{"token": ["foo"]}` (notice the list).
2. If you opt for `tokens=foo,bar` you must protect your code from people passing the parameter multiple times, i.e.: `tokens=foo,bar&tokens=baz`. 
3. Just a curiosity, query string can be exploited and Zope does a great job in preventing that. There is tons of stuff to read, start from https://en.wikipedia.org/wiki/HTTP_parameter_pollution. IIRC also facebook was found vulnerable to this attack (PHP...)

jensens:
I agree with @ale-rt - and I tend to not reinvent the wheel and use Zope list query parameter syntax.

sverbois:
just my 2c

When I design REST API I always use a plural noun when the API client can provide mutliple values. It is more predictive for the user.
Here I would use:
`tokens:list=foo`for the singleton list `tokens=['foo']`
 `tokens:list=foo&tokens:list=bar` for a multi-valued list  `tokens=['foo','bar']`

sneridagh:

sneridagh:
@ale-rt @jensens refactored to use multi valued, as in Zope. I will merge if all is good from you!

ale-rt:
Mine was just a suggestion :)
I would wait for the reviews of @tisto and @ericof ;)

sneridagh:
@ale-rt a good one! Thanks for weigh in! We already talked about it in today's Volto Team meeting and agreed to go this way.

tisto (approved):
I am good with this PR. ToDo:

* [ ] Add the multiple params approach to API design best practices
* [ ] Add test and docs to show that you can pass a single param with "tokens" as well
* [ ] Deprecate "token" for the future
* [ ] Refactor usage of other approaches (e.g. "," in expendables) to support the new "standard", deprecate the old way for the next major release

ale-rt:
Given that you are into json you might also be interested in making this happen: https://github.com/zopefoundation/Zope/issues/957

PR #1302, opened 2021-12-26
test(warning): Resolve bulk of warnings from tests

rpatterson:
Resolve all the deprecation warnings that originate in this package's code that are
exposed by running the tests.  Also, fix or workaround `ResourceWarning: unclosed
<socket.socket ...>` warnings when running the tests.

With these changes, the bulk of warnings output while running the tests have been
removed.  This means there's a *lot* less to ignore in the test output and we should not
be training ourselves to ignore test output.

### DIFF
--- a/src/plone/restapi/testing.py
+++ b/src/plone/restapi/testing.py
@@ -253,11 +253,28 @@ class RelativeSession(requests.Session):
     base if their URL is relative (doesn't begin with a HTTP[S] scheme).
     """
 
-    def __init__(self, base_url):
+    def __init__(self, base_url, test=None):
+        """
+        Capture the base URL.  Optionally also capture a test case for cleanup.
+
+        Apparently, network sockets created by the `requests` library can remain open
+        even after the full body of the response has been read, despite [the
+        docs](https://docs.python-requests.org/en/latest/user/advanced/#body-content-workflow). In
+        particular, this results in `ResourceWarning: unclosed <socket.socket ...>` leak
+        warnings when running the tests.  If the `test` kwarg is passed, it will be used
+        to register future cleanup calls to close this session and thus also the
+        sockets. If passed, it must be an object with a `addCleanup(func)` callable
+        attribute, such as instances of `unittest.TestCase`.
+        """
         super().__init__()
+
         if not base_url.endswith("/"):
             base_url += "/"
         self.__base_url = base_url
+
+        if hasattr(test, "addCleanup"):
+            # Avoid `ResourceWarning: unclosed <socket.socket ...>`
+            test.addCleanup(self.close)
 
     def request(self, method, url, **kwargs):
         if urlparse(url).scheme not in ("http", "https"):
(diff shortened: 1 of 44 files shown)


rpatterson:

rpatterson:

jensens (approved):

src/plone/restapi/testing.py
```diff
@@ -253,12 +253,16 @@ class RelativeSession(requests.Session):
     base if their URL is relative (doesn't begin with a HTTP[S] scheme).
     """
 
-    def __init__(self, base_url):
+    def __init__(self, base_url, test=None):
```
tisto, line 256: @rpatterson can we document the "test" param here? Why we need it, what kind of values we expect, etc. I'd like to be a bit more expressive here, so that other devs in the future have a chance to understand why we did this here.

tisto:
Apart from the docs issue I raised above, LGTM.

src/plone/restapi/testing.py
wesleybl: This parameter shouldn't be required, not to have more warnings in the future?

rpatterson: I don't quite follow, @wesleybl. It is optional, isn't it? Can you elaborate on your question?

rpatterson: > @rpatterson can we document the "test" param here?

Good catch and apologies for not documenting in the code originally.  Done.

rpatterson:

rpatterson:
Everything is green, LMK if there's anything further I need to do to see this through to merge, @tisto.

src/plone/restapi/testing.py
wesleybl: > I don't quite follow, @wesleybl. It is optional, isn't it? Can you elaborate on your question?

I'm questioning if the test parameter shouldn't be required. If a new test is created and the test parameter is not passed, we will have new warnings.

rpatterson: > > I don't quite follow, @wesleybl. It is optional, isn't it? Can you elaborate on your question?
> 
> I'm questioning if the test parameter shouldn't be required. If a new test is created and the test parameter is not passed, we will have new warnings.

Ah, you're suggesting maybe we should make it required.  Thanks for clarifying.  I made it optional because I figured making it opt-in would be less disruptive to other packages that use `RelativeSession`.  I know of at least one external usage and I didn't want to require that everything be released at once.

wesleybl: Understand. The ideal then is to fix these external uses, after a release, and come back here to make it required.

rpatterson: > Understand. The ideal then is to fix these external uses, after a release, and come back here to make it required.

I'm not sure I agree with making it required since I don't  think it even belongs there.  See my write up in the commit message.

wesleybl: >I'm not sure I agree with making it required since I don't think it even belongs there

For test classes with `self.api_session` attribute it would be possible to make in a "Base Test Class". But I understood in your comment that it can be done at another time. Thanks!

rpatterson: > For test classes with `self.api_session` attribute it would be possible to make in a "Base Test Class".

Yup, that's exactly what I imagined if I refactored the tests for DRY.

PR #1306, opened 2022-01-04
Be permissive when testing the schema of the querystring endpoint

reebalazs:
Allow any additional property to be defined by plone.app.querystring,
otherwise no properties can ever be added. The change also makes sure
that all code depending on any existing attributes continues to work.
This is needed in favor of
https://github.com/plone/plone.app.querystring/pull/104 .

### DIFF
--- a/src/plone/restapi/tests/test_services_querystring.py
+++ b/src/plone/restapi/tests/test_services_querystring.py
@@ -64,7 +64,14 @@ class TestQuerystringEndpoint(unittest.TestCase):
             "values": {},
             "vocabulary": None,
         }
-        self.assertEqual(expected_field_config, idx)
+        # Be permissive with the check and only check the existing
+        # attributes. (This gives plone.app.querystring to extend its schema
+        # when that becomes necessary, while making sure that all code depending
+        # on any existing attributes continues to work.)
+        filtered_idx = {}
+        for key in expected_field_config:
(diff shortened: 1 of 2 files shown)


src/plone/restapi/tests/test_services_querystring.py
```diff
@@ -64,7 +64,14 @@ def test_endpoint_shows_field_config(self):
             "values": {},
             "vocabulary": None,
         }
-        self.assertEqual(expected_field_config, idx)
+        # Be permissive with the check and only check the existing
```
tisto, line 67: @reebalazs this is certainly a smart solution. Though, when it comes to testing I prefer simplicity and readability over DRY and being smart. One of the most important assets of a test is that it explicitly explains what values we expect and that it returns meaningful responses to test failures (and does not require checking a loop).

I know this is a lot more to type but what I usually do is just add lots of those lines:

```
self.assertEqual("Text search of an item's title", idx.get("description"))
```

This approach is robust and returns an explicit failure when a value is missing.

The other option is to rely on "self.assertDictContainsSubset(".

As said, that's a matter of personal preference...

tisto:

tisto:
FYI: I will merge this one now because it does the trick. We can refactor later to improve the readability of the test if we want to.

tisto (approved):

PR #1313, opened 2022-01-16
Update plone 6a2

sneridagh:


sneridagh:
@tisto @jensens what's the trick to make it work in both 5 and 6? I tried it up and keeps breaking.

I saw that @ericof is using pip and a custom GH action in https://github.com/plone/plone.volto/blob/main/.github/workflows/tests.yml

jensens:
Buildout still works fine. We better should depend on released constraints, so we are in sync. I dont understand the comment `# Keep this file in sync with: https://github.com/kitconcept/buildout/edit/master/requirements.txt` - I think this is historical and must be changed, since we are in Plone core know.

sneridagh:
@jensens agreed, but how to maintain both builds working with the same requirements/constraints? any pointer?

jensens:
> how to maintain both builds working with the same requirements/constraints?

Impossible. One need to point to `https://dist.plone.org/release/5.2-latest/[versions.cfg|constraints.txt|requirements.txt]`, the other to `https://dist.plone.org/release/6.0[a2|-dev]/*`.

sneridagh:
Superseded by #1314

PR #1315, opened 2022-01-18
7.x.x working copy

avoinea:
Backport Working copy support for 7.x.x


avoinea:
This depends on **plone.app.iterate 4.x** which is Plone 5+ only compatible :thinking:

PR #1327, opened 2022-02-17
Added 'View comments' and 'Reply to item' permission to discussion

razvanMiu:
Related to this pr https://github.com/plone/plone.app.discussion/pull/191.

Added `permissions` object to `ConversationSerializer`:
* `view_comments` permission is used to show the conversation to users that have the `'View comments'` permission from plone.app.discussion
* `can_reply` permission is used to allow users that have the `'Reply to item'` permission to add comments

Added `can_reply` permission to `CommentSerializer` used to allow users that have `'Reply to item'` permission to reply to an existing comment.


jensens:
Permission in `configure.zcml` for all endpoints is View. Does this need refinement too? Just wondering.

src/plone/restapi/services/discussion/utils.py
```diff
@@ -28,13 +28,27 @@ def permission_exists(permission_id):
     return permission is not None
 
 
+def can_view(context):
+    """Returns true if current user has the 'View comments' permission."""
+    if not permission_exists("plone.app.discussion.ViewComments"):
```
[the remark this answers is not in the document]
razvanMiu: `View comments` permission is not yet in `plone.app.discussion`. It will be added after this pr is merged https://github.com/plone/plone.app.discussion/pull/191.
The default behaviour of comments viewlet from `plone.app.discussion` is that everyone that has `View` permission can see the comments.
So I added this fallback in case someone uses a version of `plone.app.discussion` in which `View comments` permission doesn't exist. Without the fallback `can_view` would be `False` if there is no `View comments` permission.

avoinea:
> Permission in `configure.zcml` for all endpoints is View. Does this need refinement too? Just wondering.

:smile: True. @tisto?!

:thinking: On the other hand if we change to `plone.app.discussion` permissions we'll need to add a hard dependency to it.

src/plone/restapi/serializer/discussion.py
```diff
@@ -31,6 +33,10 @@ def __call__(self):
         results["@id"] = batch.canonical_url
 
         results["items_total"] = batch.items_total
+        results["permissions"] = {
+            "view_comments": can_view(self.context),
```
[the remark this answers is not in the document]
razvanMiu: Yes, I will send an empty array if can_view is False

jensens:
> On the other hand if we change to `plone.app.discussion` permissions we'll need to add a hard dependency to it.

plone.app.discussion is a hard dep of Products.CMFPlone 

https://github.com/plone/Products.CMFPlone/blob/c42578a04a77b8a719b49b847a4da6e03b1c0be3/setup.py#L56

src/plone/restapi/services/discussion/configure.zcml
```diff
@@ -13,31 +13,31 @@
...
+      permission="plone.app.discussion.ViewComments"
       name="@comments"
       />
 
   <plone:service
       method="POST"
       factory=".conversation.CommentsAdd"
       for="*"
-      permission="zope2.View"
+      permission="plone.app.discussion.ViewComments"
```
avoinea, line 24: @razvanMiu I think this should be `Reply to item`

src/plone/restapi/services/discussion/configure.zcml
```diff
@@ -13,31 +13,31 @@
       method="GET"
       factory=".conversation.CommentsGet"
       for="*"
-      permission="zope2.View"
+      permission="plone.app.discussion.ViewComments"
```
avoinea, line 16: @razvanMiu @jensens I'm still not sure about this as it will introduce a hard-dependency on `plone.app.discussion >= 4.x`

jensens: version wise probably it does. But OTHO its clean this way.

tisto: @avoinea if we introduce such a hard dependency we need to reflect this in setup.py (with all the downsides that come with that). The alternative would be to use zcml conditional with a fallback (which increase the overall complexity). I need to find a free minute to think about this (not easy these days tbh).

avoinea: @tisto If you ask me, I would leave `zope2.View` in zcml, as we already check for `plone.app.discussion.ViewComments` in python code 

https://github.com/plone/plone.restapi/blob/611063c108488ec6b0c8b2b26f3327dd8bb98340/src/plone/restapi/serializer/discussion.py#L27-L29

https://github.com/plone/plone.restapi/blob/611063c108488ec6b0c8b2b26f3327dd8bb98340/src/plone/restapi/services/discussion/utils.py#L31-L35

jensens: Given the current situation I am fine with `View` - but in a future release this should be changed. It is probably a good idea to create an issue to not forget about it.

razvanMiu:
jenkins-run-jobs

razvanMiu:

jensens (approved):

avoinea:
@tisto I think this is ready to be merged.

avoinea (approved):

tisto (approved):

PR #1349, opened 2022-04-02
Add missing blank line in README

tisto:


jensens:
Change OK ;-) 

But looking at the list above: a) meanwhile it is misleading, giving the impression it is not used often, b) I do not think this is the right place, moreover this should go into case studies/marketing material? 
Open new issue?

/cc Marketing Team @rioksane @ericof ...

PR #1337, opened 2022-02-25
Import docs into `plone/documentation`

stevepiercy:
See https://github.com/plone/plone.restapi/issues/1326

PR #1350, opened 2022-04-02
Revert "Import docs into `plone/documentation`"

tisto:
Reverts plone/plone.restapi#1337


tisto:
@stevepiercy just a test to see if this fixes the CI build.

stevepiercy:
The same tests for Plone 6.0 still fail. It must be something else.

tisto:
@stevepiercy yeah. seems something changed outside of our repo. In theory that shouldn't be possible. So either some dependency in Plone 6 coredev is unpinned, or the released versions on dist.plone.org changed without notice. @mauritsvanrees @thet any insights that could help here? This has something to do with the "mockup" package.

mauritsvanrees:
In Plone 6 mockup is no longer a Python package: it is only npm. It should not be pulled in or imported anymore.

tisto:
@mauritsvanrees I honestly have no clue how and why the mockup package is fetched here. I am also clueless about how it is possible that the CI job starts to fail without any change in the repo. This means something outside of the repo must have changed (dist.plone.org? undeclared/unpinned dependency?).

@thet @agitator @MrTango did we change anything regarding mockup lately? Maybe released a newer package that is fetched by the Plone 6 buildout?

tisto:
@stevepiercy ok, here is the strange thing. I just re-ran the last successful build before merging your latest PR and it still passes (three times):

https://github.com/plone/plone.restapi/actions/runs/1968154868/attempts/1
https://github.com/plone/plone.restapi/actions/runs/1968154868/attempts/2
https://github.com/plone/plone.restapi/actions/runs/1968154868

Maybe this is a GHA caching issue. The only way to find that out would be to actually revert the PR and see if we manage to get the master branch green again. Then re-apply stuff. What do you think?

stevepiercy:
You can filter the failing workflow to [tests.yml](https://github.com/plone/plone.restapi/actions/workflows/tests.yml), then page through until you find a likely candidate for when tests started to fail and then they were fixed. I paged through until I found the first green for "Tests", and here that is:

https://github.com/plone/plone.restapi/pull/1194

If that doesn't help, and as a last resort, then we could try your suggestion. To me it looks like tests have been broken for quite some time, and unrelated to my changes. In fact we were all green at this point: https://github.com/plone/plone.restapi/commit/783f8aad63172632ecded6f8a8f096e2fe0b3bff

Screen Shot 2022-04-04 at 8 27 18 PM

tisto:
@stevepiercy here is the list with the filter on the master branch: 

https://github.com/plone/plone.restapi/actions/workflows/tests.yml?query=branch%3Amaster

When things do not make any sense in a CI system the only thing that worked for me in the past is to revert until the master branch is green again. Then re-apply in very small steps.

stevepiercy:
Nice filter! I think you're on to something with @mauritsvanrees' PR.

Screen Shot 2022-04-04 at 8 52 55 PM

If we compare the last passing and first failing commits:

https://github.com/plone/plone.restapi/compare/96e0bc119ecb847112b1310860b43e5c86cd5a54..ea19b3672f899fdbbbcdd2fda1b6b27fad23c534

My first guess would be that the import of `Products.CMFPlone.factory` is the cause of the failure because I saw references to it in the failing tests. Or maybe it is this?

https://github.com/plone/plone.restapi/runs/5800532592?check_suite_focus=true#step:11:633

Anyway, if the changes in the commit comparison are reverted, do tests pass?

tisto:
@stevepiercy we have a chicken-egg problem here. We rely on checkout of CMFPlone:

https://github.com/plone/plone.restapi/pull/1255#issuecomment-955668089

This makes our CI system brittle by definition. We can not rely on anything. Though, we can not get rid of this without checking in on a broken build. This happens when the Plone community ignores fundamental CI best practices...

tisto:
@stevepiercy I guess we won't get out of this mess without a new Plone 6.0.0a4 release of CMFPlone. cc @mauritsvanrees @ericof

mauritsvanrees:
Wouldn't it help to temporarily add `mockup` to the eggs?

Wait a minute, I see that I already did that. See commit 3fa8f820deb51af30811f882a3e0479096e87e97.
It is part of the long overdue PR to upgrade the PyJWT dependency, PR #1194.

The GHA tests of that PR are green, see especially the [test run after my commit](https://github.com/plone/plone.restapi/actions/runs/2039732026).

We could merge that PR and then master should be green again.
Afterwards the PyJWT pin in `buildout.coredev` needs to be updated, which is its own chicken-and-egg-problem.
Let me create a

mauritsvanrees:
I created a PR that only adds the mockup egg: PR #1354.

stevepiercy:
@tisto can this PR be closed now?

PR #1353, opened 2022-04-05
Rely on requirements.txt from dist.plone.org

tisto:


davisagli:
Something similar was done in the meantime

Issue #1322, opened 2022-02-02
portal_type is missing in @types endpoint


erral:
The portal_type id is missing in the `@types` endpoint:

https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/services/types/get.py#L73

The API consumer has to "hack" the `@id` attribute to get the portal_type in order to use it in for instance when creating the link to create a new content:

https://github.com/plone/volto/blob/master/src/components/manage/Toolbar/Types.jsx#L24

Adding such an attribute would simplify the work not only in the Toolbar but also in some other places where the portal_types are shown (search block selector for instance).

PR #1355, opened 2022-04-05
Add portal_type title

razvanMiu:
This adds a new field `type` to content response which contains the title of `portal_type`.

Example of why this is needed: 
Let's say I have a content-type named `Topic page` with the id `topic_page`. Fetching a page of this content-type will respond with:
```
{
   '@type': 'topic_page',
   ...
}
``` 
There is no way to get the human readable title of the content-type so this pr will add a new field and the response will look like:
```
{
   '@type': 'topic_page',
   'type': 'Topic page',
   ...
}
```


erral:
Fixes #1322

src/plone/restapi/serializer/dxcontent.py
```diff
@@ -55,11 +56,13 @@ def __call__(self, version=None, include_items=True):
         parent_summary = getMultiAdapter(
             (parent, self.request), ISerializeToJsonSummary
         )()
+
         result = {
             # '@context': 'http://www.w3.org/ns/hydra/context.jsonld',
             "@id": obj.absolute_url(),
             "id": obj.id,
             "@type": obj.portal_type,
+            "type": get_portal_type(obj.portal_type, self.request),
```
jensens, line 65: `@type` vs `type` is at a maximum confusing. I would propose to use something like `type_title` or `type_name`.

jensens (changes requested):

src/plone/restapi/serializer/dxcontent.py
razvanMiu: Changed type to type_name

avoinea:

src/plone/restapi/deserializer/blocks.py
```diff
@@ -162,7 +162,6 @@ def __init__(self, context, request):
         self.request = request
 
     def __call__(self, block):
-
```
nileshgulia1, line 165: @razvanMiu @avoinea I wonder how code-analysis checks pass due to this🤔

ksuess:
Assuming this is for Volto: A mapping from fti id to fti title can be received from state.types. No need to extend the serializers.

avoinea:
@ksuess Can you detail a little. Is it available for anonymous users?

ksuess:
> @ksuess Can you detail a little. Is it available for anonymous users?

So issue description has to be read as "There is no way to get the human readable title of the content-type as long the user is `Anonymous`"

avoinea:
@jensens @tisto What's preventing this for being merged?

news/1355.feature
```diff
@@ -0,0 +1 @@
+Add portal_type title (`type`) to content response [razvanMiu]
```
davisagli, line 1:
```suggestion
Add portal_type title (`type_name`) to content response [razvanMiu]
```

src/plone/restapi/serializer/dxcontent.py
```diff
@@ -62,11 +63,14 @@ def __call__(self, version=None, include_items=True):
         parent_summary = getMultiAdapter(
             (parent, self.request), ISerializeToJsonSummary
         )()
+        parent_summary["type_name"] = get_portal_type(parent.portal_type, self.request)
+
```
davisagli, line 67: `type_name` is included in the summary serializer, so I don't think it needs to be added here.

src/plone/restapi/serializer/nextprev.py
```diff
@@ -23,8 +24,9 @@ def __init__(self, context):
 class NextPrevious:
     """Facade with more pythonic interface"""
 
-    def __init__(self, context):
+    def __init__(self, context, request):
```
davisagli, line 27: This is not a backwards-compatible change. You can use getRequest from zope.globalrequest to get the request without changing the parameters here.

src/plone/restapi/serializer/utils.py
```diff
@@ -38,3 +41,10 @@ def uid_to_url(path):
         if a_href:
             return a_href
     return href
+
+
+def get_portal_type(portal_type, request=None):
```
davisagli, line 46: I'd suggest calling this `get_portal_type_title`, since `portal_type` is often used to refer to the id of the FTI, and this is not that.

src/plone/restapi/services/types/get.py
```diff
@@ -72,6 +72,7 @@ def __call__(self, expand=False):
         result["types"] = [
             {
                 "@id": f"{portal_url}/@types/{fti.getId()}",
+                "@type": fti.getId(),
```
davisagli, line 75: This seems misleading. If JSON data is labeled with `@type: "X"` I would expect it to include the fields from the X type. But this includes the properties of an FTI object. The type is FTI, not a specific portal_type.

davisagli (commented):
I have mixed feelings about this. I can imagine that it could be useful to always have the type title (so that it's easy to show the type in a listing of search results, for example). On the other hand it's a bit of extra processing that is done even if the type title will not be used. Probably nothing significant, though.

I added a few comments about specific details.

news/1355.feature
davisagli: (Why do we call it name instead of title, since it's the title?)

nileshgulia1: @davisagli type confuses with `@type` and changed as per https://github.com/plone/plone.restapi/pull/1355#discussion_r862338290

src/plone/restapi/services/types/get.py
nileshgulia1: @davisagli The idea is to avoid getting `@type` from `@id` like https://github.com/plone/volto/blob/52975f103cb7520fa375bbd5ae2a844b6e446ed4/src/components/manage/Toolbar/Types.jsx#L24  See: https://github.com/plone/plone.restapi/issues/1322

jensens:
I think it is okay to get the FTI title property delivered as well. 

To overcome the processing overhead I suggest to use the [dexterity FTI cache and its `lookup_fti` method](https://github.com/plone/plone.dexterity/blob/master/plone/dexterity/schema.py#LL54C5-L54C15).

As @davisagli mentioned, the naming is confusing and needs some love.

Also, I do not see any documentation added.

stevepiercy:
I would also suggest that the dozens of changes in this PR that remove a blank line be reverted and put into a separate PR. That should be handled by code quality tools.

nileshgulia1:
> I would also suggest that the dozens of changes in this PR that remove a blank line be reverted and put into a separate PR. That should be handled by code quality tools.

Indeed. I just ran `make black` and `make zpretty` and saw a lot of files formatted. This happens even if you do it in a fresh clone in master 🤔 .

jensens:
> That should be handled by code quality tools.

This package lacks of plone/meta so, this will solve it once applied here.

nileshgulia1:

nileshgulia1:
One thing I don't get is where to put the documentation about this extra field?  Isn't it self-explanatory with being just added in `http-examples`?

src/plone/restapi/serializer/nextprev.py
nileshgulia1: Done in https://github.com/plone/plone.restapi/pull/1355/commits/ff979592b7c204a54c419ef6879e2011b3b74ab4

jensens:
> One thing I don't get is where to put the documentation about this extra field? Isn't it self-explanatory with being just added in `http-examples`?

@stevepiercy may I head this one over to you?

stevepiercy:
@jensens, I agree with @nileshgulia1. What do you think is needed in the way of documentation?

I found no place where the REST API authors have documented the keys in a response. Do they want to start?

nileshgulia1:

nileshgulia1:

src/plone/restapi/services/types/get.py
davisagli: @nileshgulia1 My point is that `@type` has a meaning in JSON-LD: it identifies the format of the object that is represented in the JSON there. But you are using it for something else (the id of the portal type that is defined by this FTI) ... uh, it's a bit hard to explain because there are two different kinds of types we're talking about here

Can you change it to use `id` instead? Then I am fine with it.
```suggestion
                "id": fti.getId(),
```

src/plone/restapi/tests/test_services_types.py
```diff
@@ -81,7 +80,8 @@ def test_get_types(self):
         )
         for item in response.json():
             self.assertEqual(
-                sorted(item), sorted(["@id", "title", "addable", "immediately_addable"])
+                sorted(item),
+                sorted(["@id", "title", "addable", "immediately_addable", "@type"]),
```
davisagli, line 84:
```suggestion
                sorted(["@id", "id", "title", "addable", "immediately_addable"]),
```

nileshgulia1:

davisagli (approved):

davisagli:
@jensens okay now from your point of view?

avoinea:

avoinea:

src/plone/restapi/services/contextnavigation/get.py
```diff
@@ -251,7 +250,7 @@ def heading_link_target(self):
             # Go to the item /view we have chosen as root item
             return nav_root.absolute_url()
 
-    def root_type_name(self):
+    def root_type_title(self):
```
tisto, line 253: @razvanMiu @nileshgulia1 can you elaborate on where this name change comes from?

tisto: I guess this is a search/replace issue. I will revert that line. Let me know if that's ok with you.

PR #1357, opened 2022-04-06
Add test that tests the redirector behavior of plone.restapi

tisto:


tisto:
@ericof @sneridagh I wrote a test to prove that plone.restapi behaves in a wrong way when re-using a plone.app.redirector redirect. Though, it turns out the test passed. I just proved that it also works like this in real life. I have no clue why this didn't work before.

ericof:
@tisto I "fixed" the test, now it proves there is an infinite loop

ericof:
A PR is open on [plone.rest](https://github.com/plone/plone.rest/pull/134)

sneridagh:
@tisto released it, however, then should we make p.restapi depend on p.rest 2.0.0a5 as minimum requirement?

ericof:
@sneridagh Plone 6 should depend on p.rest 2.0.0a5, right?

tisto:
@sneridagh @ericof I'd prefer to avoid hard requirements in setup.py. They cause lots of trouble for everyone who is using buildout. I'd prefer to pin the newer versions in our buildout and in the released Plone versions.

tisto:
@sneridagh @ericof I am totally out of the loop here. Is that test something we want in plone.restapi or shall we close the issue?

davisagli:
@tisto The issue was fixed in plone.rest and a similar test added there. I don't think we need a copy of the test here.

PR #1358, opened 2022-04-06
Checkout specific commit of CMFPlone.

mauritsvanrees:
Fixes buildout error for missing plone.base package, from the merged PLIP 3395.

See [failed run](https://github.com/plone/plone.restapi/runs/5856639094?check_suite_focus=true) on plone.restapi master.

### DIFF
--- a/plone-6.0.x.cfg
+++ b/plone-6.0.x.cfg
@@ -7,9 +7,10 @@ find-links = https://dist.plone.org/release/6.0.0a3/
 versions=versions
 auto-checkout =
     Products.CMFPlone
+always-checkout = true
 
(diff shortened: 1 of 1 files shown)


mauritsvanrees:
An alternative is to extend https://dist.plone.org/release/6.0-dev/versions.cfg as I do on commit 4bec09220abb432acda4119f98bc3df1ce1dc28d of a [different branch](https://github.com/plone/plone.restapi/compare/maurits-use-plone-60-dev). But that seems to need a newer zc.buildout and/or setuptools, so it would be a bit more work.

tisto (approved):

PR #1359, opened 2022-04-08
Fix tests for https://github.com/plone/Products.CMFPlone/issues/3486

pbauer:


tisto:
@pbauer can you elaborate on this? I guess we might need an if/else here for different Plone versions? Keep in mind that plone.restapi needs to support Plone 5.2 and Plone 6 at the same time.

pbauer:
I'll add if/else clases. After https://github.com/plone/plone.restapi/pull/1364 is merged we can check here for Plone 6 or not without messing with alphas.

pbauer:
Wrong, since the tests run with 5.2 and 6.0.0a4 and 6.0 coredev all cases need to be covered. I'd rather set `displayed_types` to the defaults of Plone 6  at the beginning of the tests.

pbauer:
I guess now it's ready for merge

pbauer:

PR #1360, opened 2022-04-08
Upgrade to Plone 6.0.0a4. Remove CMFPlone checkout.

tisto:

PR #1364, opened 2022-04-11
Upgrade CI to Plone 6.0.0a4

mauritsvanrees:
This takes the first commit of PR #1360 and then fixes the test setup by splitting a new `plone6.yml` off from the largely unchanged `tests.yml` which is then still used for Plone 5.

### DIFF
--- a/.github/workflows/plone6.yml
+++ b/.github/workflows/plone6.yml
@@ -1,4 +1,4 @@
-name: Tests Plone 5
+name: Tests Plone 6
 on: [push]
 jobs:
   build:
@@ -6,8 +6,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        plone-version: ['5.2']
-        python-version: ['3.6' '3.7', '3.8']
+        plone-version: ['6.0']
+        python-version: ['3.7', '3.8', '3.9']
 
     steps:
       # git checkout
@@ -27,11 +27,9 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-pip-
 
-      # python install
-      - run: pip install virtualenv
-      - run: pip install wheel
+      # pip install
       - name: pip install
-        run: pip install -r requirements.txt -r requirements-docs.txt
+        run: pip install -r requirements-60.txt -r requirements-docs.txt
 
       # buildout
       - name: buildout
@@ -43,22 +41,12 @@ jobs:
       - name: code analysis
         run: bin/code-analysis
 
-      # build sphinx
-      - name: sphinx
-#        run: if [ "${{ matrix.plone-version }}" == "5.2" ] && [ ${{ matrix.python-version }} == '3.7' ]; then bin/sphinxbuilder; fi
-        run: if [ "${{ matrix.plone-version }}" == "5.2" ] && [ ${{ matrix.python-version }} == '3.7' ]; then make docs-html; fi
-
       # test
       - name: test
         run: bin/test
 
+      # Commented out a few actions that should not currently be run on Plone 6.
+      # Kept a note here, so we do not forget about them when we drop Plone 5 support.
+      # build sphinx
       # test no uncommited changes
-      - name: test no uncommited changes
-        run: bin/test-no-uncommitted-doc-changes
-        env:
-          PYTHON_VERSION: ${{ matrix.python-version }}
-          PLONE_VERSION: ${{ matrix.plone-version }}
-
       # test sphinx warnings
-#      - name: sphinx
-#        run: if [ "${{ matrix.plone-version }}" == "5.2" ] && [ ${{ matrix.python-version }} == '3.7' ]; then bin/test-no-sphinx-warnings; fi
--- a/requirements-60.txt
+++ b/requirements-60.txt
@@ -1,0 +1,12 @@
+# from https://dist.plone.org/release/6.0.0a4/requirements.txt
+setuptools==62.0.0
+zc.buildout==3.0.0rc3
+pip==22.0.4
+wheel==0.37.1
+
+# Windows specific down here (has to be installed here, fails in buildout)
+# Dependency of zope.sendmail:
+pywin32 ; platform_system == 'Windows'
+
+# SSL Certs on windows, because Python is missing them otherwise:
+certifi ; platform_system == 'Windows'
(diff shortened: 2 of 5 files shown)


mauritsvanrees:
Green now, when you ignore the changelog and Jenkins tests.

mauritsvanrees:
@plone/restapi-team please have a look.

jensens (approved):

requirements-60.txt
```diff
@@ -0,0 +1,12 @@
+# from https://dist.plone.org/release/6.0.0a4/requirements.txt
```
tisto, line 1: @mauritsvanrees if I recall correctly we agreed that we will consider all files on dist.plone.org as release artifacts (except "latest" of course) which means they must not change under any circumstances. If we agree on this we could also rely on dist.plone.org directly. If we can not make that commitment, we should keep a copy IMO.

.github/workflows/plone6.yml
```diff
@@ -0,0 +1,52 @@
+name: Tests Plone 6
```
tisto, line 1: @mauritsvanrees thanks for this! I personally tend to prefer keeping the test configuration in one file and rather work with an if/else statements. The main reasoning behind this is that we should try to establish a common best practice and naming conventions for GHA files for all Plone repos. If we establish tests.yml for tests, devs always know where to look for the test matrix. When we use ci.yml, plone5.yml, plone6.yml, we won't have this. I'd be happy to refactor this myself. What do you think?

jensens: I like the idea of having more than one file, because it is understandable. Huge conditional files a re clumpsy, tend to introduce side effects and are difficult to maintain.

requirements-60.txt
mauritsvanrees: I treat them as unchangeable. In 6.0.0a4 after release a problem with `plone.volto` was discovered and then fixed. Since this was within a day (half a day even) and in an alpha release, I decided to change the version pin. That is definitely an exception.

.github/workflows/plone6.yml
mauritsvanrees: I would expect them in one file for most packages.
But in this case there are several small differences, requiring conditions, making it hard to read like Jens says.
Also, I think `tests.yml` may be changed by your company-internal script that you occasionally use to update the Makefile and other files.
And if Plone 5 and 6 are separate, it becomes easier for Jens, me and others to keep the Plone 6 part up to date, without having to worry about accidentally changing Plone 5 in a way that you don't agree with.

tisto: @mauritsvanrees there is no company-internal script whatsoever that touches the GHA. I'd like to keep it in a way that it is maintainable for me in the long run when plone.restapi will continue to support multiple Plone versions at the same time (e.g. Plone 6.1, 6.2, 7.x, etc.). The GHA structure should also stay consistent for plone.restapi, plone.volto and other Volto add ons. That's also something that we need to consider here.

Plone RESTAPI is an abstraction layer to the underlying Plone version and it needs to stay that way, otherwise we won't be able to develop the frontend and the backend in two different velocities in the future. If we start treating Plone 5 and 6 differently (and even split responsibilities) we are going in the wrong direction IMHO.

Can you elaborate on how plone6.yml and plone5.yml differ except for using a different buildout cfg and requirements.txt?

requirements-60.txt
tisto: If we treat dist.plone.org as a host for release artifacts there should be no exception from that rule IMHO. In that case, we should make a new release. If a CI system can not 100% rely on an external resource to not be changed it shouldn't use them. That's a lesson I had to learn the hard way over and over again in the past years.

.github/workflows/plone6.yml
mauritsvanrees: > there is no company-internal script whatsoever that touches the GHA.

Maybe not a script then, but you still copy over some files from a central repository now and then, right? Ah, that is done by the [`Makefile`](https://github.com/plone/plone.restapi/blob/master/Makefile#L1-L2). It says to keep this in sync with a kitconcept Makefile that I _thought_ was internal because I was not allowed view the url at the top of this file. But that is because it points to the _edit_ form of this file.

[`make update`](https://github.com/plone/plone.restapi/blob/3eb90104e84df25abbfddc40b89eae6f9fb308ea/Makefile#L33-L39) indeed does not touch the GHA yaml files. It *does* make lots of changes when I run it on master, including removing Plone 6 and documentation improvements in the Makefile. But maybe this should just be taken now as an initial step, followed by a manual check and fixes.

Now about the yaml differences for 5 and 6 in this PR. In `plone6.yml`:

* No `pip install virtualenv wheel`, because `pip install -r requirements-60.txt` is enough.
* I thought of not running `pip install -r requirement-docs.txt` because we do not need it, but kept it for now as it works.
* I also thought of letting `bin/buildout` only install the `test` and `code-analysis` parts, but kept it as all parts can be built.
* Don't build sphinx.
* Don't run `bin/test-no-uncommitted-doc-changes`
* Don't run `bin/test-no-sphinx-warnings`

It could be better instead to move all the sphinx stuff to a separate yaml file that only runs on 5.2-3.7, and probably simply runs on the code that is in the repository, without running the tests. That would make `tests.yml` more understandable, and then the remaining differences for 5.2 and 6.0 are manageable. Running `bin/test-no-uncommitted-doc-changes` on all is okay, as the script itself quits when run on anything else than 5.2-3.7.

Also, sphinx is installed both by buildout and by pip. We should use only one method.

Also, I wonder if it is still useful to install and run `bin/code-analysis` in `tests.yml`, when there are already separate yaml files for black, flake8, pyroma and zpretty.

jensens: The point is: Please keep it simple, base it on the "standards" we have (buildout.coredev, buildout.plonetest, jenkins.plone.org, Plone GH-Action , things like that. The current state of the whole here is not approachable. Best proof is that even our @mauritsvanrees  starts an own file in order to sort out things. I struggle too. 

Copying from company specific repos (except for experimental stuff) is a no go. After experimental phase this should be moved in collective or plone orga.

mauritsvanrees:
Superseded by the now merged PR #1410.

PR #1400, opened 2022-05-04
Docs: add best practices for naming attributes.

tisto:


docs/source/conventions.md
```diff
@@ -73,6 +73,41 @@ Mixing singular and plural is confusing.
 For example, using `/users` for retrieving users, and `/user/21` to retrieve a single user.
 
 
+### URL Attributes (singular vs plural)
+
+**Rule:** Use plural for attributes that can contain one or multiple items. Use singular for attributes that can contain only one single item.
```
davisagli, line 78: I'm not sure what you mean by URL attributes, which is not a term I've seen used before. #1295 was talking about querystring parameters but the examples here look like path segments. Maybe you want this to apply to both?

jensens: Do you want to rename a whole bunch of endpoints and deprecate the other form?  If not we end up with a non consistent API.

tisto: I am referring to attributes that we pass via GET requests though the URL.

davisagli: I'm pretty sure these are usually called query string params or parameters, not attributes. If I search for url attribute on google it actually corrects it and shows results for url parameters

jensens: ah, got it, yes, that should be called parameters.

tisto: Changed it. Thx for the feedback.

stevepiercy (approved):

docs/source/conventions.md
```diff
@@ -73,6 +73,41 @@ Mixing singular and plural is confusing.
...
+```text
+/token
+/token_list
+/token_array
+/token_set
+```
+
+**Reason:**
+
+The naming should clearly indicate if an attribute expects a single (singular) item or multiple items (plural). We decided to use plural instead of appending "_list" or "_array" or "_set".
```
ale-rt, line 107: Uhm, please check:
- https://docs.quantifiedcode.com/python-anti-patterns/readability/putting_type_information_in_a_variable_name.html
- https://stackoverflow.com/questions/111933/why-shouldnt-i-use-hungarian-notation

tisto: @ale-rt thanks for the pointers. Though, we do NOT recommend to use "_array", "_set" and stuff. Maybe we have to work on the wording if that paragraph is misleading. :)

Feel free to improve the wording or tell me if you object to the decision to go with plural for parameters that accept multiple parameters.

ale-rt: I am not objecting! 
And I do not like to waste our time on philosophical digression on how variables should be name.
I am just pointing out that other people did that and concluded that `token_list` was a bad variable name.

With that said, you can use it anyway, nobody will sue you!

Sorry for the noise and have a great sprint!

stevepiercy: I had to read that paragraph twice myself. @ale-rt I think we both got confused. The intent is "Do not append the type to the name of the variable".

Let me take a crack at rewording this so that it does not trip up other people, too.

ale-rt: Oh, yeah... sorry

docs/source/conventions.md
```diff
@@ -73,6 +73,41 @@ Mixing singular and plural is confusing.
...
+
+If the parameter allows passing a single value only:
+
+```text
+/token
+```
+
+**Don't:**
+
+If the parameter allows passing multiple values:
```
stevepiercy, line 96:
```suggestion
If the parameter allows passing multiple values, do not use the singular form, and do not append a type to the variable name. The following are examples of what _not_ to do:
```

docs/source/conventions.md
stevepiercy, line 107:
```suggestion
The naming should clearly indicate if an attribute expects a single (singular) item or multiple items (plural).
We decided to use the plural form, instead of appending a possible type—such as `_list`, `_array`, or `_set`—to the variable name.
```

stevepiercy (commented):
Possible suggestions to clarify intent of this section.

docs/source/conventions.md
stevepiercy: No need to apologize. It threw me for a loop, too. 🤪

tisto: @ale-rt @stevepiercy thank you both for your feedback! The fact that you misunderstood the paragraph shows that we need to improve it. I am open to suggestions here. :)

tisto: Just scrolled down. :) Thanks @stevepiercy!

stevepiercy (approved):
Yay!

PR #1408, opened 2022-05-05
use plone.base and overhaul isort

jensens:
- isort was not following out standards - now it does
- use plone.base to prevent deprecation warnings


mauritsvanrees:
Add a '_compat.py' with conditional imports and let the rest of the code import from there? Then it is contained in one place.

tisto:
@jensens let's discuss this 1:1 :)

mauritsvanrees:
I guess a `_compat.py` like this already covers 99 percent of the cases:

```
import pkg_resources
try:
    pkg_resources.find_distribution("plone.base")
except DistributionNotFound:
    from Products.CMFPlone import interfaces
    from Products.CMFPlone import utils
else:
    from plone.base import interfaces
    from plone.base import utils
```

Elsewhere:

```
from ._compat import interfaces, utils
```

But I will leave you two to it. :-)

Issue #1397, opened 2022-05-04
@rules endpoint


avoinea:
In Plone Classic for each object we have a menu called `Rules` pointing to `/@@manage-content-rules`. Here you can see and assign Content Rules contextually.

In order to be able to replicate this behavior in Volto, we need a RestAPI endpoint called `@rules`.
It should be listed in components and expandable and it should follow the RestAPI CRUD approach.

Verb | URL | Action
-- | -- | --
GET | `/test-page/@rules` | List rules assigned here
POST | `/test-page/@rules` | Assign rule to this context
DELETE | `/test-page/@rules/{:id}` | Unassign content rule to this context

See also https://github.com/plone/plone.restapi/issues/324

ericof:
Implemented by #1496

Issue #324, opened 2017-04-20
Content rules


robgietema:
Add endpoints to manage content rules.

tisto:
+1

rpatterson:
I've been assigned to work on this by Eau de Web so I went ahead and assigned myself to this issue.  LMK if there's any reason I should proceed, if there's any existing work, or if there's anything else I should know.

ericof:
Implemented by #1496

PR #1411, opened 2022-05-06
@rules endpoint

avoinea:
See #1397 and #324


fredvd:
As mentionned at the Volto team meeting tuesday 7th, 'we' are aware of the recent testing issues. These are unstable robottests that fail occasionally that crept into core and where a problem with the reporting let the number increase too much before we noticed them.  And then it became a situation of always some (and too many) seemingly random tests failing, but it was always a subset. 

We're almost back to zero failing tests most of the runs, I manually started the pull-request-6.0-3.8 yesterday evening, which we have modfied locally on jenkins to only run robot tests and with some other tweaks/experiments to improve the overall tests.   I'll get back to this PR when we are finished and run the jobs again.

fredvd:

avoinea:

avoinea:
To be continued https://github.com/plone/plone.restapi/pull/1496

PR #1414, opened 2022-05-06
move caching ruleset to definitions of endpoints

jensens:
see https://github.com/plone/plone.app.caching/pull/100


jensens:

ericof (commented):
I just have a question about the cache:ruleset element. Is it always available? Because the solution I was thinking of was to create plone.restapi.cache and conditionally include it if plone.app.caching was installed

jensens:
There is a zcml condition on each. Having it side by side with the endpoint registration seems better to me and documents best practice.

Then NS declaration does nothing, lookup kicks in after true condition.

I just found I need to adjust the tests in plone.app.caching, may move them here.

ericof:
Amazing!

mauritsvanrees (approved):
LGTM

tisto:

mauritsvanrees:
The 'no uncommitted changes' test [fails on 3.7-6.0](https://github.com/plone/plone.restapi/runs/6348778500?check_suite_focus=true). Has nothing to do with this PR. Let me try a quick fix in a separate PR.

ericof (approved):
LGTM

PR #1420, opened 2022-05-09
Move the generation of two test scripts out of buildout.

mauritsvanrees:
The buildout parts theoretically could automatically replace something in the scripts, but this feature was not used.
So in practice it simply renamed the files and copied them to the bin directory.
Instead, I now rename them in the repository, and call them directly in GHA.
Simplifies things.

Case in point: the 3.7-6.0 tests currently fail because `bin/test-no-uncommitted-doc-changes` is not found:
https://github.com/plone/plone.restapi/pull/1414#issuecomment-1120897719


mauritsvanrees:
Green enough: changelog and Jenkins checks can be ignored for this change.

Issue #1409, opened 2022-05-06
Type schema is available only for authenticated users


erral:
The Type Schema (`/@types/WhateverType`) is available only for authenticated users:

https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/services/types/get.py#L120
https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/services/types/get.py#L23

If we want to have a DefaultView in Volto for schema based content-types and this view should be viewed by anonymous users, we need to remove the permission check here.

Opinions?

@ionlizarazu @pbauer

avoinea:
@erral I think we should remove that check and rely only on the permission defined within zcml https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/services/types/configure.zcml#L11

mauritsvanrees:
The zcml permission currently says anyone with View permission in the Plone Site (or anywhere) can access it. So that will be everyone for most sites.

This is an example of the output you get:
https://plonerestapi.readthedocs.io/en/latest/types.html#get-the-schema-with-get

I suppose it is possible that there is a Choice field where the possible options contain secrets. Not too likely, but it could happen. In that case I would expect the field to have a permission, granted to for example only Site Administrators.

So I think we should check the permission for fields, and only include those fields that the user is allowed to see. It does not look like [this happens currently](https://github.com/plone/plone.restapi/blob/d428463dc763539adba726cd9ce0ef94b3ad1129/src/plone/restapi/types/utils.py#L351). If we do not check this, then the auto-generated view will contain all fields, also those that the user is not allowed to see.

PR #1423, opened 2022-05-09
allow anonymous users to access @types endpoint

erral:
Addresses #1409


tisto:
@erral what's the status of this PR? Can you elaborate your use case?

erral:
I don't recall why we added this PR...

I think it must be somehow related to the work we did during the Beethoven Sprint when we added in Volto a default view for schema based content-types

davisagli:
@tisto @erral I'm aware of 2 use cases that came up for this.

The first is the default schema-based view, like you said. That still does not work for anonymous users. But, this change stalled because there were some concerns about exposing private values in the schemas (such as vocabulary values).

The second is the use case of showing content type titles in listings. We solved that a different way in https://github.com/plone/plone.restapi/pull/1355

PR #1427, opened 2022-05-12
Black=22.3.0

tisto:


mauritsvanrees (changes requested):
You need a version less than 22, otherwise Python 2 code may break.

erral:
Be careful because this change can't be backported to the 7.x.x branch, because in version 22.1.0 the python2 formatting support was removed.

mauritsvanrees:
Ah right, master is still for 5.2, but not for 2.7. So should be fine here after all.

mauritsvanrees (approved):

versions.cfg
```diff
@@ -5,8 +5,9 @@ zc.buildout =
 plone.restapi =
 
 # code analysis
-black = 20.8b1
+black = 22.3.0
```
wesleybl, line 8: With version 22.3.0 of `black`, no need to pin `click = 8.0.4`. See https://github.com/psf/black/pull/2966

wesleybl: Wouldn't it be better to pin the `black` in a constraints.txt? `black` is not used in buildout.

tisto: Though, on second thought we are about to move away from buildout in the long run, so it might be worth to consider. Will look into it...

PR #1428, opened 2022-05-12
Do not break path2uid with strange links

cekk:
We had a some strange links in migrated data from a very old Plone. For example "../..".
With these changes, the method does not break and returns the given link.


cekk:

tisto (approved):

PR #1431, opened 2022-05-17
Proposal for the relations-endpoint

### LINKED ISSUE #1432: Add relations endpoint
A relation endpoint should allow to get, add and delete relation tuples per context similar to the relation module in plone.api (https://6.docs.plone.org/plone.api/relation.html). 

It should also allow to get, add and delete relation tuples globally to a relations control panel (https://github.com/plone/volto/issues/3096).

pbauer:
Fixes #1432

**TODO**
- [ ] GET: limit parameter


ksuess:
GET request: It would be nice to get not only the dictionary of relation tuples by relation, but also the request and some items_total info.
grafik

ksuess:
I played with the new module. Apart from
- adding additional info to the GET request: requesting uri and items_total info
- the not solved question, if and how a second endpoint @relations-catalog could be abanndoned (please, what's the use case for a relation affecting the Plone site root?)
I would like to suggest some changes on documentation.
Instead of many change suggestions I'll provide a PR.

sneridagh:
My 2cents!

I'm still kind of torn on the naming, I still haven't gone through the new duality full-fledged DX root vs portal root. I guess it's my main pain point here. Aside naming, I can understand that for the relations control panel, you need a centralised query point in the "portal root", where you can issue general queries to retrieve bulk and generic relations data, but the fact that we are introducing the ability of doing all the operations in two ways, centralised and in-object... it kind of unease me.

@ksuess since you have experienced and tried out first hand while doing the control panel, what's your take on this?

@plone/volto-team opinions?

ksuess:
It's the decision if Plone site root should be a potential source of a relation element.

If not, then /root/@relations is the endpoint request for global stuff.

pbauer:
That's not really a open decision since the site-root is now dexterity and has a intid it can have relations. The use-case to relate from there to other content also makes sense to me. 
The only option I see to have only one endpoint and avoid duplication and is to remove the one on the content and handle everything using `target`, `source` and `relationship` parameters on the site-root.

ksuess:
@pbauer, @sneridagh, @tisto, adding 'UID' to metadata_fields in ISerializeToJsonSummary is not necessary. I reverted the commit. Instead the UID is added just for @relations and @relations-catalog endpoint.

https://github.com/plone/plone.restapi/pull/1431/commits/0773403428060ca0a2aced7a0e95f048c1dde7c2

src/plone/restapi/services/relations/get.py
```diff
@@ -0,0 +1,148 @@
+from AccessControl.SecurityManagement import getSecurityManager
+from collections import defaultdict
+from plone.restapi.interfaces import ISerializeToJsonSummary
+from plone.restapi.serializer.converters import json_compatible
+from plone.restapi.services import Service
+from Products.CMFCore.utils import getToolByName
+from Products.CMFPlone.relationhelper import get_relations_stats
```
erral, line 7: If this targets also Plone 5.2 we need to rework this import and write the method directly here because this does not exist in Plone 5.2.x

src/plone/restapi/services/relations/get.py
```diff
@@ -0,0 +1,148 @@
...
+        data = get_relations(
+            source=source,
+            target=target,
+            relationship=relationship,
+            request=self.request,
+        )
+        # @ksuess prefers "@id": self.request.environ["REQUEST_URI"]
+        # as this is all that is of interest, but uri is a convention.
+        return {
+            "@id": f'{self.request["SERVER_URL"]}{self.request.environ["REQUEST_URI"]}',
```
erral, line 108: This breaks for me in Plone 5.2 with `KeyError: 'REQUEST_URI'`

erral (commented):
The `@relations-catalog` endpoit is a bit slow in a client's site with 3571 relations, but apart from that I think that the `@relations` endpoint is very useful.

pbauer:
At the Volto team meeting on 2022-05-24 we discussed two options:

1. We could drop the endpoint on content and only have a `/@relations` endpoint on the portal that works with the  source/target/relation parameters as proposed for the `@relations-catalog` endpoint. 
2. We could use `/@controlpanels/relations` instead of `/@relations-catalog` on the site-root. This would be similar to the dexterity types endpoint (https://plonerestapi.readthedocs.io/en/latest/controlpanels.html#dexterity-types). 

No decision has been made so far.

avoinea:
@stevepiercy Could you please take a look at the conflicting file?

stevepiercy:
> @stevepiercy Could you please take a look at the conflicting file?

Done. I merged `master` into this branch to bring in the reorganization change, then cleaned up `relations.md`, which could still use a little more filling in, assuming y'all want to use request and response files for the examples.

@pbauer this PR still needs a changelog entry and blackification, but the docs are OK by me to get something published now. We can always fill in the details later, if details are desired.

davisagli:
The latest work on this is in https://github.com/plone/plone.restapi/pull/1602, if I understand correctly.

PR #1433, opened 2022-05-19
Tests: patch unique url for scale in old or new way.

mauritsvanrees:
This is only in serializer tests for images.

uuid4 is only used until plone.scale 6.0.0a4.
https://github.com/plone/plone.scale/pull/57 introduces a new way.

We also patch the _modified_since method to always return True.
Otherwise you may get info from a different scale back, precisely because we give all scales the same "unique" id,
which then is of course no longer unique, making the logic unstable.
This is needed for the newer plone.scale versions, but should be perfectly fine for the older ones.

The `plone.scale` PR allows us to get the urls without really generating the actual scales and store them in blobs. But this PR does **not** use it yet. Happy to do that, but that can be done slightly later, like after there is a new plone.scale release. With the current PR the tests are at least prepared. They were failing in Jenkins with checkouts of plone.scale and plone.namedfile.

### DIFF
--- a/news/57.bugfix
+++ b/news/57.bugfix
@@ -1,0 +1,3 @@
+Tests: patch unique url for scale in old or new way.
+This is only in serializer tests for images.
+[maurits]
--- a/src/plone/restapi/tests/helpers.py
+++ b/src/plone/restapi/tests/helpers.py
@@ -1,4 +1,7 @@
+from contextlib import contextmanager
+from plone.scale import storage
 from Products.CMFCore.utils import getToolByName
+from unittest.mock import patch
 from urllib.parse import urlparse
 
 import quopri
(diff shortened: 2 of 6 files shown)


mauritsvanrees:

mauritsvanrees:
All tests are green including Jenkins:

- https://jenkins.plone.org/job/pull-request-6.0-3.7/1378/
- https://jenkins.plone.org/job/pull-request-6.0-3.8/1372/
- https://jenkins.plone.org/job/pull-request-6.0-3.9/1304/

So this can be reviewed and merged. @plone/restapi-team 

I will now start jobs together with the plone.scale and plone.namedfile PRs. But no need to wait on that.

mauritsvanrees:

PR #1438, opened 2022-05-25
Fix broken link to Python requests library docs

stevepiercy:


stevepiercy:
@ksuess builds were not complete, and there was at least one failure. Is it OK to ignore that when the change clearly does not cause the failure?

PR #1442, opened 2022-06-06
WIP Enhance relation field serialization for image content types

reebalazs:
- In better support of the preview image link
- Reuses serialization from existing field serializers
- Flexible adapation allows new types to be added in the future

WIP until preview_image_link has been merged into plone.volto


base.cfg
```diff
@@ -18,6 +18,7 @@ parts =
 develop = .
 sources-dir = extras
 auto-checkout =
+    plone.volto
```
sneridagh, line 21: We don't really need to tie this to the plone.volto change. So I'd remove this for the merge.

src/plone/restapi/serializer/relationobject.py
```diff
@@ -0,0 +1,85 @@
...
+        serializer = getMultiAdapter((field, self.rel_obj, self.request))
+        # Extend the default values with the content specific ones
+        additional = serializer()
+        if additional is not None:
+            result.update(additional)
+        return result
+
+
+@adapter(IImage, IRelationChoice, IDexterityContent, Interface)
+class ImageRelationObjectSerializer(FieldRelationObjectSerializer):
```
sneridagh, line 81: Interesting approach, wondering how this will scale in the case that we have a custom content type and we want to return, let's say two images.

sneridagh: or a combination of fields. Not sure if make a 1:1 content type => field serialised I would like to include in the final serialization is a good choice.

sneridagh (commented):
Take a look at my comments. 

/cc @ericof @tisto @tiberiuichim I'd like to know your point of view too.

sneridagh:
@tiberiuichim @avoinea this is what I was talking about yesterday. The idea would be that once we have this Relation serializers, we could use them in the smart fields (href, teaser's one) so if what's behind it's really an image, then pull off the enhanced serialization.

davisagli (commented):
I am not sure I understand the motivation for this feature.

If I understand correctly, it adds a way to customize the serialization of a related item from a RelationChoice field, with the ability to use a different adapter based on what type of object is related.

But it seems strange to me that we would want different fields included in the serialization when an object was found via a RelationChoice field compared to some other way like a catalog query or getting the actual content path.

For the use case of images in particular, I think it is now possible to include image scale sizes in the image summary serialization generated from catalog brains, and that could be done whether or not the image was referenced from a relation field.

jensens:
+1 - this would speed up usage of referenced images a lot.

PR #1445, opened 2022-06-09
Pin Sphinx version and make Netlify happy

ksuess:


tisto (approved):

netlify.toml
```diff
@@ -2,4 +2,4 @@
   PYTHON_VERSION = "3.8"
 
 [build]
-  ignore = "git diff --quiet $CACHED_COMMIT_REF $COMMIT_REF ./docs/"
+  ignore = "git diff --quiet $BRANCH $HEAD -- ./docs/source/"
```
ksuess, line 5: Comparing the changes on the branch. This has the effect, that the preview is done or at least kept, even when a further commit to the branch does not include any changes in /docs/source/.

stevepiercy: The Netlify log says `fatal: bad revision 'pull/1445/head'`:

https://app.netlify.com/sites/plone-restapi/deploys/62a194ddac9c4d0008ab7e5a#L14

It still builds, so maybe it is OK? `¯\_(ツ)_/¯` Or perhaps it builds for every commit, and not only for changes in `/docs/source`? You can test by pushing two trivial commits, one with a change in `/docs/source` only, and one outside.

I also checked the [Netlify logs](https://app.netlify.com/sites/plone-restapi/deploys), and it was building only for changes inside `/docs/source/` up until this change. It's possible it was a one-time fluke.

While we're looking at this directory, can you delete `/docs/requirements.txt`? It has been replaced by `/requirements-docs.txt`.

stevepiercy (changes requested):
We need to test diffs with the new Netlify configuration. See comment.

Also remove obsolete file.

tisto:
@stevepiercy @ksuess status?

stevepiercy:
@tisto I merged master to this branch, fixed merge conflicts, and pushed. Let's see what Netlify builds.

davisagli:
We switched from Netlify to Read the Docs in #1798, so this is outdated.

PR #1450, opened 2022-06-10
handle subblocks in IPloneSiteRoot in Plone 5.2

### LINKED ISSUE #1449: Subblocks are not handled in IPloneSiteRoot in Plone 5.x
Plone Site Root serializer has special code to handle the serialization in Plone 5.x:

https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/serializer/site.py#L93

This code handles the calls to `IBlockFieldSerializationTransformer` manually instead of using the `IFieldSerializer` adapters for `IJSONFields` at https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/serializer/blocks.py#L24

But that code does not handle the subblocks that may be present in the root of the site, which the IFieldSerializer does for `IBlock`:

https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/serializer/blocks.py#L60

https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/serializer/site.py#L138

erral:
Fixes #1449 

cc: @ionlizarazu @libargutxi

### DIFF
--- a/src/plone/restapi/serializer/site.py
+++ b/src/plone/restapi/serializer/site.py
@@ -6,7 +6,7 @@ from plone.restapi.batching import HypermediaBatch
 from plone.restapi.interfaces import IFieldSerializer
 from plone.restapi.interfaces import ISerializeToJson
 from plone.restapi.interfaces import ISerializeToJsonSummary
-from plone.restapi.serializer.blocks import handle_block_value
+from plone.restapi.serializer.blocks import apply_block_serialization_transforms
 from plone.restapi.serializer.converters import json_compatible
 from plone.restapi.serializer.expansion import expandable_elements
 from plone.restapi.services.locking import lock_info
@@ -134,5 +134,5 @@ class SerializeSiteRootToJson:
         if not blocks:
             return blocks
         for id, block_value in blocks.items():
-            blocks[id] = handle_block_value(block_value, self.context)
+            blocks[id] = apply_block_serialization_transforms(block_value, self.context)
         return blocks
--- a/src/plone/restapi/serializer/site.py
+++ b/src/plone/restapi/serializer/site.py
@@ -135,6 +135,7 @@ class SerializeSiteRootToJson:
         if not blocks:
             return blocks
         for id, block_value in blocks.items():
+            self.handle_subblocks(block_value)
             block_type = block_value.get("@type", "")
             handlers = []
             for h in subscribers(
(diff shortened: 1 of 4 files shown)


erral:

erral:

src/plone/restapi/serializer/site.py
```diff
@@ -149,3 +150,34 @@ def serialize_blocks(self):
 
             blocks[id] = block_value
         return blocks
+
```
davisagli, line 153: I see at least one other difference between what serialize_blocks does here and what the "normal" blocks serializer does in https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/serializer/blocks.py#L21 -- the latter also sets h.blockid on the handler.

Maybe it would be worth refactoring BlocksJSONFieldSerializer to use a function that could also be used from here. Then there would be less code duplication and less risk of it diverging over time.

erral: I tried to unify both in one place and write, say, a common base class for both. But one is a `IFieldSerializer` and the other one a `ISerializeToJson` so when trying to inherit one from the other, Zope complains of not having a proper zcml declaration.

Moreover the methods here are using `self.context` and `self.request` to get some subscribers to do transforms, so converting them to plain functions would mean to start passing context and request as parameters to those functions, what at least seems a bit ugly.

I am not favouring the "copy-paste" solution, just I don't see some other way to do it.

jensens: Another option is to factor the relevant part out into a general function and call it from both classes/its methods.

erral:
I have extracted the block handling methods to be functions and have call them from the relevant places. It is a bit ugly to pass `context` as a parameter but that's what we have 🤷‍♂️

erral:

davisagli:
@erral Thank you! I think it is fine passing the context like that.

Small suggestion: I might name the function to be slightly more descriptive, so `apply_block_serialization_transforms` rather than `handle_block_value`. But this is a nitpick and I'll go ahead and approve.

davisagli (approved):

erral:
Thanks. I think the build is broken right now because of https://github.com/plone/plone.restapi/pull/1471 I think we first need to apply https://github.com/plone/plone.restapi/pull/1475 to revert the other one and then merge this (and others)

erral:

davisagli (approved):

PR #1451, opened 2022-06-10
Test-only fix: normalize white space in html in some tests.

mauritsvanrees:
Needed to not fail with newer plone.outputfilters.
Similar to the fix in https://github.com/plone/plone.app.discussion/pull/202, see explanation there.


mauritsvanrees:

pbauer (approved):

PR #1452, opened 2022-06-10
Tests: fixed order of roles in http examples.

mauritsvanrees:
"test no uncommitted changes" fails, see for example [this run](https://github.com/plone/plone.restapi/runs/6806798950?check_suite_focus=true#step:12:67). No idea why the PR from earlier this week that added this was green.

(Plus tiny fix in an unrelated news snippet so another check is green.)


mauritsvanrees:

PR #1461, opened 2022-07-20
fix: plone.restapi must work with dependency on only Products.CMFPlone

jensens:
Currently restapi depends on `plone.app.iterate`. 
The assumption, that it is always there, is wrong: it is an (internal) addon, only installed when dependening on 'Plone'.


jensens:

jensens:
The failing test seems unrelated to my changes.

jensens:

mauritsvanrees (approved):

PR #1464, opened 2022-07-29
Add Users and Groups controlpanels to default registry

nileshgulia1:
I just noticed we still don't have "users" controlpanel in the default registry. I need this to have the schema for `many_users` and `many_groups`.

### DIFF
--- a/news/1463.feature
+++ b/news/1463.feature
@@ -1,0 +1,1 @@
+add users and groups controlpanel to registry. [nileshgulia1]
--- a/src/plone/restapi/controlpanels/registry.py
+++ b/src/plone/restapi/controlpanels/registry.py
@@ -113,3 +113,21 @@ class UserGroupControlpanel(RegistryConfigletPanel):
     configlet_category_id = "plone-users-and-groups"
     group = "Users and Groups"
     title = "User and Group Settings"
+
+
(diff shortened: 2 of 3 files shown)


ksuess:
@nileshgulia1 many_users is available via 'usergroup' panel. You can get it in Volto via

```
  const many_users = useSelector(
    (state) => state.controlpanels.controlpanel?.data?.many_users,
  );
  const many_groups = useSelector(
    (state) => state.controlpanels.controlpanel?.data?.many_groups,
  );


  useEffect(() => {
    dispatch(getControlpanel('usergroup'));
  }, [dispatch]);
```

Did I get you wrong maybe? As I see, user panel and group panel are not needed here.

nileshgulia1:
> useEffect(() => {
    dispatch(getControlpanel('usergroup'));
  }, [dispatch]);

@ksuess I didn't want to hardcode "usergroup" in a "users" controlpanel endpoint as it might look confusing. Also in future we might have multiple schema in it. I think it's nice to have a controlpanel registered with their specific names 🤔. But yes, it works well with `getControlpanel("usergroup")`

nileshgulia1:
Well @ksuess , Let's stick to your fix in Volto by not depending on this PR anymore 🙂

PR #1467, opened 2022-08-03
Use Plone latest in buildout

wesleybl:
For a PR to be accepted, it needs to pass Jenkins tests. In Jenkins, the buildout pins the newest versions. So Plone latest is "closer" to Jenkins, than a pinned version of Plone, which ends up being outdated for a long time. Testing with the most similar versions decreases the chances of the test passing on github and failing on Jenkins, and vice versa.

### DIFF
--- a/plone-6.0.x.cfg
+++ b/plone-6.0.x.cfg
@@ -1,9 +1,7 @@
 [buildout]
 extends =
-    https://dist.plone.org/release/6.0.0a4/versions.cfg
+    https://dist.plone.org/release/6.0-latest/versions.cfg
     base.cfg
-find-links = https://dist.plone.org/release/6.0.0a4/
-versions=versions
 
 [buildout:python37]
 parts =
@@ -25,3 +23,7 @@ cffi = 1.14.6
 # Requirement of flake8>=2.4.0: importlib-metadata<4.3
 # Error: The requirement ('importlib-metadata<4.3') is not allowed by your [versions] constraint (4.11.2)
 importlib-metadata = 4.2.0
+
+# Requirement of Markdown 3.4.1+: importlib-metadata>=4.4
+# Error: The requirement ('importlib-metadata>=4.4') is not allowed by your [versions] constraint (4.2.0)
+Markdown = 3.3.4
--- a/requirements-5.2.txt
+++ b/requirements-5.2.txt
@@ -1,5 +1,2 @@
-# Keep this file in sync with: https://github.com/kitconcept/buildout/edit/master/requirements.txt
-setuptools==42.0.2
-zc.buildout==2.13.4
-wheel
+-r https://dist.plone.org/release/5.2-latest/requirements.txt
 zpretty
--- a/plone-5.2.x.cfg
+++ b/plone-5.2.x.cfg
@@ -1,18 +1,10 @@
 [buildout]
 extends =
-    https://dist.plone.org/release/5.2.6/versions.cfg
+    https://dist.plone.org/release/5.2-latest/versions.cfg
     base.cfg
-find-links += https://dist.plone.org/thirdparty/
-versions=versions
 
 [versions]
 black = 21.7b0
-
-# Error: The requirement ('virtualenv>=20.0.35') is not allowed by your [versions] constraint (20.0.26)
-virtualenv = 20.0.35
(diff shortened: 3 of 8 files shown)


plone-6.0.x.cfg
```diff
@@ -25,3 +23,7 @@ cffi = 1.14.6
 # Requirement of flake8>=2.4.0: importlib-metadata<4.3
 # Error: The requirement ('importlib-metadata<4.3') is not allowed by your [versions] constraint (4.11.2)
 importlib-metadata = 4.2.0
+
+# Requirement of Markdown>=2.6.5: importlib-metadata>=4.4
+# Error: The requirement ('importlib-metadata>=4.4') is not allowed by your [versions] constraint (4.2.0)
+Markdown = 3.3.4
```
mauritsvanrees, line 29: This seems wrong:

- We pin `importlib-metadata = 4.2.0` a few lines above.
- The comment says `Markdown>=2.6.5` requires at least `importlib-metadata` 4.4.
- Yet we pin `Markdown` 3.3.4, which should conflict with the older `importlib-metadata`.

requirements-5.2.txt
```diff
@@ -1,5 +1 @@
-# Keep this file in sync with: https://github.com/kitconcept/buildout/edit/master/requirements.txt
-setuptools==42.0.2
-zc.buildout==2.13.4
-wheel
-zpretty
```
mauritsvanrees, line 5: Are you sure it is okay to drop the `zpretty` requirement?

mauritsvanrees (changes requested):
Generally I am in favour of automatically testing with the latest Plone. I know Timo is less a fan of this. But I would say it makes sense here because of what you say: this package is already being tested on Jenkins, so if all tests pass there, they are more likely to pass on latest than on a version from months ago.
So generally +1.

I added two inline comments though.

wesleybl:
>But I would say it makes sense here because of what you say: this package is already being tested on Jenkins, so if all tests pass there, they are more likely to pass on latest than on a version from months ago.

Exactly. If the package was not tested on Jenkins, it would make sense to pin the Plone version. I think it was fixed so that jobs don't fail for things that weren't done in PR. But Jenkins "does automatic updates". So it can happen on the side in Jenkins anyway. Making the two sides more similar makes more sense to me. The version pinned today in the `plone.restapi` buildout is 6.0.0a4 but we are already at 6.0.0b1.

plone-6.0.x.cfg
wesleybl: Newer versions of `flake8` fix `importlib-metadata<4.3` for Python < 3.8. See:

https://github.com/PyCQA/flake8/blob/5.0.4/setup.cfg#L45

So the pin that exists today:

`importlib-metadata = 4.2.0`

is necessary.

Newer versions of `Markdown` fix `importlib-metadata>=4.4` for Python < 3.10. See:

https://github.com/Python-Markdown/markdown/blob/3.4.1/setup.py#L77

This conflicts. So I pin an older version of `Markdown`, where the `importlib-metadata` constraint didn't exist yet.

requirements-5.2.txt
wesleybl: You're right. It is used in the [Makefile](https://github.com/plone/plone.restapi/blob/dfb88d317460c824fa9a99abbcc9320e2e40bc2b/Makefile#L107).

I will bring it back.

plone-6.0.x.cfg
mauritsvanrees: My point is that the comment suggests that Markdown 2.6.5 already requires the newer importlib-metadata, so I expect to get a version conflict here. I guess it is just buildout that reports this, and gets the 2.6.5 from somewhere. If you change the comment to "Requirement of Markdown 3.4.1+" then I would no longer be confused.

requirements-5.2.txt
wesleybl: Done

plone-6.0.x.cfg
wesleybl: Done

mauritsvanrees (approved):

wesleybl:
@tisto @ale-rt @jensens can you take a look here please?

Issue #1448, opened 2022-06-09
Serializer should use plone.app.uuid.utils.uuidToObject


mauritsvanrees:
Compare these two imports:

https://github.com/plone/plone.restapi/blob/b86c20cca15120be62b73ed24f944db22d189f44/src/plone/restapi/services/content/utils.py#L4

https://github.com/plone/plone.restapi/blob/b86c20cca15120be62b73ed24f944db22d189f44/src/plone/restapi/serializer/utils.py#L1

We should use the one from `plone.app.uuid` in both places, as I want to deprecate the one from `plone.outputfilters`. In fact, I [have already deprecated it](https://github.com/plone/plone.outputfilters/pull/49/commits/6ee71503c52c422a795fba74625b46f3f6a553df) as small cleanup in a larger PR, but will revert that for now.

Problem is that then two tests fail, because the `plone.app.uuid` one does a security check. Sample failure:

```
Error in test test_resolveuid_serialize_take_care_of_primary_fields (plone.restapi.tests.test_resolveuid.TestBlocksResolveUID)
Traceback (most recent call last):
  File "/Users/maurits/.pyenv/versions/3.10.1/lib/python3.10/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/Users/maurits/.pyenv/versions/3.10.1/lib/python3.10/unittest/case.py", line 591, in run
    self._callTestMethod(testMethod)
  File "/Users/maurits/.pyenv/versions/3.10.1/lib/python3.10/unittest/case.py", line 549, in _callTestMethod
    method()
  File "/Users/maurits/community/plone-coredev/6.0/src/plone.restapi/src/plone/restapi/tests/test_resolveuid.py", line 346, in test_resolveuid_serialize_take_care_of_primary_fields
    value = self.serialize("blocks", blocks)
  File "/Users/maurits/community/plone-coredev/6.0/src/plone.restapi/src/plone/restapi/tests/test_resolveuid.py", line 242, in serialize
    return serializer()
  File "/Users/maurits/community/plone-coredev/6.0/src/plone.restapi/src/plone/restapi/serializer/blocks.py", line 72, in __call__
    block_value = handler(block_value)
  File "/Users/maurits/community/plone-coredev/6.0/src/plone.restapi/src/plone/restapi/serializer/blocks.py", line 94, in __call__
    value[field] = uid_to_url(link)
[...]

PR #1471, opened 2022-08-21
Add unrestricted keyword argument to uuidToObject()

anirudhhkashyap:
Fixes: [#1448](https://github.com/plone/plone.restapi/issues/1448)

Changes:
- Pass unrestricted keyword argument to uuidToObject()
- The unrestricted argument value is set to True

Setting the unrestricted flag to True invokes the unrestrictedTraversal() method instead of the restrictedTraversal() method as shown [here](https://github.com/plone/plone.app.uuid/blob/65cdc16a81a9ed12453829fd3fd37d3aca5ac328/plone/app/uuid/utils.py#L94-L96). Earlier, invoking the latter threw the Unauthorized Exception. Now the access to the object can be controlled using the unrestricted flag. This will help in deprecating uuidToObject() from plone.outputfilters.browser.resolveuid.

### DIFF
--- a/src/plone/restapi/serializer/utils.py
+++ b/src/plone/restapi/serializer/utils.py
@@ -8,7 +8,7 @@ import re
 
 RESOLVEUID_RE = re.compile("^[./]*resolve[Uu]id/([^/]*)/?(.*)$")
 
-
+# Takes the resolveID URL and returns a URL to the actual object
 def uid_to_url(path):
     if not path:
         return ""
@@ -23,10 +23,10 @@ def uid_to_url(path):
     if suffix:
         href += "/" + suffix
     else:
-        """ Pass unrestricted flag as true so the object is accessible.
-        At uuidToObject(), this leads to unrestrictedTraverse() to be invoked instead of restrictedTraverse().
-        """
-        target_object = uuidToObject(uid, unrestricted = True)
+        # Pass unrestricted flag as true so the object is accessible.
+        # At uuidToObject(), this leads to unrestrictedTraverse() to be invoked instead of restrictedTraverse().
+        
+        target_object = uuidToObject(uid, unrestricted=True)
         if target_object:
             adapter = queryMultiAdapter(
                 (target_object, target_object.REQUEST), IObjectPrimaryFieldTarget
--- a/src/plone/restapi/serializer/utils.py
+++ b/src/plone/restapi/serializer/utils.py
@@ -1,4 +1,4 @@
-from plone.outputfilters.browser.resolveuid import uuidToObject
+from plone.app.uuid.utils import uuidToObject
 from plone.outputfilters.browser.resolveuid import uuidToURL
 from plone.restapi.interfaces import IObjectPrimaryFieldTarget
 from zope.component import queryMultiAdapter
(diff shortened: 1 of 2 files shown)


anirudhhkashyap:

src/plone/restapi/serializer/utils.py
```diff
@@ -23,7 +23,10 @@ def uid_to_url(path):
     if suffix:
         href += "/" + suffix
     else:
-        target_object = uuidToObject(uid)
+        """ Pass unrestricted flag as true so the object is accessible.
+        At uuidToObject(), this leads to unrestrictedTraverse() to be invoked instead of restrictedTraverse().
+        """
+        target_object = uuidToObject(uid, unrestricted = True)
```
jensens, line 29:
```suggestion
        target_object = uuidToObject(uid, unrestricted=True)
```

jensens:
ad `unrestricted`: Question is, why would we want to get objects we are not allowed to view? That sounds odd.

davisagli:
@jensens Maybe a user has permission to edit an item that includes references to a different item that they do not have permission to view. But, I agree that it is not obvious whether this is the right design for every use case, and might need a more careful analysis of everywhere this function is used.

jensens:
If the user has no `View` permission for the target, we need to check if the user has at least `Access content information`. If both are missing, we can verify the UUID (does target exist) and keep the UUID. But we are not able to display anything else than the plain UUID.

jensens:
I know took a deeper look here. The main problem is, the purpose of the existing code this PR changes is not very clear at the first look. The purpose is to fix the test here:

https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/tests/test_resolveuid.py#L221-L232

It check if an anonymous user can resolve a UUID to a non-published page. This is probably fine, it takes the resolveUID url and return an URL to the actual object. 

The test fails if the helper function touched in this PR is not modified. IMO the test is fine. Having an URL rendered is fine as long as when clicking on the link gives an unauthorized.

I propose - while at it - to add a docstring to the function explaining its purpose, fix the minor code style glitch and the merge the PR.

src/plone/restapi/serializer/utils.py
```diff
@@ -23,7 +23,10 @@ def uid_to_url(path):
     if suffix:
         href += "/" + suffix
     else:
-        target_object = uuidToObject(uid)
+        """ Pass unrestricted flag as true so the object is accessible.
+        At uuidToObject(), this leads to unrestrictedTraverse() to be invoked instead of restrictedTraverse().
+        """
```
jensens, line 28:
```suggestion
        # Pass unrestricted flag as true so the object is accessible.
        # At uuidToObject(), this leads to unrestrictedTraverse() to be invoked instead of restrictedTraverse().
```

anirudhhkashyap (commented):
@jensens Sorry for the lack of clarity. I'm kinda new to this project. Thanks for the feedback!

jensens (approved):

mauritsvanrees:
This needs your PR https://github.com/plone/plone.app.uuid/pull/14 which was merged, but is not in a release yet. And it won't be in Plone 5.2, with `plone.restapi` is still being tested against. So I cannot imagine that this works. Maybe this corner case is not caught by tests.

I would propose to leave `plone.restapi` unchanged for now. Instead fix https://github.com/plone/plone.outputfilters/issues/52
Once we drop 5.2 support in `plone.restapi`, we can revisit this.

wesleybl:
@anirudhhkashyap I reverted this PR at #1475, because it depends on a `plone.app.uuid` release. See:

https://github.com/plone/plone.restapi/pull/1471#issuecomment-1225594991

Can you please open another PR with these changes here? Then we can leave it waiting for a new release of `plone.app.uuid`.

anirudhhkashyap:
> @anirudhhkashyap I reverted this PR at #1475, because it depends on a `plone.app.uuid` release. See:
> 
> [#1471 (comment)](https://github.com/plone/plone.restapi/pull/1471#issuecomment-1225594991)
> 
> Can you please open another PR with these changes here? Then we can leave it waiting for a new release of `plone.app.uuid`.

Sure, no problem.

wesleybl:
@anirudhhkashyap and please, create a branch in plone.restapi instead of make in a fork.

PR #1475, opened 2022-08-26
Revert PR #1471

wesleybl:
PR #1471 depends on a version of `plone.app.uuid`, which has not been released yet. See:

https://github.com/plone/plone.restapi/pull/1471#issuecomment-1225594991

This fixes the broken tests.

Could not revert with git command.

### DIFF
--- a/news/1448.bugfix
+++ b/news/1448.bugfix
@@ -1,2 +1,0 @@
-Fix for test cases failing when deprecating uuidToObject() from `plone.outputfilters.browser.resolveuid`.
-[anirudhhkashyap]
--- a/src/plone/restapi/serializer/utils.py
+++ b/src/plone/restapi/serializer/utils.py
@@ -1,4 +1,4 @@
-from plone.app.uuid.utils import uuidToObject
+from plone.outputfilters.browser.resolveuid import uuidToObject
 from plone.outputfilters.browser.resolveuid import uuidToURL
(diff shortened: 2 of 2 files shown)


wesleybl:

jensens:
maybe its easier/better to ping @mauritsvanrees or @esteele to release it?

jensens:
And we should fix the test setup to pick this up.

wesleybl:

wesleybl:
@jensens  if I understand @mauritsvanrees ' comment correctly, a release of `plone.app.uuid` would only be for Plone 6. Here Plone 5.2 is tested as well. So it would need some backport. Maybe it's not that fast.

So I thought of merging this PR and creating another PR with the changes from #1471 . So the changes could wait for the PR, and the tests would turn green.

erral:
We need this, the tests are broken in master without this change.

davisagli (approved):
Let's go ahead and merge this to unbreak the 5.2 builds so that other changes are not blocked. Then we can figure out the right way to make the change without time pressure.

jensens:
A revert of the revert PR would be great

mauritsvanrees:
I have released [plone.app.uuid 2.2.0](https://pypi.org/project/plone.app.uuid/2.2.0/). So plone.restapi could check if this version is available, and then use it, calling `uuidToObject(uuid, unrestricted=True)`.

Issue #1476, opened 2022-08-30
TTW DX Layout disables IBlocks behavior and with it all the indexers and transformers


avoinea:
**Issue:** 

By enabling TTW DX Layout for content-types the **volto.blocks** behavior is removed in order to be able to edit **blocks** and **blocks_layout** defaults. The side effect of this is that also the **IBlocks** interface is removed from ctype / objects and with it all the registered IBlocks `indexers`, `serializers`, `deserializers`. 

e.g.: https://github.com/plone/plone.restapi/blob/9f205c1b9514d09e45c855899a820b3c703d49f3/src/plone/restapi/serializer/blocks.py#L141-L143

**Solution:**

A new marker interface behavior for TTW DX layout blocks:

```Python
class IBlocksEditableLayout(IBlocks):
    """ Volto Blocks Editable Layout marker interface """
```

PR #1477, opened 2022-08-30
Volto blocks editable layout behavior

avoinea:
See #1476


avoinea:

avoinea:
Looks like the Jenkins test are failing again, unrelated to this PR.

avoinea:

sneridagh:

sneridagh:
@davisagli @tisto could you please take a look? I think @avoinea needs it.

davisagli:
@sneridagh @avoinea It seems a little strange to add this in plone.restapi, since someone could be using plone.restapi without using Volto at all. (It's also strange to have the existing IBlocks behavior here, but I guess that ship already sailed.) Could it go in plone.volto instead?

On the other hand, I guess it doesn't really make a difference since plone.volto is now a dependency of Plone 6 and behaviors will show up whether or not the plone.volto profile has been installed.

I guess I'm okay going ahead with this since I don't have a better proposal at the moment.

jensens:
> On the other hand, I guess it doesn't really make a difference since plone.volto is now a dependency of Plone 6 and behaviors will show up whether or not the plone.volto profile has been installed.

This s not true, you can install CMFPlone without Volto, but with restapi. 

I propose to add behaviors related to volto in plone.volto. The blocks behavior should go there as well at some point in time, but not before 6.0 final is released.

avoinea:
@jensens @davisagli I do agree that the blocks related behaviors should go into `plone.volto` package. But until we move them there, I think they should stay together, here.

jensens:
I agree to keep this here for Plone 6.0. Thanks to behavior short-names its easy to move them around in future.

@avoinea btw., is there already an issue in either plone.volto or plone.restapi which covers this move out? If not would you mind to add one?

davisagli:
I am also okay with merging this now and moving them later.

tisto:
@jensens @davisagli when we implemented the IBlocks behavior there was no plone.volto package. We will have to move stuff over for Plone 6.1 or 7.

tisto:
@avoinea  Overall this LGTM. The thing I am wondering about is the naming. "Editable blocks" confuses me because all blocks in Volto are editable. I am not a native English speaker but isn't the functionality we are talking about rather a "template" or "page template"? So maybe "Blocks Page Template" or "Editable Page Template"? Just thinking out loud here. Options?

avoinea:
@tisto Actually it's called `Blocks (Editable Layout)`:
<pre>
        name="volto.blocks.editable.layout"
        title="Blocks (Editable Layout)"
</pre>

And I avoided to call it `editable blocks` because of the reason you mentioned. 

Now, as the functionality in ControlPanel is labeled `Layout` I wouldn't go into `Page templates` or something else :thinking: 

Best we can do: `Blocks (Editable Page Layout)`, but then: is this referring to the `Page`  content-type or the generic WebPage :smile: 

Naming is always a pain :smile:

tisto:

tisto (approved):
LGTM. Let's merge it now. We can always amend the wording via translations.

tisto:
Hmm. CI fails (https://github.com/plone/plone.restapi/actions/runs/3188037698/jobs/5200243129):

```
-Authorization: Basic YWRtaW46c2VjcmV0
+Authorization: Basic YWRtaW46Y29ycmVjdCBob3JzZSBiYXR0ZXJ5IHN0YXBsZQ==
```

sneridagh:
weird... nothing has changed in that regard.

@avoinea ?

avoinea:
@sneridagh Doesn't look related to this PR. Something may be changed on the Plone side.

tisto:
That's super odd. Are we pinning Plone versions and all dependencies in our builds?

sneridagh:
Right, we are using `-latest` in requirements. 

@mauritsvanrees do you recall a change in b3 that could trigger this change?

https://github.com/plone/plone.restapi/pull/1477#issuecomment-1268126898

It seems that the serialization of the Basic auth has changed... (it seems like a different hash type).

tisto:
How many times do I have to tell folks not to do this: https://github.com/plone/plone.restapi/blob/master/requirements-6.0.txt

sneridagh:
I think the last time @davisagli convinced you otherwise :)

tisto:
I created a PR to pin the versions properly: https://github.com/plone/plone.restapi/pull/1512

mauritsvanrees:
The minimum password length is now 8 characters instead of 5, and for example the default password in `plone.app.testing` was updated accordingly. Maybe that is it? Just a guess.

tisto:
Looks like that's the culprit:

https://github.com/plone/plone.app.testing/commit/8f93055c933abd9d9f9265d36232051cbab455e6

sneridagh:
@mauritsvanrees this breaks if you update Volto to b3:
https://github.com/plone/volto/actions/runs/3189141686/jobs/5202614190

That means that it might break projects that do not use the Cypress autologin from core :( 

I'd say it's quite a breaking change... at least from the testing environments out there. How about the old Robot tests that people might have in place? If they do not use the default auto login it will also break, right?

davisagli:
@mauritsvanrees @tisto @sneridagh Sorry, I didn't think about the fact that this would affect browser tests too.

It might be hard to avoid the breaking change, because increasing the minimum password length is an important security measure and I think that overrides the desire for backwards compatibility. But, I will look to see if there is some way to bypass the check for this password. And certainly we can use something more intuitive than "correct horse battery staple", which was me optimizing for compatibility with future increases of the minimum length, when I thought the password was only used within the Python tests.

In regard to the question about whether to test using -latest:
- I agree that builds should be repeatable, so we should use a specific version pin.
- However, this delays finding out when there is a breaking issue in the backend that only becomes obvious when testing the frontend against it, like this issue. Even when using -latest there is still a delay. To reduce the delay we need a CI build of the Plone backend that runs the frontend tests against it. This sound like something to mention in the "future" part of my talk on testing Plone next week...

wesleybl:
Guys, sorry for the noise. I was the one who updated to latest.

I think it would be good to discover problems with this as soon as possible. In my view, it was good to see that now. But this does not seem to be unanimous. The specific Plone version ends up getting out of date.

Sorry.

davisagli:
@wesleybl It's not your fault; some others of us also reviewed and approved the change, but my thinking has shifted since then. I actually think we want both: well specified versions in the repository, and something automated to test updating them as early as possible.

PR #1488, opened 2022-09-09
Validate the id in the content POST service

### LINKED ISSUE #1487: It's possible to set an invalid short name when adding content
If an `id` is specified while creating content via the [FolderPost](https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/services/content/add.py) endpoint, it is used without validating that it meets the criteria for a valid id. For example, it's possible to add an item with an id containing spaces.

What I would expect: either the id should be normalized (this is what already happens if it is set while editing the item later), or trying to add an item with an invalid id should return a 400 response with a helpful error message.

Note: This might be considered a backwards incompatible change, since someone calling the API may be assuming that the item has the id that was specified, and not a normalized form of it. But, it could also be considered a bugfix, because it never should have been possible to add items with an invalid id.

Thoughts on how to handle this, @tisto?

davisagli:
Fixes #1487

### DIFF
--- a/news/1487.bugfix
+++ b/news/1487.bugfix
@@ -1,0 +1,4 @@
+Use the name chooser to normalize the short name of content
+even when an `id` is specified explicitly in the content POST endpoint.
+This prevents setting the id to include invalid characters.
+[davisagli]
--- a/news/1487.bugfix
+++ b/news/1487.bugfix
@@ -1,4 +1,3 @@
-Use the name chooser to normalize the short name of content
-even when an `id` is specified explicitly in the content POST endpoint.
-This prevents setting the id to include invalid characters.
+When an `id` is specified explicitly in the content POST endpoint,
+return a 400 error response if it is invalid or unavailable.
 [davisagli]
--- a/src/plone/restapi/services/content/add.py
+++ b/src/plone/restapi/services/content/add.py
@@ -90,7 +90,11 @@ class FolderPost(Service):
         if not getattr(deserializer, "notifies_create", False):
             notify(ObjectCreatedEvent(obj))
 
-        obj = add(self.context, obj, rename=not bool(id_))
+        try:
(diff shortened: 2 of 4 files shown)


davisagli:

erral:
can you also add tests for special IDs?

For instance if you try to add a content titled 'layout' the created item is 'layout-1' or if you try to manually set an id with an space, the space is exchanged with a - in classic ui.

See https://6-classic.demo.plone.org/en/demo/layout-1

https://6-classic.demo.plone.org/en/demo/i-ill-try-to-add-a-whitespace-in-the-id-1

Or when adding a page titled 'subjects' it fails https://6-classic.demo.plone.org/en/demo/subjects/view

davisagli:
@erral Sure. This should take care of that too, but I'll add a test

davisagli:
@erral There was already a test for converting spaces to dashes when I first submitted the PR.

I added a test for layout --> layout-1 in https://github.com/plone/plone.restapi/pull/1488/commits/cce52eb33ea81508d95205f1a476dd80a3f118a4

The case with `subjects` is different; I think that needs to be fixed somewhere else since the classic UI is also allowing it. Possible in https://github.com/plone/plone.base/blob/main/src/plone/base/utils.py#L351

mauritsvanrees (commented):
Should be okay, but I have just voiced a doubt in the [issue](https://github.com/plone/plone.restapi/issues/1487#issuecomment-1242551621).

jensens (approved):

davisagli:
Based on the discussion in #1487, I decided that it's better to return an error response instead of normalizing the id, since the client explicitly requested a specific id and may not want to get something different.

davisagli:

tisto (approved):

tisto:

(seems the Github reporting of Jenkins is failing, all jobs pass except the last one)

PR #707, opened 2019-03-23
Add @userschema endpoint for getting the user schema.

sneridagh:
Related to #706

PR #1494, opened 2022-09-20
Use json_compatible when serializing users

### LINKED ISSUE #1493: User with datetime schema field is not serialized correctly
If you add a datetime field to a userschema, it is saved as a "DateTime" object, but if later you request to serialize that object you get a "Datetime object is not JSON serializable" error.

The point is that the @users endpoint doesn't convert the values to JSON compatible values using the existing json_compatible function.

erral:
Fixes #1493 

cc: @ionlizarazu

### DIFF
--- a/news/1493.bugfix
+++ b/news/1493.bugfix
@@ -1,0 +1,2 @@
+Use json_compatible when serializing users in @users endpoint
+[erral]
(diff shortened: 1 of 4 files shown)


erral:

erral:

erral:

sneridagh (approved):

sneridagh:
@erral this does relate to: https://github.com/plone/plone.restapi/pull/707?

erral:
It is related. We do not need it in the context of #707 but it will be useful there too.

We reached to the current issue when adding a DateTime field in the user schema.

PR #1498, opened 2022-09-29
Don't generate image scales until they are needed

davisagli:
When available, use the new `pre` flag on the scale method of the imaging view so that we can serialize metadata about image scales without actually generating the scales. Improves performance and consequently should also reduce the chance of conflict errors.

### DIFF
--- a/news/1498.feature
+++ b/news/1498.feature
@@ -1,0 +1,1 @@
+- Improve performance of serializing image scales. [davisagli]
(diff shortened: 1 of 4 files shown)


davisagli:

mauritsvanrees:
O, the same could be done in `get_original_image_url` in the same `imaging.py` file.

davisagli:
> O, the same could be done in get_original_image_url in the same imaging.py file.

good catch, actually I saw that too but then forgot to change it

davisagli:

PR #1499, opened 2022-09-29
Fix description field missing from form fieldset descriptions

JeffersonBledsoe:
This PR fixes the 'description' field being missing from fieldset queries.


tisto:
@JeffersonBledsoe good catch! Could you please rebase with master branch and add a changelog entry to "/news" instead of directly commit to CHANGES.md (we use towncrier in Plone core)?

JeffersonBledsoe:
@tisto Hey! Thanks for the changelog comment, I've updated that now. For green tests, do I need to manually update all of the `http-examples`?

tisto:
@JeffersonBledsoe CI currently fails because the docs are off (to expected):

https://github.com/plone/plone.restapi/actions/runs/3166413336/jobs/5156114487

You can update the docs automatically by running the test_documentation tests:

````
bin/test -t test_documentation
````

File: src/plone/restapi/tests/http-examples/translated_messages_types_folder.resp [hunk 1]
```diff
@@ -14,6 +14,7 @@ Content-Type: application/json+schema
         },
         {
             "behavior": "plone",
+            "description": "None",
```
davisagli, line 17: @JeffersonBledsoe This doesn't look right... If the field has no description, I would expect this to get serialized as `null` or an empty string, not a string containing "None"

You might need to do `translate(group.description, context=getRequest()) if group.description is not None else ""` below

davisagli (changes requested):

davisagli (approved):

src/plone/restapi/tests/http-examples/translated_messages_types_folder.resp
JeffersonBledsoe: Hey @davisagli, thanks for catching this! It didn't show up in my cursory local testing of the PR

davisagli (approved):

PR #1504, opened 2022-10-01
Plone 6 as first class citizen

sneridagh:
Update GHA to remove checks on Python 3.6 and added 3.10 for Plone 6. Removed 3.7 from Plone 6 as well.

The amendment in Makefile was needed because of 

https://community.plone.org/t/plone-6-0-0b2-released/15643

we were initially installing the 5.x buildout then on the top the Plone 6 one. This provoked the build to fail due that an initial outdated setuptools+zc.buildout was already in place. Now it only does it once, for each. Thus, the `Makefile` amendment.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -7,17 +7,15 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - python-version: '3.6'
-            plone-version: '5.2'
           - python-version: '3.7'
             plone-version: '5.2'
           - python-version: '3.8'
             plone-version: '5.2'
-          - python-version: '3.7'
-            plone-version: '6.0'
           - python-version: '3.8'
             plone-version: '6.0'
           - python-version: '3.9'
+            plone-version: '6.0'
+          - python-version: '3.10'
             plone-version: '6.0'
 
     steps:
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,8 @@ docs/.DS_Store
 docs/_build
 docs/plone.restapi.http-examples
 docs/source/README.rst
+docs/Makefile
+docs/make.bat
 sphinx_build.log
 htmlcov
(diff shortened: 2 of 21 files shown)


sneridagh:

sneridagh:

.github/workflows/tests.yml
```diff
@@ -7,18 +7,16 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - python-version: '3.6'
```
tisto, line 10: @sneridagh wondering if we can remove 3.6 here. If Plone 5.2 officially supports it, we should continue to test it here. Or am I missing something?

tisto (approved):
Apart from my question regarding dropping Python 3.6 from the Plone 5.2 LGTM.

sneridagh:
3.6 reached its EOL in Dec.2021

.github/workflows/tests.yml
tisto: "Supports Python 2.7, 3.6, 3.7, 3.8. Note that both Python 2.7 and 3.6 have reached end of life, and will get harder to test and support."

https://plone.org/download/release-schedule#:~:text=Current%20release%3A%205.2.,%2C%203.6%2C%203.7%2C%203.8.

sneridagh: In that case ok, wondering if we have to review that policy given the current expectative and the fact that Python is releasing a major every year.

mauritsvanrees: Plone 5.2 has dropped 3.6 support. See my announcement earlier this week: https://community.plone.org/t/plone-5-2-drops-python-3-6-support/15706
I have updated the text on the release schedule now.

davisagli (commented):
@sneridagh I didn't have a chance to look at this until now, but it looks perfect. Thanks!

PR #1515, opened 2022-10-06
Cache buildout eggs between builds

### LINKED ISSUE #1514: Add buildout eggs cache to Github Actions configuration
The GHA tests spend quite a long time running the buildout. It should be possible to improve this by making sure the buildout eggs eggs directory is cached between builds

davisagli:
Fixes #1514

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -33,7 +33,10 @@ jobs:
       - uses: actions/cache@v3
         with:
           path: eggs
-          key: ${{ runner.os }}-py${{ matrix.python-version }}-plone${{ matrix.plone-version }}-buildout-
+          key: ${{ runner.os }}-py${{ matrix.python-version }}-plone${{ matrix.plone-version }}-${{ hashFiles('*.cfg') }}
+          restore-keys: |
+            ${{ runner.os }}-py${{ matrix.python-version }}-plone${{ matrix.plone-version }}-
+            ${{ runner.os }}-py${{ matrix.python-version }}-
 
       # python install
       - run: pip install virtualenv
@@ -43,7 +46,7 @@ jobs:
 
       # buildout
       - name: buildout
-        run: buildout -t 10 -c plone-${{ matrix.plone-version }}.x.cfg code-analysis:return-status-codes=True
+        run: buildout -t 10 -c plone-${{ matrix.plone-version }}.x.cfg
         env:
           CI: true
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -30,7 +30,7 @@ jobs:
           cache: 'pip'
 
       # buildout eggs cache
-      - uses: actions/cache@v1
+      - uses: actions/cache@v3
         with:
           path: eggs
           key: ${{ runner.os }}-py${{ matrix.python-version }}-plone${{ matrix.plone-version }}-buildout-
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -33,7 +33,7 @@ jobs:
       - uses: actions/cache@v1
         with:
           path: ~/eggs
-          key: ${{ runner.os }}-buildout-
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -24,17 +24,16 @@ jobs:
...
-      - uses: actions/cache@v1
+      # buildout eggs cache
+      - uses: actions/cache@v3
         with:
-          path: ~/.cache/pip
-          key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
-          restore-keys: |
-            ${{ runner.os }}-pip-
+          path: eggs
+          key: ${{ runner.os }}-py${{ matrix.python-version }}-plone${{ matrix.plone-version }}-buildout-
```
mauritsvanrees, line 36: This is close to what we use in [coredev](https://github.com/plone/buildout.coredev/blob/6.0/.github/workflows/run-buildout.yml). I would add some restore-keys though. If I have it correctly in my head, if the key matches exactly then it is restored. Otherwise GitHub looks for the closest matching restore key. Here I guess only one makes sense:

```
restore-keys: |
    ${{ runner.os }}-py${{ matrix.python-version }}-
```

This should mean that when GitHub actions runs the 6.0 tests and it does not find the exact key, say `linux-3.8-plone6.0-buildout-` then it looks for keys starting with `linux-3.8-` and it might find `linux-3.8-plone5.2-buildout-`. Then you get an eggs directory with eggs for 5.2. At a guess, about one fourth of these can be reused on 6.0, so running buildout takes roughly one fourth less time.

When I think about it, I expect that when all goes well in such a run, at the end a new cache gets stored and this then contains all 6.0 eggs from the new buildout run, plus the previous 5.2 eggs. So the cache is bigger than it has to be. This may or may not be what you want. I would say it is okay.

mauritsvanrees (approved):
LGTM. One inline comment, feel free to pick that up or not.

.github/workflows/tests.yml
davisagli: @mauritsvanrees Thanks for the suggestion. I thought about this, but I figured it's not a big deal if there is a cache miss and everything is downloaded again from PyPI. After all, that's what we've been doing on every build so far.

I'm more worried about what happens when there is a minor Plone version bump, which does not currently change the cache key. Then there will be a cache hit, but with older versions. That's okay, because buildout will download the new ones, but because there was a cache hit, the cache action will not update the cache. So it will be slower again on every build.

To avoid that I guess we would need to put the complete Plone version somewhere it can be easily extracted to use in the cache key. Probably not that hard but I didn't spend time trying to solve it yet.

mauritsvanrees: Good point. In that case adding `${{ hashFiles('versions*.cfg') }}` to the key should help. That hash will change for every Plone version because we explicitly specify the Plone version in it again. So:

```
key: ${{ runner.os }}-py${{ matrix.python-version }}-plone${{ matrix.plone-version }}-buildout-${{ hashFiles('versions*.cfg') }}-
restore-keys: |
    ${{ runner.os }}-py${{ matrix.python-version }}-plone${{ matrix.plone-version }}-buildout-
    ${{ runner.os }}-py${{ matrix.python-version }}-
```

davisagli: @mauritsvanrees yeah, that's the right idea, but in this repo versions.cfg does not include the Plone versions. That's done in plone-6.0.x.cfg or plone-5.2.x.cfg. So it depends which buildout configuration is actually run...

mauritsvanrees: Sorry, I meant to say `hashFiles('*.cfg')`, or maybe `hashFiles('plone*.cfg')` could also work. This does mean that a change in `plone-5.2.x.cfg` causes a needless exact cache miss in the Plone 6.0 job. But the first `restore-key` will still match.

davisagli: @mauritsvanrees That sounds like it might work; I'll give it a try.

PR #1516, opened 2022-10-06
Fix pinns

wesleybl:
- Use specific version of Plone in requirements.txt. As we are using specific version in versions.cfg, we need to use
specific versions in requirements.txt as well, to avoid problems in the future. Ref: #1512

- Remove unnecessary pins.

### DIFF
--- a/news/1516.internal
+++ b/news/1516.internal
@@ -1,0 +1,1 @@
+Use specific version of Plone in requirements.txt. Remove unnecessary pins. [wesleybl]


davisagli (approved):
@wesleybl thanks, good catch

wesleybl:
> @wesleybl can you add a changelog in news/1516.internal?

@davisagli Is this really necessary? I don't see that this information would be useful for using plone.restapi. Wouldn't that just be a "change polluter"?

davisagli:
@wesleybl That's why it goes in the internal section, to keep it separate. I see your point, but currently there is a check for a changelog entry on each PR.

wesleybl:
@davisagli done.

tisto:
Hey @wesleybl, thanks a lot for taking care of this! And sorry for my previous comment. I highly value your contributions and it was certainly not your fault that we mixed up things in the repo. Turned out we all had different ideas of what we consider to be best practices. We will try to figure this out together and come up with a clear recommendation for how to handle things.

PR #1517, opened 2022-10-06
Remove code-analysis from Makefile

wesleybl:
`plone.recipe.codeanalysis` was removed from buildout. The `importlib-metadata` pin was needed only because of code-analysis. So we're removing the pin.

Ref: #1507

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,14 +96,8 @@ test-performance-locust-querystring-search:
 test-performance-locust-querystring-search-ci:
 	bin/locust -f performance/querystring-search.py --host http://localhost:12345/Plone --users 100 --spawn-rate 5 --run-time 5m --headless --csv=example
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -98,12 +98,10 @@ test-performance-locust-querystring-search-ci:
 
 .PHONY: Code Analysis
 code-analysis:  ## Code Analysis
```
davisagli, line 100: @wesleybl Is there any reason to keep this command in the Makefile? We can probably remove it entirely.

wesleybl: I think whoever put the make command was not to use it in the CI. I even think that make was never used in CI. I think it was for local execution, without changing the files. But I will remove it then.

jensens (approved):

PR #1520, opened 2022-10-18
Fix test_statictime for p.a.discussion comments

instification:
Needed for https://github.com/plone/plone.app.discussion/pull/204

Tests won't pass until https://github.com/plone/plone.app.discussion/pull/204 is merged as it is a co-dependent change.

Updating p.a.discussion to set timezone aware times.

The current test is checking for a timezone, but there isn't one so the following code wasn't getting executed:

```
if pydt.tzinfo is not None:
    now = pydt.tzinfo.localize(now)
```

but `pydt.tzinfo.localize` is not a function so I have removed this check.

As the times coming from p.a.discussion will be tz aware, this update fixes the tests to compare using timezones.

### DIFF
--- a/src/plone/restapi/tests/test_statictime.py
+++ b/src/plone/restapi/tests/test_statictime.py
@@ -3,6 +3,8 @@ from DateTime import DateTime
 from datetime import timedelta
 from dateutil import tz
 from operator import itemgetter
+from pkg_resources import get_distribution
+from pkg_resources import parse_version
 from plone import api
 from plone.app.discussion.interfaces import IConversation
 from plone.app.discussion.interfaces import IDiscussionSettings
@@ -27,6 +29,11 @@ from plone.restapi.serializer.converters import json_compatible
 import transaction
 import unittest
 
+padiscussion_version = get_distribution("plone.app.discussion").version
+if parse_version(padiscussion_version) > parse_version("4.0.0b4"):
+    NEW_PADISCUSSION = True
+else:
+    NEW_PADISCUSSION = False
 
 class TestStaticTime(unittest.TestCase):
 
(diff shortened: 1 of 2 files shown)


davisagli:
@instification Is it possible to make this work both with and without the change in plone.app.discussion? The challenge is that plone.restapi still supports both Plone 5.2 and Plone 6.0 from a single release branch, but the change in plone.app.discussion to use timezone-aware datetimes is for Plone 6 only.

src/plone/restapi/tests/test_statictime.py
```diff
@@ -25,6 +29,11 @@
 import transaction
 import unittest
 
+padiscussion_version = get_distribution("plone.app.discussion").version
+if parse_version(padiscussion_version) > parse_version("4.0.0b4"):
```
instification, line 33: @davisagli I'm guessing 4.x is for Plone 6? Maybe this could just be `>=4.0.0`?

src/plone/restapi/tests/test_statictime.py
```diff
@@ -25,6 +29,12 @@
 import transaction
 import unittest
 
+padiscussion_version = get_distribution("plone.app.discussion").version
+if parse_version(padiscussion_version) >= parse_version("4.0.0b4.dev1"):
+    NEW_PADISCUSSION = True
+else:
+    NEW_PADISCUSSION = False
```
davisagli, line 36: @instification In general, it's better to test for the presence of some feature rather than checking versions. I also have some suspicions about the effects of using get_distribution on import performance. What about something like this?
```
import plone.app.discussion.browser.comment
NEW_PADISCUSSION = hasattr(plone.app.discussion.browser.comment, "localized_now")
```

(Maybe even call the variable something more meaningful like HAS_TZ_AWARE_COMMENTS)

davisagli (approved):

PR #1481, opened 2022-08-31
Add @portrait endpoint

sneridagh:
Fixes: #1480 and addresses https://github.com/plone/volto/pull/3487 (the /portal_memberdata/portrait traversal is not working with ++api++).

Let me know what do you think.

/cc @jmevissen

PR #1466, opened 2022-08-02
Use _getsafeMemberId when checking for user portrait

instification:
This is a very minor PR.

Currently when checking for user portrait it checks to see if the user is in `memberdata.portraits` however it is not sanitising the user id, meaning some users don't get their portrait returned even if it exists. For example if there is a hyphen in the user id.

PR #1524, opened 2022-10-19
Fix portrait endpoint for email address user id's

instification:
The previous pr #1466 resolved an issue where the endpoint was returning unsanitized user id's.

Following #1481 the new @portrait endpoint does not require sanitized user id's so when email login is enabled, portraits break.

The solution would either be to revert the change made in #1466 and return the raw user id, or to update the @portrait endpoint to handle sanitized id's.

This PR does the latter as well as:

 - Updating the endpoint for discussion/comments to use the new url
 - Updates the tests
   - Uses a Plone user instead of the zmi admin user, as this causes issues rendering the portrait
   - Adds a tests for users where login by email is enabled
   - Updates comments tests

### DIFF
--- a/src/plone/restapi/services/users/get.py
+++ b/src/plone/restapi/services/users/get.py
@@ -26,12 +26,13 @@ def getPortraitUrl(user):
     if not user:
         return
     portal = getSite()
-    portal_membership = getToolByName(portal, "portal_membership")    
+    portal_membership = getToolByName(portal, "portal_membership")
     portrait = portal_membership.getPersonalPortrait(user.id)
     if portrait and not isDefaultPortrait(portrait):
         safe_id = portal_membership._getSafeMemberId(user.id)
         return f"{portal.absolute_url()}/@portrait/{safe_id}"
     return
+
 
 def isDefaultPortrait(value):
     portal = getSite()
@@ -238,14 +239,11 @@ class PortraitGet(Service):
             user = decleanId(self.params[0])
             portrait = self.portal_membership.getPersonalPortrait(user)
         elif len(self.params) == 0:
-            current_user_id = \
-                self.portal_membership.getAuthenticatedMember().getId()
-            portrait = \
-                self.portal_membership.getPersonalPortrait(current_user_id)
+            current_user_id = self.portal_membership.getAuthenticatedMember().getId()
+            portrait = self.portal_membership.getPersonalPortrait(current_user_id)
         else:
             raise Exception(
-                "Must supply exactly zero (own portrait) or one parameter "
-                "(user id)"
+                "Must supply exactly zero (own portrait) or one parameter " "(user id)"
             )
         # User uploaded portraits have a meta_type of "Image"
         if not portrait or isDefaultPortrait(portrait):
--- a/src/plone/restapi/services/users/get.py
+++ b/src/plone/restapi/services/users/get.py
@@ -9,6 +9,7 @@ from plone.restapi.services import Service
 from Products.CMFCore.utils import getToolByName
 from Products.CMFPlone.utils import normalizeString
 from Products.PlonePAS.tools.membership import default_portrait
+from Products.PlonePAS.utils import decleanId
 from urllib.parse import parse_qs
 from zExceptions import BadRequest
 from zope.component import getMultiAdapter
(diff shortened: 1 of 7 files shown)


instification:

tisto (approved):
LGTM. @instification thanks for the fix!

tisto:

src/plone/restapi/services/users/get.py
```diff
@@ -222,7 +223,12 @@ def _get_user_id(self):
     def render(self):
         if len(self.params) == 1:
             # Retrieve the user portrait
-            portrait = self.portal_membership.getPersonalPortrait(self.params[0])
```
jensens, line 225: Here be dragons.

There is a long (>15y) history of patching `MembershipTool.getPersonalPortrait` in several addons dealing with it. Membrane did/does, pas.plugins.ldap as well, probably others too.

Before you merge this please check if this works with a recent version of `pas.plugins.ldap` or similar (should, but better safe than sorry).  I am also not sure if the meta_type assumption  always holds.

instification: So if I understand your comment @jensens you're suggesting it would be dangerous to not use `getPersonalPortrait`?

Because other addon's may patch it to give a different result, so by not using it we could be missing out on those changes?

I can see in pas.plugins.ldap it is patching it to basically solve the issue we are seeing here: https://github.com/collective/pas.plugins.ldap/blob/e1f8da6ab300bed50c779febac9036930b2cb91f/src/pas/plugins/ldap/monkey.py#L58-L86

Unfortunately, `portal_membership.getPersonalPortrait` doesn't return a portrait if the user id is already sanitised.

Perhaps the better solution would be to 'de-sanitize' the user id before passing it to `getPersonalPortrait`?

jensens: Exact, yes. And I am not sure if the meta_type is always the one you expect.

instification: ok thanks for the feedback. I'll take another look at the meta_type check, I'm sure there is a better way to determine if the users portrait is unset.

instification: @jensens I have refactored the code.

So now both the code that returns the portrait url and fetches the portrait will use getPersonalPortrait.

When fetching from the @portrait endpoint, it is expecting a 'safe id' so is calling `decleanId` from pas before calling `getPersonalPortrait`.

I don't really have an effective setup to test pas.ldap (and certainly not with custom portrait code), but as we are using the `getPersonalPortrait` function as intended I can't see it being an issue there due to this change.

instification:
@sneridagh I have also updated the endpoints/tests for comments as they were returning the old portrait url, which was causing it to break in volto. I can refactor the change into a new PR if you think it is better to do it that way

instification:

instification:

src/plone/restapi/services/users/get.py
jensens: Yes, this will do it. Great work. It is always good to see others getting their fingers dirty with our membership code ;-)

jensens (approved):

instification:
I noticed in local testing that the `isDefaultPortrait` function was always returning true, even when a users portrait wasn't set.

For some reason `getattr(portal, default_portrait, None)` always returns None, but `portal.unrestrictedTraverse(default_portrait, None)` does return the root portrait. I guess at some point in the past this was changed from being an object in zope to something else.

Additionally, even though `portal_membership.getPersonalPortrait('admin`)` returns `<FSImage /Plone/defaultUser.png>` and the traversed default portrait is also `<FSImage /Plone/defaultUser.png>` they don't compare. So I updated the function to compare on `.getPhysicalPath()` which does match for some reason.

I added a test to check for a 404 from the endpoint when the portrait isn't set.

instification:

src/plone/restapi/tests/test_services_users.py
```diff
@@ -1011,6 +1014,42 @@ def test_get_user_portrait_anonymous(self):
...
+        portrait = urlre.group(1)
+
+        response = self.api_session.get(
+            f"/@portrait/{portrait}",
+        )
+
+        self.assertEqual(200, response.status_code)
+        self.assertEqual(response.headers["Content-Type"], "image/gif")
+
+    def test_get_user_defalut_portrait(self):
```
davisagli, line 1047:
```suggestion
    def test_get_user_default_portrait(self):
```

PR #1531, opened 2022-11-07
fix time to be returned with a timezone specifier

### LINKED ISSUE #1530: time is returned without timezone specifier, causing the client to not use local time
The api returns time information of the entries without a timezone specifier.  See https://stackoverflow.com/questions/71024015/how-to-get-timezone-from-iso-time-string-in-python for a deeper explanation.

The lack of the timezone specifier causes Intl (and derivative libraries such as  the `FormattedDate` component  of `react-intl`) in the browser not to use local time, which is disturbing for the user who accepts local time. Thus this is considered a bug in applications. 

To solve the problem we can add the timezone specifier to the end of the datetime string. Since the python timestamp does not contain the information about the originating timezone itself, this is always UTC and we can simply add Z to the end in all cases. This ensures that the date will be interpreted properly by the client.

reebalazs:
Fixes #1530

The api returns time information of the entries without a timezone specifier. See https://stackoverflow.com/questions/71024015/how-to-get-timezone-from-iso-time-string-in-python for a deeper explanation.

The lack of the timezone specifier causes Intl (and derivative libraries such as the FormattedDate component of react-intl) in the browser not to use local time, which is disturbing for the user who accepts local time. Thus this is considered a bug in applications.

To solve the problem we can add the timezone specifier to the end of the datetime string. Since the python timestamp does not contain the information about the originating timezone itself, this is always UTC and we can simply add Z to the end in all cases. This ensures that the date will be interpreted properly by the client.

### DIFF
--- a/news/1530.bugfix
+++ b/news/1530.bugfix
@@ -1,0 +1,1 @@
+fix time to be returned with a timezone specifier @reebalazs
--- a/news/1530.bugfix
+++ b/news/1530.bugfix
@@ -1,1 +1,1 @@
-fix time to be returned with a timezone specifier @reebalazs
+Fix time to be returned with a timezone specifier in history endpoint @reebalazs
--- a/src/plone/restapi/services/history/get.py
+++ b/src/plone/restapi/services/history/get.py
@@ -71,7 +71,13 @@ class HistoryGet(Service):
             # Versioning entries use a timestamp,
             # workflow ISO formatted string
             if not isinstance(item["time"], str):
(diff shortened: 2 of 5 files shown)


sneridagh (approved):

reebalazs:

news/1530.bugfix
```diff
@@ -0,0 +1 @@
+fix time to be returned with a timezone specifier @reebalazs
```
davisagli, line 1:
```suggestion
Fix time to be returned with a timezone specifier in history endpoint @reebalazs
```

davisagli (approved):

PR #1535, opened 2022-11-14
Fortify user api against missing user data (None)

reebalazs:
Real project produced user data None (possibly via ldap plugin) which breaks the api. As we never would want to display such user data, we fortify the api and filter None.

Note this cannot normally happen via the restapi, so it's untestable with a traditional functional test. (Unittests provided.)

### DIFF
--- a/src/plone/restapi/services/users/get.py
+++ b/src/plone/restapi/services/users/get.py
@@ -17,6 +17,8 @@ from zope.component import queryMultiAdapter
 from zope.component.hooks import getSite
 from zope.interface import implementer
 from zope.publisher.interfaces import IPublishTraverse
+from Products.PlonePAS.tools.memberdata import MemberData
+from typing import Sequence, Iterable
 
 DEFAULT_SEARCH_RESULTS_LIMIT = 25
 
@@ -68,14 +70,11 @@ class UsersGet(Service):
         return self.portal_membership.getMemberById(user_id)
 
     @staticmethod
-    def _sort_users(users):
-        """users is a list of user objects or None"""
-        users.sort(
-            key=lambda x: normalizeString(x.getProperty("fullname", ""))
-            if x is not None
-            else ""
+    def _sort_users(users: Iterable[MemberData]) -> Sequence[MemberData]:
+        """users is an iterable of MemberData objects, None is not accepted"""
+        return sorted(
+            users, key=lambda x: normalizeString(x.getProperty("fullname", ""))
         )
-        return users
 
     def _principal_search_results(
         self, search_for_principal, get_principal_by_id, principal_type, id_key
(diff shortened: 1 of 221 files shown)


src/plone/restapi/services/users/get.py
```diff
@@ -92,6 +93,8 @@ def _principal_search_results(
     def _get_users(self, **kw):
         results = {user["userid"] for user in self.acl_users.searchUsers(**kw)}
         users = [self.portal_membership.getMemberById(userid) for userid in results]
+        # Filtering for None which might happen due to some unknown condition
+        users = users.filter(lambda x: x is not None)
```
jensens, line 97: I would use generators here and above in l95 in order to reduce the memory footprint if many users are requested.

src/plone/restapi/services/users/get.py
```diff
@@ -71,8 +71,9 @@ def _get_user(self, user_id):
     @staticmethod
     def _sort_users(users):
         users.sort(
-            key=lambda x: x is not None
-            and normalizeString(x.getProperty("fullname", ""))
+            key=lambda x: normalizeString(x.getProperty("fullname", ""))
+            if x is not None
```
jensens, line 75: This must never happen if L97 was running before. I would skip the test here. To clarify what is expected we could add  a type hint and some doc-string to the function.

jensens (changes requested):

jensens:
> Note this cannot normally happen via the restapi, so it's untestable with a traditional functional test.

On unit test level it is perfectly testable.

src/plone/restapi/services/users/get.py
reebalazs: ok in l95 but not good in l96 because the result will get in-place sorted, which is not possible with generators.

jensens: You could use `sorted()` instead of in place sort and pass generators all way long until the result is serialized to json.

src/plone/restapi/services/users/get.py
reebalazs: Done, I added a docstring but not a type hint as I am not sure how to express "class instance but not None" and not sure if we can assert the actual class of the input.

reebalazs:

jensens (approved):

mauritsvanrees (dismissed):
LGTM, but the changes to test_documentation and http-examples should not be needed. I believe this is a known issue, where we sometimes get different results, so a failure in 'test no uncommitted changes' is apparently okay for the moment, until someone finds time to fix it.

tisto:
@mauritsvanrees this is indeed a known issue and I would love if someone could look into how we could fix this. We would have to come up with a "freeze" for the version that is exposed in the HTTP examples.

tisto (approved):

tisto:
@reebalazs can you rebase with master? Seems the HTTP examples are off.

reebalazs:

reebalazs:
@tisto, rebase done.

PR #1544, opened 2022-11-27
Mock addon versions in documentation tests

### LINKED ISSUE #1538: Automatically update version in translated_messages_addons.resp after release.
The version number in `http-examples/translated_messages_addons.resp` must be updated after every release, otherwise the 'no uncommitted changes' test will fail the next time someone runs the tests or makes a PR.

This should be possible with a [`zest.releaser` entry point](https://zestreleaser.readthedocs.io/en/latest/entrypoints.html) specific for `plone.restapi`. An example of such an entry point is in `i18ndude` for updating the output of the command line help in the documentation. The parts are:

- [`setup.py` registration](https://github.com/collective/i18ndude/blob/5.5.0/setup.py#L62-L65)
- [`setup.cfg` registration](https://github.com/collective/i18ndude/blob/5.5.0/setup.cfg#L1-L3)
- [the code for the hook](https://github.com/collective/i18ndude/blob/5.5.0/src/i18ndude/utils.py#L190)

I guess it only needs to be in one of `setup.py` or `setup.cfg`, not both. But it has been a while since I did this, so I am not sure. Having it in `setup.cfg` should mean that you get an error when the hook is for some reason not found. This may or may not be what you want. Having it in `setup.py` should mean it is available globally whenever you release any package, so you need to take care to only have it really do something when releasing `plone.restapi`, but this is an easy check to add.

The best hook would probably be `zest.releaser.postreleaser.middle`.

erral:
This way we do not have to update plone.restapi's version in documentation tests each time a new release is created.

Fixes #1538

### DIFF
--- a/src/plone/restapi/tests/helpers.py
+++ b/src/plone/restapi/tests/helpers.py
@@ -3,6 +3,7 @@ from plone.scale import storage
 from Products.CMFCore.utils import getToolByName
 from unittest.mock import patch
 from urllib.parse import urlparse
+from plone.restapi.services.addons.addons import Addons
 
 import quopri
 
@@ -73,3 +74,14 @@ def patch_scale_uuid(value):
             storage.AnnotationStorage, "_modified_since", return_value=True
         ):
             yield
+
+
+@contextmanager
+def patch_addon_versions(value):
+    """Patch the method that extracts the version of each of the addons.
+    Without this patch we get version errors each time an addon is upgraded
+    which is really annoying when a new version of plone.restapi is released
+    because all PRs need to be updated
+    """
+    with patch.object(Addons, "get_product_version", return_value=value):
+        yield
--- a/src/plone/restapi/tests/http-examples/translated_messages_addons.resp
+++ b/src/plone/restapi/tests/http-examples/translated_messages_addons.resp
@@ -13,7 +13,7 @@ Content-Type: application/json
             "title": "Soporte de Pol\u00edtica de Flujo de trabajo (CMFPlacefulWorkflow)",
             "uninstall_profile_id": "Products.CMFPlacefulWorkflow:uninstall",
             "upgrade_info": {},
-            "version": "3.0.0b2"
+            "version": "1.2.3"
         },
         {
             "@id": "http://localhost:55001/plone/@addons/collective.MockMailHost",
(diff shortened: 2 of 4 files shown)


erral:

erral:

src/plone/restapi/tests/http-examples/translated_messages_addons.resp
```diff
@@ -28,10 +28,10 @@ Content-Type: application/json
                 "available": false,
                 "hasProfile": true,
                 "installedVersion": "2.0.0",
-                "newVersion": "2.0.0",
-                "required": false
+                "newVersion": "1.2.3",
+                "required": true
```
davisagli, line 32: Why did this change the value of `required`?

davisagli (approved):
Thanks for working on this!

src/plone/restapi/tests/http-examples/translated_messages_addons.resp
erral: it looks like it's because the values of `installedVersion` and `newVersion` are different now that I patch the current version...

erral: I have reworked the patch and patched also the `upgrade_info` dict.

The code there does many things and it is hard to patch just the version numbers to get the `required` value to be false.

davisagli: @erral I'd rather not replace the `upgrade_info` with an empty dict like that. It starts to make the response less useful where it appears in documentation, because it's not realistic. Probably better to keep it as it was before.

src/plone/restapi/tests/helpers.py
```diff
@@ -3,6 +3,8 @@
 from Products.CMFCore.utils import getToolByName
 from unittest.mock import patch
 from urllib.parse import urlparse
+from plone.restapi.services.addons.addons import Addons
+from Products.GenericSetup.tool import SetupTool
```
jensens, line 7: unused according to flake8 
```
src/plone/restapi/tests/helpers.py:7:1: F401 'Products.GenericSetup.tool.SetupTool' imported but unused
```

src/plone/restapi/tests/http-examples/translated_messages_addons.resp
erral: I have reverted the changes while I check what can I do with this.

tisto:

tisto (approved):
LGTM. @erral thanks for working on this. You rock!

PR #1465, opened 2022-07-29
@site and @navroot endpoints

erral:
Fixes #1462

PR #1547, opened 2022-11-29
Add "UID" to content type endpoint response of "Plone Site"

ksuess:


ksuess:

src/plone/restapi/tests/test_documentation.py
```diff
@@ -177,6 +177,8 @@ def setUp(self):
         self.portal = self.layer["portal"]
         self.portal_url = self.portal.absolute_url()
 
+        setattr(self.portal, "_plone.uuid", "SomeUUIDPortal")
```
davisagli, line 180: I would generate an actual uuid (once, and then include it as a literal here in the test) so that the documentation shows the expected format.

davisagli (approved):
One small suggestion, but I'm also fine merging it as is.

src/plone/restapi/tests/test_documentation.py
ksuess: Sounds reasonable.

ksuess:

davisagli (approved):

tisto (approved):

erral:
I know that I come late to the party but I think we need to revisit the way we create static UUID in tests.

Right now we are using a [Static UUID generator utility](https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/testing.py#L341-L344) that we register as a global utility and [is pushed to the test_documentation tests on `setUp`](https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/tests/test_documentation.py#L171-L184)

During the development of #1465 I found that that registration happens in some cases too late such as when testing multilingual sites, because the site is prepared for multilingual *before* the utility registration, and thus the LRFs used do not have predictable UUIDs and everything is messed up when connecting translations (because remember that translations are linked using UUIDs). You can find some more information about it [in this comment](https://github.com/plone/plone.restapi/pull/1465#issuecomment-1248205293)

I finally reverted those changes and work around those tests in some other way. 

I don't think that setting a UUID manually is the solution in the long run. We should properly figure out how to produce stable and static UUID through all the testing process for the documentation to be predictable.

Just my 2 cents.

davisagli:
@erral That sounds like the utility needs to be registered in the test layer before the site is created -- and maybe put the utility's counter in the ZODB so it gets rolled back whenever other changes do during test teardown.

PR #1238, opened 2021-10-06
move translations into plone.app.locales

erral:
Fixes #841 

Merge together: 
- https://github.com/collective/plone.app.locales/pull/332
- https://github.com/plone/buildout.coredev/pull/743

PR #1548, opened 2022-12-01
Internationalization of @users endpoint error messages

wesleybl:
The `plone.restapi` domain was used.

Fixes https://github.com/plone/volto/issues/3958

### DIFF
--- a/src/plone/restapi/services/users/add.py
+++ b/src/plone/restapi/services/users/add.py
@@ -1,4 +1,6 @@
 from AccessControl import getSecurityManager
+from plone.app.users.schema import ICombinedRegisterSchema
+from plone.restapi import _
 from plone.restapi.deserializer import json_body
 from plone.restapi.interfaces import ISerializeToJson
 from plone.restapi.services import Service
@@ -13,6 +15,7 @@ from zope.component import getAdapter
 from zope.component import getMultiAdapter
 from zope.component import queryMultiAdapter
 from zope.component.hooks import getSite
+from zope.i18n import translate
 from zope.interface import alsoProvides
 from zope.interface import implementer
 from zope.publisher.interfaces import IPublishTraverse
@@ -32,6 +35,22 @@ class UsersPost(Service):
         # Consume any path segments after /@users as parameters
         self.params.append(name)
         return self
+
+    def translate(self, msgid):
+        return translate(
+            msgid,
+            context=self.request,
+        )
+
+    def translate_fieldname(self, fieldname):
+        """Not all fields that appear on Add User Form in Volto match in Plone."""
+        if fieldname == "roles":
+            msgid = _("Roles")
+        elif fieldname == "sendPasswordReset":
+            msgid = ICombinedRegisterSchema["mail_me"].title
+        else:
+            msgid = ICombinedRegisterSchema[fieldname].title
+        return self.translate(msgid)
 
     def validate_input_data(self, portal, original_data):
         """Returns a tuple of (required_fields, allowed_fields)"""
@@ -62,11 +81,23 @@ class UsersPost(Service):
         # check input data
         for fieldname in required:
             if not data.get(fieldname, None):
-                self.add_field_error(fieldname, f"Property '{fieldname}' is required.")
+                translated_fieldname = self.translate_fieldname(fieldname)
+                self.add_field_error(
+                    fieldname,
+                    _(
+                        "Property '${fieldname}' is required.",
+                        mapping={"fieldname": translated_fieldname},
+                    ),
+                )
         for fieldname in data:
             if fieldname not in allowed:
+                translated_fieldname = self.translate_fieldname(fieldname)
                 self.add_field_error(
-                    fieldname, f"Property '{fieldname}' is not allowed."
+                    fieldname,
+                    _(
+                        "Property '${fieldname}' is not allowed.",
+                        mapping={"fieldname": translated_fieldname},
+                    ),
                 )
 
         password = data.get("password")
@@ -75,16 +106,16 @@ class UsersPost(Service):
             if password is None and send_password_reset is None:
                 self.add_field_error(
                     "sendPasswordReset",
-                    "You have to either send a password or sendPasswordReset.",
+                    _("You have to either send a password or sendPasswordReset."),
                 )
             if password and send_password_reset:
                 self.add_field_error(
                     "sendPasswordReset",
-                    "You can't send both password and sendPasswordReset.",
+                    _("You can't send both password and sendPasswordReset."),
                 )
 
-    def add_field_error(self, field, message):
-        self.errors.append({"field": field, "message": message})
+    def add_field_error(self, field, msgid):
+        self.errors.append({"field": field, "message": self.translate(msgid)})
 
     def errors_to_string(self):
         return " ".join([error["message"] for error in self.errors])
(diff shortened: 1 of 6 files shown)


wesleybl:

erral:
I would leave the messages in the plone.restapi domain, and move them later to plone when we break plone 5 compatibility in plone.restapi.

ericof:
@erral Do we even have a `plone.restapi` domain? Is it worth creating one?

erral:
Yes we have it with some german translations, you can ask around :wink: 

We tried to move the existing translation to plone domain:
https://github.com/plone/plone.restapi/pull/1238

But we deferred it because we are keeping compatibility with Plone 5.

That's why I think it is better to keep these translations in plone.restapi domain and move them to plone when we break compatibility with Plone 5.

Otherwise it makes little sense adding some string in plone.restapi and others in plone domains.

erral:
Yes but I later remembered the PR to move plone.restapi translations into
plone domain and ended mixing things up.

If we add these translations to plone domain, we should move the existing 2
or 3 strings also to plone (see my other PR about this).

Then I can manage to update plone.app.locales for plone domain.

wesleybl:
> If we add these translations to plone domain, we should move the existing 2
or 3 strings also to plone (see my other PR about this).
>
> Then I can manage to update plone.app.locales for plone domain.

So should I update your PR? Or should this only be done when support for Plone 5 is removed?

erral:
opinions @tisto?

wesleybl:
@erral since there was already an attempt to change the domain, but it didn't go ahead, I changed it to the plone.restapi domain here, so we can proceed. Can you have a look please?

wesleybl:

erral (approved):
Translation-wise I see it OK.

wesleybl:
@erral some error messages, which come from Plone, have the `plone` domain. So I had to make a modification to consider both domains.

Another thing.. Field names are not enclosed in a `_( )`. 
For example: https://github.com/plone/plone.restapi/blob/8644250e6f1065727a1e8b90141235df37085598/src/plone/restapi/services/users/add.py#L86
Should I put them in Python code, so that i18n considers them? Or should I create a "manual" pot file?

erral:
Instead of using fieldnames can you use the field title? field titles should be msgid-s and thus the translate call would translate them. Something like this:

```python
from zope.i18n import translate

translate(field_title)
```

If those titles are not msgids, the translate call would do nothing, so it should be OK.

wesleybl:
> If those titles are not msgids, the translate call would do nothing, so it should be OK.

Field names are string only:

https://github.com/plone/plone.restapi/blob/8644250e6f1065727a1e8b90141235df37085598/src/plone/restapi/services/users/add.py#L55-L69

Field names are returned in the error message:

https://github.com/plone/plone.restapi/blob/8644250e6f1065727a1e8b90141235df37085598/src/plone/restapi/services/users/add.py#L77-L82

So they need to be translated.

erral:
If we just have field names and not titles, I would leave them as they are, or get their title from Plone and prepare a dict that does the translation from field name to field title, and then use that title there.

wesleybl:
Some of these fields aren't even in Plone. For example `sendPasswordReset` is in Volto:

https://github.com/plone/volto/blob/7f5bf0248ce60a3a23c57f179e0f1c8df1611b91/src/components/manage/Controlpanels/Users/UsersControlpanel.jsx#L430

And for others, the Plone translation doesn't match the Volto translation. So, I think it's better to create a translation unrelated to Plone, and more focused on Volto.

wesleybl:
> And for others, the Plone translation doesn't match the Volto translation.

By the way, shouldn't the translation of the add user form fields in Volto be the same as in Plone? Should Volto use the Plone terms? Or the other way around? In the case of pt_BR, the translations of the two do not match.

erral:
It *should* be, but perhaps we are hard-coding the user form directly in Volto instead of relying on the userschema. We are already using the userschema for the personal-information form (see https://github.com/plone/volto/pull/2917) so we should be using that here too.

wesleybl:
@erral while this is not done, I changed to using the title of the fields in translations, and treated the fields that do not exist in Plone:

https://github.com/plone/plone.restapi/blob/da4ebc95a5eb25c261a0bb7cbedca3c521527a5f/src/plone/restapi/services/users/add.py#L46-L55

I think it's ready now. Thank you for your help!

wesleybl:

erral:
> @erral while this is not done, I changed to using the title of the fields in translations, and treated the fields that do not exist in Plone:
> 
> https://github.com/plone/plone.restapi/blob/da4ebc95a5eb25c261a0bb7cbedca3c521527a5f/src/plone/restapi/services/users/add.py#L46-L55
> 
> I think it's ready now. Thank you for your help!

Thanks for working on this.

Can you open a new issue to use Plone user schema for this kind of this? This way we have it under the radar.

wesleybl:
@erral I create the issue:

https://github.com/plone/volto/issues/4018

wesleybl:
Can we merge this?

@erral @ericof @davisagli @jensens

src/plone/restapi/services/users/add.py
```diff
@@ -33,6 +36,24 @@ def publishTraverse(self, request, name):
...
+    def translate_fieldname(self, fieldname):
+        """Not all fields that appear on Add User Form in Volto match in Plone."""
+        msgid = ""
+        if fieldname == "roles":
+            msgid = "Roles"
+        elif fieldname == "sendPasswordReset":
+            msgid = ICombinedRegisterSchema["mail_me"].title
+        else:
+            msgid = ICombinedRegisterSchema[fieldname].title
+        return self.translate(msgid, "plone")
```
davisagli, line 55: @wesleybl I think it might be better to _not_ pass a domain when translating the field title. Zope message ids have an attached domain (inherited from the message factory used to construct it) and zope.i18n.translate will use it unless a different domain is passed in: https://github.com/zopefoundation/zope.i18n/blob/master/src/zope/i18n/__init__.py#L166

src/plone/restapi/services/users/add.py
```diff
@@ -62,11 +83,27 @@ def validate_input_data(self, portal, original_data):
...
+                translated_fieldname = self.translate_fieldname(fieldname)
                 self.add_field_error(
-                    fieldname, f"Property '{fieldname}' is not allowed."
+                    fieldname,
+                    self.translate(
+                        _(
+                            "Property '${fieldname}' is not allowed.",
+                            mapping={"fieldname": translated_fieldname},
+                        )
+                    ),
```
davisagli, line 106: Since the basic structure is repeated here twice, that makes me think that maybe the translation logic should go inside `add_field_error` to avoid repeating. This is just an idea and not critical, though.

davisagli (approved):
@wesleybl Thanks for working on this! I added a couple comments, but I don't think either of them are critical, and I would be okay with merging this as is.

src/plone/restapi/services/users/add.py
wesleybl: @davisagli the problem is the Roles field up there:

https://github.com/plone/plone.restapi/blob/da4ebc95a5eb25c261a0bb7cbedca3c521527a5f/src/plone/restapi/services/users/add.py#L49-L50
The `ICombinedRegisterSchema` interface does not have a Roles field. An interface that has this field is `IRoleCondition`:

https://github.com/plone/plone.app.contentrules/blob/4d3e8e28eb4c8a1d0cfed36d9f4543338c94f470/plone/app/contentrules/conditions/role.py#L23-L28

But I don't know if it makes sense to import it here.

Another option would be to create an entry in the `plone.restapi` domain:

```
_("Roles")
```
What would be the best option here?

wesleybl: Well, I thought it best to put the Roles in the plone.restapi domain.

Thus, I no longer pass the domain.

src/plone/restapi/services/users/add.py
wesleybl: Now I call the translation inside the _error and add_field_error methods.

wesleybl:

PR #1550, opened 2022-12-06
Fix timezone for Locking creation date

iFlameing:


src/plone/restapi/services/locking/__init__.py
```diff
@@ -16,7 +16,11 @@ def creator_url(username):
 
 
 def creation_date(timestamp):
-    return datetime.fromtimestamp(timestamp).isoformat()
+    # Note that isoformat does not add the Z at the end of the string, The lack of the
+    # timezone specifier causes Intl (and derivative libraries) in the browser not to
+    # use local time, which is considered a bug in applications. Therefore we add the Z
+    # to the end to make sure that the date will be interpreted properly by the client.
+    return datetime.fromtimestamp(timestamp).strftime("%Y-%m-%dT%H:%M:%SZ")
```
davisagli, line 23: Timestamp is a lock modification timestamp which was created in the local timezone here: https://github.com/zopefoundation/Zope/blob/master/src/OFS/LockItem.py#L95

To get it in UTC, we need to use utcfromtimestamp here:
```suggestion
    return datetime.utcfromtimestamp(timestamp).strftime("%Y-%m-%dT%H:%M:%SZ")
```

davisagli (changes requested):
There are also failing tests. The snapshot of the response for this service needs to be updated to include Z

jensens:

reebalazs:
@iFlameing tests fixed, please proceed as needed.

jensens (approved):

PR #1566, opened 2023-01-06
Improve blocks linkintegrity retrieval of internal links

### LINKED ISSUE #1565: Blocks linkintegrity does not find some internal links
The implementation of link integrity for blocks has a generic retriever to find links in any `url` or `href` field. But currently it only handles string URLs in those fields. The resolveuid transform handles lists of URLs and lists of objects that have an `@id` property, and so should the linkintegrity retriever.

davisagli:
Fixes #1565

The linksintegrity retriever should handle the same data formats that are handled by the resolveuid transform.

### DIFF
--- a/news/1565.bugfix
+++ b/news/1565.bugfix
@@ -1,0 +1,1 @@
+Fix blocks linkintegrity to find some links in `url` and `href` fields that were previously ignored. [davisagli]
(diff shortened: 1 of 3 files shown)


davisagli:

jensens (approved):

PR #1568, opened 2023-01-11
Testfix for changes in `zope.schema >= 7.0.0`

### LINKED ISSUE #1567: `zope.schema >= 7.0.0` changed type of `schema.Int`
Just figured out, that `zope.schema.Int` Field changed it's validation basis `_type` from `(int, )` to `int` since version `7.0.0` (they dropped the `zope.schema._compat.py` module). Some tests here check the `WrongType` exception on validation and have to be updated accordingly.

petschki:
fixes #1567

### DIFF
--- a/news/1567.bugfix
+++ b/news/1567.bugfix
@@ -1,0 +1,2 @@
+Fix tests for `zope.schema >= 7.0.0`
+[petschki]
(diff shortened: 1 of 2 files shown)


petschki:

petschki:
Hm ... I think this needs review of @plone/release-team ... is it planned to ship Plone 6.0.x with Zope 5.8 ? And does it make sense to test Plone 5.2 on master branch? `buildout.coredev#5.2` has `branch=7.x.x` in its sources.cfg for plone.restapi ...

mauritsvanrees:
Yes, Plone 6.0.1 will ship with Zope 5.8. There is no 5.7.x maintenance branch, so there really is no other option than following along.
As a temporary workaround we could pin an older zope.schema in coredev 6.0.

And yes, plone.restapi is one of a few packages that we (so far) are trying to keep running on both Plone 5.2 and 6.0, although only on Python 3. On 5.2 it is an optional upgrade for those that want it.

petschki:
OK, I think I can change this testfix to work for both zope.schema versions in that case.

petschki:

mauritsvanrees (approved):
Thanks!

PR #1570, opened 2023-01-19
Use incoming request to produce location for @tus-upload

### LINKED ISSUE #1569: tus endpoint via ++api++ doesn't give api url in location
The @tus-upload endpoint creates a temporary upload url which is returned in the header. A follow up PATCH request is made to that url.

If the original @tus-upload POST is made via the ++api++ url, the resulting location in the response header does not include  ++api++ in the path and so the follow up post must be made directly to the plone instances.

This breaks for situations where it is being used in volto.

The code in question is https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/services/content/tus.py#L122

instification:
Fixes #1569


instification:
This change is breaking tests related to tus so I will resolve those and create some specific tests for using ++api++

instification:

src/plone/restapi/services/content/tus.py
```diff
@@ -117,9 +117,12 @@ def reply(self):
         tus_upload = TUSUpload(uuid4().hex, metadata=metadata)
 
         self.request.response.setStatus(201)
+        url = self.request.getURL()
+        if url.endswith("@tus-replace"):
+            url = url[:-7] + "upload"
```
tisto, line 122: @instification I am wondering if we shouldn't be more explicit here to make the code easier to read.

src/plone/restapi/tests/test_tus.py
```diff
@@ -541,6 +541,35 @@ def record_event(event):
         sm.unregisterHandler(record_event, (IObjectAddedEvent,))
         sm.unregisterHandler(record_event, (IObjectModifiedEvent,))
 
+    def test_tus_with_api(self):
```
tisto, line 544: @instification I had to consult the test code to understand what this PR does.

instification:

src/plone/restapi/services/content/tus.py
instification: @tisto I hav updated the code, hopefully it makes it more readable now?

JeffersonBledsoe:
Hey @tisto, any chance of looking at this again/ is it likely to be merged?

JeffersonBledsoe:

tisto:
@JeffersonBledsoe will add it to my todo list for the weekend...

JeffersonBledsoe:
Thanks! 🙂

JeffersonBledsoe:

JeffersonBledsoe:
@tisto Sorry to nudge you again on this one. Any chance you'd be able to take a look over this after the sprint?

djay:
@jensens can we get this merged?

JeffersonBledsoe:

tisto:
@sneridagh @davisagli I am not too familiar with the semantics of the tus upload. Is that something you can review?

sneridagh (approved):
LGTM, not sure if it's a breaking change or not...

@instification @JeffersonBledsoe any insight?

djay:
@sneridagh it's basically broken right now. Is it a breaking change to fix something that doesn't work?

sneridagh:
@djay but could be that someone else is using it without the `++api++`, right? or I misunderstood?

djay:
@sneridagh are you saying that it's valid to call restapi endpoints without ++api++?

djay:
@sneridagh the code doesn't force ++API++. It just ensures that the first part of the Tus exchange returns urls of the same form as the first one rather the current behaviour of ripping out the ++API++ if it's there. Broken.

tisto:
@djay ++api++ is one way (the recommended way) of hooking into the REST API, we never deprecated the other ways.

davisagli (approved):
@tisto @sneridagh This looks fine to me, and fine to merge for 8.x as a bugfix.

Before the fix, if the client accesses the endpoint without `++api++` then it produces the correct URL. But if the client accesses the endoint with `++api++` then it produces a URL that is missing `++api++`, and won't work if `++api++` is how the frontend webserver knows to route traffic to the backend.

With the fix, if the client accessed the endpoint without `++api++` then it gets a redirect without `++api++`, and if it accessed the endpoint with `++api++` then it gets a redirect that includes it. Sounds good to me.

We might need a more general fix to include `++api++` in other places, but I don't see harm in merging this focused fix in the meantime.

tisto (approved):

tisto:

tisto:

PR #1573, opened 2023-01-25
Fix auth cookie for Zope users in login endpoint

### LINKED ISSUE #1572: Login endpoint sets bad cookie for Zope-level user
Steps to reproduce:
1. Log in to volto as a Zope-level admin user
2. Create a new page titled Test and leave it in the private workflow state
3. Go to /++api++/test

Expected: the JSON response for the content endpoint is displayed
Actual: Unauthorized

The problem seems to be that for Zope-level users, the `__ac` cookie is set by the credentials_cookie_auth plugin in the root-level PAS, instead of the session plugin in the Plone-level PAS. But credentials_cookie_auth needs access to the user's password to set the cookie correctly, and it is getting lost in the chain from the login endpoint -> PlonePAS -> PAS updateCredentials.

Bugfix incoming.

davisagli:
Fixes #1572

### DIFF
--- a/news/1572.bugfix
+++ b/news/1572.bugfix
@@ -1,0 +1,1 @@
+Fix bug where the `@login` endpoint did not set the correct `__ac` cookie for Zope users. [davisagli]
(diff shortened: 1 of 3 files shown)


davisagli:

sneridagh (approved):

ericof (approved):

PR #1582, opened 2023-02-13
Provide own groups if requesting own user data

ksuess:


ksuess:

src/plone/restapi/serializer/user.py
```diff
@@ -68,18 +68,19 @@ class SerializeUserToJson(BaseSerializer):
...
-        batch = HypermediaBatch(self.request, groups)
-        groups_data = {
-            "@id": batch.canonical_url,
-            "items_total": batch.items_total,
-            "items": sorted(batch, key=lambda x: x["title"]),
-        }
-        if batch.links:
-            groups_data["batching"] = batch.links
+        gtool = getToolByName(self.context, "portal_groups", None)
+        if gtool:
```
nileshgulia1, line 72: @ksuess Is this a bugfix or a feature? because as far as I can see we have own groups data in users.

ksuess: It's a feature. The change from `ISerializeToJsonSummary`to `ISerializeToJson`adds goups to the response.

The gtool stuff is for users not registered in Plone instance but above. They do not have a `portal_groups`. So the change in `src/plone/restapi/serializer/user.py` catches the error.

nileshgulia1 (approved):

PR #1584, opened 2023-02-13
Provide a JsonCompatible adapter for RichTextValue objects

### LINKED ISSUE #1583: Provide IJsonCompatible adapter for RichTextValue objects
When defining a default value in a RichText field, this value needs to be passed through a `IJsonCompatible` adapter, but we are missing the converter for `RichTextValue` objects, and thus the edit form rendering breaks in Volto.

Dexterity object definition:
```python
    download_full_dataset_text = RichText(
        title=_(
            'Text of the "Download full dataset entry',
        ),
        description=_(
            "Enter the text that will be shown in the Download tab",
        ),
        default=(
            "<p>If you want to download the full dataset, click <a"
            ' href="/en/how-to-guides/how-to-download-spatial-data/how-to-download-m2m">here</a>'
            " to learn more.</p>"
        ),
        required=False,
        readonly=False,
    )
```

plone.restapi fail when rendering the Volto edit form:
```python
2023-02-13 17:05:44,536 ERROR   [Zope.SiteErrorLog:22][waitress-1] TypeError: http://localhost:3000/en/@types/DataSet
Traceback (innermost last):
  Module ZPublisher.WSGIPublisher, line 162, in transaction_pubevents
  Module ZPublisher.WSGIPublisher, line 371, in publish_module
  Module ZPublisher.WSGIPublisher, line 266, in publish
  Module ZPublisher.mapply, line 85, in mapply
  Module ZPublisher.WSGIPublisher, line 63, in call_object
  Module plone.rest.service, line 22, in __call__
  Module plone.restapi.services.types.get, line 114, in reply
  Module plone.restapi.services.types.get, line 136, in reply_for_type
  Module plone.restapi.types.utils, line 329, in get_info_for_type
  Module plone.restapi.types.utils, line 272, in get_jsonschema_for_portal_type
  Module plone.restapi.types.utils, line 261, in get_jsonschema_for_fti
  Module zope.component.hooks, line 135, in adapter_hook
  Module plone.restapi.serializer.converters, line 136, in dict_converter
  Module plone.restapi.serializer.converters, line 68, in json_compatible
  Module zope.component.hooks, line 135, in adapter_hook
  Module plone.restapi.serializer.converters, line 136, in dict_converter
  Module plone.restapi.serializer.converters, line 68, in json_compatible
  Module zope.component.hooks, line 135, in adapter_hook
  Module plone.restapi.serializer.converters, line 80, in default_converter
TypeError: No converter for making RichTextValue object. (Did you mean <attribute>.raw or <attribute>.output?) (<class 'plone.app.textfield.value.RichTextValue'>) JSON compatible.

```

erral:
Fixes #1583

### DIFF
--- a/news/1583.bugfix
+++ b/news/1583.bugfix
@@ -1,0 +1,2 @@
+Fix serialization of RichTextValue objects when an explicit context is not known.
+[erral]
--- a/src/plone/restapi/serializer/converters.py
+++ b/src/plone/restapi/serializer/converters.py
@@ -11,6 +11,7 @@ from plone.restapi.bbb import safe_text
 from plone.restapi.interfaces import IContextawareJsonCompatible
 from plone.restapi.interfaces import IJsonCompatible
 from zope.component import adapter
+from zope.component import getMultiAdapter
 from zope.component import queryMultiAdapter
 from zope.component.hooks import getSite
 from zope.globalrequest import getRequest
@@ -177,16 +178,10 @@ def timedelta_converter(value):
 @adapter(IRichTextValue)
 @implementer(IJsonCompatible)
 def richtextvalue_converter(value):
+    # Use the context-aware converter with the site as the context
     context = getSite()
-    if context is not None:
-        adapter = queryMultiAdapter((value, context), IContextawareJsonCompatible)
-        if adapter:
-            return adapter()
-    return {
-        "data": json_compatible(value.output),
-        "content-type": json_compatible(value.mimeType),
-        "encoding": json_compatible(value.encoding),
-    }
+    adapter = getMultiAdapter((value, context), IContextawareJsonCompatible)
+    return adapter()
 
 
 @adapter(IRichTextValue, Interface)
--- a/src/plone/restapi/serializer/converters.py
+++ b/src/plone/restapi/serializer/converters.py
@@ -7,12 +7,12 @@ from decimal import Decimal
 from persistent.list import PersistentList
 from persistent.mapping import PersistentMapping
 from plone.app.textfield.interfaces import IRichTextValue
-from plone.dexterity.interfaces import IDexterityContent
 from plone.restapi.bbb import safe_text
 from plone.restapi.interfaces import IContextawareJsonCompatible
 from plone.restapi.interfaces import IJsonCompatible
 from zope.component import adapter
 from zope.component import queryMultiAdapter
+from zope.component.hooks import getSite
 from zope.globalrequest import getRequest
 from zope.i18n import translate
 from zope.i18nmessageid.message import Message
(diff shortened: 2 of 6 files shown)


jensens:
Your solution provides only the rendered safe-HTML. But what if we want to get the raw data from the field, for an editor or alike? This could be any content type as raw, i.e. markdown.  I do not think this implementation is enough. I would expect something like a dict with output, raw and its configured types, i.e something like:
```python
{
    "output": {
        "data": "<p>foo<p>",
        "type": "text/safe-html",
    "raw": {
        "data": "foo",
        "data": "text/plain",
}
```

erral:
This is the initial draft.

I first checked what happens when you add a simple HTML as a default value of the RichText field. I will test further with some other values.

jensens:
OK! Since this is not a draft PR nor labels were applied, my understanding was like it is ready for review.

erral:
I have updated the PR to implement the more generic fallback mechanism discussed in issue #1583.

Instead of adding a new global adapter, I improved the `json_compatible` helper to fall back to the Plone site (`getSite()`) when no context is provided. This allows the existing context-aware converter for `RichTextValue` to work in all scenarios (nested, schema defaults, etc.) while keeping the output format consistent with standard fields.

news/1583.bugfix
```diff
@@ -0,0 +1,2 @@
+Provide a JsonCompatible adapter for RichTextValue objects to fix crashes during nested serialization and schema default values.
+[erral]
```
davisagli:
```suggestion
Fix serialization of RichTextValue objects when an explicit context is not known. @erral
```

src/plone/restapi/serializer/converters.py
```diff
@@ -62,12 +62,15 @@ def json_compatible(value, context=None):
...
+        context = getSite()
+
     if context is not None:
         adapter = queryMultiAdapter((value, context), IContextawareJsonCompatible)
         if adapter:
             return adapter()
-    else:
-        return IJsonCompatible(value, None)
+
+    return IJsonCompatible(value, None)
```
davisagli, line 71: @erral It looks like this adds some overhead every time `json_compatible` is called without a context, because now it has to call getSite and queryMultiAdapter every time.

Maybe it does not have enough impact to matter; I have not measured.

But I think there is an easy way to implement this that avoids that: Add an IJsonCompatible adapter for RichTextValue which calls the existing RichtextDXContextConverter using getSite() as the context. Then we are not adding overhead when we convert other types.

davisagli (changes requested):

erral:

davisagli:

davisagli (approved):

PR #1588, opened 2023-02-16
Implement IPurgePaths for RestAPI traversal

### LINKED ISSUE #1587: Implement IPurgePaths for ++api++ urls
When `plone.app.caching` is installed and setup, purging the cache should also purge `plone.restapi` endpoints that use seamless mode (++api++ traversal)

ericof:
Fix #1587

### DIFF
--- a/news/1587.feature
+++ b/news/1587.feature
@@ -1,0 +1,1 @@
+Implement IPurgePaths for RestAPI traversal (++api++) @ericof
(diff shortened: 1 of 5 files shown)


jensens (approved):
You may want to add z3c.caching to the test dependencies. OTOH in our integrated test setup it is not a problem.

mauritsvanrees:
It may make sense to only do this when plone.restapi is activated. Something like:

```
from Products.CMFPlone.utils import get_installer

if not get_installer(portal, context.REQUEST).is_product_installed("plone.restapi"):
    return []
```

Otherwise this needlessly adds purge paths for sites that do not actually use restapi.

I guess executing the current code does not take long, I am not so worried about that. But I wonder if this then means an extra five or so purge requests to Varnish, which may be more noticeable, or may end up overflowing the Varnish purge queue, if that still exists in recent versions of Varnish.

PR #1592, opened 2023-03-06
Add `show_revert` property to @history endpoint response

sneridagh:
We need this for the frontend because we can't rely in "being the last element" for not showing the "Revert to this version" button, since the history elements can be of different nature.

The original template relies in `diff_current_url` property for it, I'm renaming it to a more sensible one, also making it boolean.

Related Volto PR: https://github.com/plone/volto/pull/4471

### DIFF
--- a/news/1593.feature
+++ b/news/1593.feature
@@ -1,0 +1,1 @@
+Add show_revert property to @history endpoint response @sneridagh


sneridagh:

ericof (approved):
LGTM

davisagli (changes requested):
I discussed with @sneridagh; I think we can determine whether to show the link on the frontend only without adding to the API.

sneridagh:
Reworked it in: https://github.com/plone/volto/pull/4471 so we don't have to rely on a backend setting.

@ericof @davisagli let me know, then we can close this one.

PR #1595, opened 2023-03-06
Added support for nested schemas with resolveuid block deserializer

### LINKED ISSUE #1594: resolveuid deserializer doesn't handle lists or nested dicts
The resloveuid deserializer currently makes the assumption that the `url` or `href` field is in the root of the block data, or it has been created by a volto_object_browser that is set to `single` mode. This means that making use of nested or lists of block data or lists (such as with the `object_list` widget in Volto) doesn't cause the field to be converted to use `resolveuid`.

For example, the following block data:

```json
{
  "123": {
    "@type": "foo",
    "my_values": [
      {
        "url": "/news"
      }
    ]
  }
}
```

Should end up being stored as this:

```json
{
  "123": {
    "@type": "foo",
    "my_values": [
      {
        "url": "../resolveuid/{doc_uid}"
      }
    ]
  }
}
```

But instead is saved as in the original

JeffersonBledsoe:
As the initial implementation only iterated over the `smart` field, there may be a performance impact with this PR with very large blocks as there is no way of knowing if a nested value will have a smart field without iterating through them all. Any comments on this/ input on if this could be improved are welcomed!

Closes #1594


JeffersonBledsoe:

src/plone/restapi/deserializer/blocks.py
```diff
@@ -105,29 +105,33 @@ def __init__(self, context, request):
         self.context = context
         self.request = request
 
+    def _process_data(self, data, field=None):
```
razvanMiu, line 108:
```
def _process_data(self, data, field=None):
    if isinstance(data, str) and field in self.fields:
        return path2uid(context=self.context, link=data)
    if isinstance(data, list):
        return [self._process_data(data=value, field=field) for value in data]
    if isinstance(data, dict):
        if data.get('@type', None) is 'URL' and data.get('value', None):
            data['value'] = path2uid(context=self.context, link=data["value"])
        elif data.get('@id', None):
            item_clone = deepcopy(data)
            item_clone["@id"] = path2uid(
                context=self.context, link=item_clone["@id"]
            )
            return {
                field: self._process_data(data=value, field=field)
                for field, value in item_clone.items()
            }
        return {
            field: self._process_data(data=value, field=field)
            for field, value in data.items()
        }
    return data
    
I have a proposal to also check for a field type like I did above. Based on https://github.com/plone/volto/issues/2252

JeffersonBledsoe:
@tisto Thoughts on this? David expressed some concerns in #1594

JeffersonBledsoe:

davisagli (approved):

PR #1599, opened 2023-03-18
Allow passing additional parameters to the @users endpoint when deleting a user

### LINKED ISSUE #1598: Add parameters to users DELETE endpoint
We have been hit by this issue that was raised [in the Community Forum some time ago](https://community.plone.org/t/alternate-way-do-delete-users/3009). We can't delete users in our Volto site because it has a lot of users and content, and the deleting goes through all objects walking up them all.

The solution is to pass `delete_memberareas` and `delete_localroles` as `False` to the `deleteMembers` method of `portal_membership` (knowing, of course, that local roles will not be removed).

So, how can we expose this in the `@users` endpoint's `DELETE` method?

It looks like `DELETE` method calls should not have any request body, so the only alternative is to pass those options as parameters in the url, say

```http
DELETE /@users/erral?delete_localroles=0&delete_memberareas=0
```

Is this acceptable or do you see any other way to handle this?

erral:
Fixes #1598 

This way we can signal not to delete local roles or memberarea folders, which is useful in large sites that do not rely on local roles or member areas.

That's why I added a warning in the documentation signaling that this may be potentially a security issue


erral:

erral:

erral:
Mmm how is that the exact same tests here in GitHub Action pass and in Jenkins fail? 🤔

davisagli:
@erral Hmm, good question. I looked a bit and didn't find the explanation. I can reproduce the failing tests if I run the plone.restapi tests from the Plone coredev buildout.

erral:

erral:
I think I found it, I had some unneeded transaction.commits() which, when run through buildout.coredev made the tests fail.

erral:

erral:

tisto:

erral:

erral:

erral:

tisto:

src/plone/restapi/services/users/delete.py
```diff
@@ -5,6 +5,9 @@
 from zope.publisher.interfaces import IPublishTraverse
 
 
+FALSE_VALUES = (0, "0", False, "false", "no")
```
tisto, line 8: @erral I am wondering if we should define this higher up to make sure it is used everywhere. This is not a blocker for a merge though.

tisto (approved):

tisto:
@erral PR merged. As said, I owe you a few beers when we meet in Eibar! :)

src/plone/restapi/services/users/delete.py
erral: Yes, it could be useful indeed.

PR #1610, opened 2023-03-30
CI: run the tests on Ubuntu 20.04, instead of latest. [7.x.x]

mauritsvanrees:
Python 2.7 and 3.6 are not available on latest.
See broken runs on this PR job:
https://github.com/plone/plone.restapi/actions/runs/4565635527?pr=1608

BTW, for me it would be fine to stop testing with Python 3.6: it is no longer supported by Plone 5.2. But if this PR gets it going for a few more months, then all is well.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,7 @@ name: Plone RESTAPI CI
 on: [push]
 jobs:
   build:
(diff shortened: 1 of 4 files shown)


mauritsvanrees:
Green on gh-actions, and no relevant changes for Jenkins.

PR #1612, opened 2023-03-31
Test on Plone 6.0.3

ale-rt:


erral:
@ale-rt test fail because a previously untranslated string was translated into Spanish for 6.0.3.

The translation is correct, so if you fix the.documentation file, tests will pass and we can merge

erral (changes requested):
Change please the documentation file that creates the test fail, and we can merge

davisagli:
@ale-rt @erral The version update and test fix were done separately, so I'll close this.

PR #1615, opened 2023-04-02
Update declared Python support to match what we are testing

davisagli:


jensens (approved):

setup.py
```diff
@@ -10,7 +10,7 @@
     3,
     6,
```
mauritsvanrees, line 11: This 6 should become a 7, otherwise the assertion does not fail on 3.6.

I doubt anyone still sees this message, since we already use `python_requires` and pip won't even find a package if it does not match.  But with older pip/buildout/setuptools, I guess you can still see it.

tisto:
@davisagli @mauritsvanrees @jensens this is technically a breaking change. I have to check if we declared this in an earlier release. Please do not make a release.

davisagli:
@tisto We stopped testing Python 3.6 in October (https://github.com/plone/plone.restapi/pull/1504), so I would say this is just fixing the package metadata to match what is already the case. You asked about it there and @sneridagh pointed out that Python 3.6 reached EOL in December 2021. We could revert the change to python_requires though, in case someone was relying on using it even though it's not tested or supported.

tisto:
@davisagli you are right. This might have been a mistake and would have required a major version bump. The compromise for me back then was to stop testing it but not break existing installations, by amending the check or removing the support in setup.py. In any case, this should have been more explicit, at least in the release notes (https://github.com/plone/plone.restapi/pull/1504/files#diff-49ccbfd65a22c789ebae6b7a7884e526be5af092af5cfa2ab0547be4952f7aa3R1).

We were in a hurry and I didn't have much time to review and tackle this properly back then. :(

I'd like us to get together at Beethoven Sprint to discuss how we want to handle release management for plone.restapi and Volto in the future (and possibly Plone). I would like us to do a better job by properly defining our "API" that developers can rely on to not break within minor versions. If we do not properly define this and then stick to it, semantic versioning has no meaning and we can as well just stop putting effort into it IMHO.

mauritsvanrees:
Let me compare this with core Plone 5.2.  Here we no longer support 3.6, and will stop supporting 3.7 in summer.  But I try not to introduce changes that break on 3.6 or 3.7.  So currently the classifiers in CMFPlone are this:

```
        "Programming Language :: Python :: 2.7",
        "Programming Language :: Python :: 3.7",
        "Programming Language :: Python :: 3.8",
```

But the `python_requires` still allow 3.6:

```
    python_requires='>=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*,!=3.5.*',
```

In summer I will drop the 3.7 classifier, but keep `python_requires` the same.
If some code no longer works on 3.6 or 3.7, PRs are welcome.  But breakage is not so likely, because we still support 2.7.  The biggest likely problems are that you may need to change some version pins, which we are definitely not doing in coredev: we no longer have a `[versions:python36]` section.  That took too much time to maintain.

Translated to `plone.restapi` master I would say:

1. Change the assertion to only check for Python 2 (and we probably should not use an assertion):

```
if sys.version_info.major == 2:
    raise ValueError(
        "plone.restapi 8 requires Python 3. "
        "Please downgrade to plone.restapi 7 for Python 2 and Plone 4.3/5.1."
    )
```

2. Keep the classifiers as they are since this PR.

3. Reset the `python_requires` to 3.6 (and the trailing zero is not needed):

```
python_requires=">=3.6",
```

4. Clarify the situation in README/CHANGES.rst.

Does that sound good?

As for bumping major versions:

- In the Plone release as a whole, so practically `Products.CMFPlone` and `Plone`, we _can_ and _have_ dropped official support for Python versions in bugfix releases, with the caveat mentioned above: we try not to break it, but can give no guarantees.  Otherwise 4.3.17 (or somewhere around that time) should have been called 5.0 because it dropped 2.6 support.  Of course 5.0 already existed, so we could have called it 4.4.  Further, 5.2.10 dropped 3.6 so should have been 5.3, 5.2.13 in summer drops 3.7 so should be 5.4.  It gets too crazy.
- According to [semver.org](https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api) a dependency update without changing the public API would be compatible, so a bugfix release is fine.
- For individual packages like `plone.restapi` I *would* say: make a major release when you drop a Python version.
- If a package already has a release that is Plone 6 only, then dropping Python 2 support is a non-event: just a bit of cleanup.  In that case we can choose to do this in a bugfix or minor release of the package.

tisto:
@mauritsvanrees thanks for the detailed write-up Maurits! That all makes sense to me. I created a new PR taking your recommendations into account:

https://github.com/plone/plone.restapi/pull/1619

It makes sense that we aggressively drop Python versions that reached the EOL. We just need to officially declare this and make sure people are aware.

One of my main goals for the upcoming Beethoven Sprint is to discuss the release policy of plone.rest(api) and Volto and properly document this. I would love to have you around for your insights and a broader discussion that could include Plone (Core/Classic) as well.

Do we have a place where we document what you just wrote in your comment? That might be a good starting point for our documentation efforts.

mauritsvanrees:
https://plone.org/download/release-schedule covers most of it.

stevepiercy:
For specific things for the project, I would recommend `/RELEASING.md`. @polyester and I were just discussing this today for Training and Documentation. For example, what are the steps needed to deploy `2023.training.plone.org` in 2024? It can cover branch policy, as in [Volto](https://6.docs.plone.org/volto/developer-guidelines/contributing.html#branch-policy).

PR #1622, opened 2023-04-17
Fix param handling in querystringsearch GET

### LINKED ISSUE #1621: Regressions in `@querystringsearch` GET endpoint param handling
The GET version of the `@querystringsearch` endpoint added in #1616 has some regressions in handling query parameters:
- `b_start` and `b_size` are not in the request for ZCatalog to use
- The fields used by the summary serializer are not where it is looking

davisagli:
Copy the params from the querystring into the request body, because there are too many bits of code that expect to find things there.

Adjusted the test to confirm pagination works.

Fixes #1621

### DIFF
--- a/src/plone/restapi/services/querystringsearch/get.py
+++ b/src/plone/restapi/services/querystringsearch/get.py
@@ -7,8 +7,6 @@ from plone.restapi.services import Service
 from urllib import parse
 from zope.component import getMultiAdapter
 
-import json
-
 
 zcatalog_version = get_distribution("Products.ZCatalog").version
 if parse_version(zcatalog_version) >= parse_version("5.1"):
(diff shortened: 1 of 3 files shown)


davisagli:

tisto (approved):

src/plone/restapi/services/querystringsearch/get.py
```diff
@@ -73,14 +72,16 @@ class QuerystringSearchPost(Service):
...
+        return querystring_search()
 
 
 class QuerystringSearchGet(Service):
     """Returns the querystring search results given a p.a.querystring data."""
 
     def reply(self):
+        # We need to copy the JSON query parameters from the querystring
+        # into the request body, because that's where other code expects to find them
+        self.request["BODY"] = parse.unquote(self.request.form.get("query", "{}"))
```
jensens, line 84: This is so ugly. I understand, fast results and so, but this has to be fixed in a sane way at some point.

PR #1630, opened 2023-04-26
update_password should check password validity

tschorr:
`p.restapi` is using `PasswordResetTool.resetPassword` for updating passwords, however this method doesn't check for password validity (https://github.com/plone/Products.CMFPlone/blame/master/Products/CMFPlone/PasswordResetTool.py#L103).
This results in Volto bypassing the default password policy. Although the minimum password length has been updated to 8 characters (https://github.com/plone/Products.PlonePAS/pull/69), Volto still mentions a minimum of 5 characters (https://github.com/plone/volto/blob/master/src/components/theme/PasswordReset/PasswordReset.jsx#L54) and Plone lets you get away with this. I actually could use a 3 letter password.

### DIFF
--- a/src/plone/restapi/services/users/add.py
+++ b/src/plone/restapi/services/users/add.py
@@ -266,6 +266,7 @@ class UsersPost(Service):
         pas = getToolByName(self.context, "acl_users")
         mt = getToolByName(self.context, "portal_membership")
         pwt = getToolByName(self.context, "portal_password_reset")
+        registration_tool = getToolByName(self.context, "portal_registration")
 
         if target_user is None:
             self.request.response.setStatus(404)
@@ -273,7 +274,6 @@ class UsersPost(Service):
 
         # Send password reset mail
         if list(data) == []:
-            registration_tool = getToolByName(self.context, "portal_registration")
             registration_tool.mailPassword(username, self.request)
             return
 
@@ -299,6 +299,13 @@ class UsersPost(Service):
         # Reset the password with a reset token
         if reset_token:
             try:
+                err = registration_tool.testPasswordValidity(new_password)
+                if err is not None:
+                    return self._error(
+                        403,
+                        "Invalid password",
+                        _(err),
+                    )
                 pwt.resetPassword(username, reset_token, new_password)
             except InvalidRequestError:
                 return self._error(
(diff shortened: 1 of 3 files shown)


tschorr:
@pbauer this PR will allow to use `P.PasswordStrength` with Volto.

tschorr:

tschorr:
Jenkins test failures look unrelated, but I might be missing something.

src/plone/restapi/services/users/add.py
```diff
@@ -299,6 +299,13 @@ def update_password(self, data):
         # Reset the password with a reset token
         if reset_token:
             try:
+                err = registration_tool.testPasswordValidity(new_password)
+                if err is not None:
+                    return self._error(
+                        400,
+                        "Invalid password",
+                        _(err),
```
wesleybl, line 307: It seems that sometimes `testPasswordValidity` returns an [already translated string](https://github.com/plone/Products.CMFPlone/blob/8d33738b058b2e9d8d73566d4a161a3379b79c81/Products/CMFPlone/RegistrationTool.py#L225), sometimes it returns a [msgid](https://github.com/plone/Products.CMFPlone/blob/8d33738b058b2e9d8d73566d4a161a3379b79c81/Products/CMFPlone/RegistrationTool.py#L198). Then you need to do a test to see what is returned. This should be fixed in Plone.

wesleybl: uhnn but you are not passing the `confirm` parameter. Shouldn't you pass it here?

tschorr: > It seems that sometimes `testPasswordValidity` returns an [already translated string](https://github.com/plone/Products.CMFPlone/blob/8d33738b058b2e9d8d73566d4a161a3379b79c81/Products/CMFPlone/RegistrationTool.py#L225), sometimes it returns a [msgid](https://github.com/plone/Products.CMFPlone/blob/8d33738b058b2e9d8d73566d4a161a3379b79c81/Products/CMFPlone/RegistrationTool.py#L198). Then you need to do a test to see what is returned. This should be fixed in Plone.

I would expect `_(...)` to do the right thing in any case? +1 for fixing this in the `RegistrationTool`.

> uhnn but you are not passing the `confirm` parameter. Shouldn't you pass it here?

There's a sensible default and we do not have a password confirmation in this case.

> Another weird thing is we have an update_password in add.py. Shouldn't this be on update.py?

Maybe, but imho that's for another ticket.

tschorr:

jensens (approved):

src/plone/restapi/services/users/add.py
tschorr: @wesleybl I suggest you open the tickets yourself.

PR #1635, opened 2023-05-01
Do not expose private metadata via relationfield serializer.

### LINKED ISSUE #1634: Do not expose information via relationfield serializer
**Description**
The relationfield serializer should respect the `View` permission. 

**Problem**
Some metadata like, Title and Description can be exposed thru the relationsfield serializer

**Reproduce**
1. Create 3 pages
2. Refer to page 2 and 3 from page 1 via relatedItems
3. Remove View permission for page 2, or page 3 (You can simply publish page 1 and for example page 2 and keep page 3 private)

Result: bot page 2 and page 3 are serialized in relatedItems

```
...
    "relatedItems": [
        {
            "@id": "http://localhost:8080/Plone/page-2",
            "@type": "Document",
            "UID": "4f0b980e8a144a55b9244b8b942e93d2",
            "description": "",
            "image_field": null,
            "image_scales": null,
            "review_state": "published",
            "title": "Page 2"
        },
        {
            "@id": "http://localhost:8080/Plone/page-3",
            "@type": "Document",
            "UID": "bec10b954049433088df076e6eb0208d",
            "description": "",
            "image_field": null,
            "image_scales": null,
            "review_state": "private",
            "title": "Page 3"
        }
    ],
...
```

**Expected result**
Only list accessible relatedItems and do not expose any metadata information

``` 
...
    "relatedItems": [
        {
            "@id": "http://localhost:8080/Plone/page-2",
            "@type": "Document",
            "UID": "4f0b980e8a144a55b9244b8b942e93d2",
            "description": "",
            "image_field": null,
            "image_scales": null,
            "review_state": "published",
            "title": "Page 2"
        }
    ],
...
```

maethu:
Fixes #1634 

This PR implements two changes:
1. Only return summary if the user has `View` permission on target object.
2. Filter `None` values in RelationListFieldSerializer, which result from not having a summary for inaccessible content.


maethu:

src/plone/restapi/serializer/relationfield.py
```diff
@@ -33,4 +35,9 @@ class RelationChoiceFieldSerializer(DefaultFieldSerializer):
 @adapter(IRelationList, IDexterityContent, Interface)
 @implementer(IFieldSerializer)
 class RelationListFieldSerializer(DefaultFieldSerializer):
-    pass
+    def __call__(self):
+        value = self.get_value()
+        if value:
+            return [item for item in json_compatible(value) if item]
+        else:
+            return super().__call__()
```
jensens, line 43:
```suggestion
        if value:
            return [item for item in json_compatible(value) if item]
        return super().__call__()
```

erral:
If the user hasn't the `View` permission on the target object, should it be returned as a related object **at all**?

src/plone/restapi/serializer/relationfield.py
```diff
@@ -17,7 +18,8 @@
 @adapter(IRelationValue)
 @implementer(IJsonCompatible)
 def relationvalue_converter(value):
-    if value.to_object:
+    mtool = api.portal.get_tool("portal_membership")
+    if value.to_object and mtool.checkPermission("View", value.to_object):
```
jensens, line 22: this is too much overhead and probably slow. Use the Zope basics, those are perfectly fine here:
```
from AccessControl.SecurityManagement import getSecurityManager
getSecurityManager().checkPermission(permission, obj)
```

jensens: btw, if you anyway use `plone.api`, there is `api.user.check_permission` too as an alternative.

erral: If I am not wrong, plone.api may only be used in plone.restapi tests right? @plone/restapi-team

jensens: I am not sure what the rules are, but if there are those rules, they must be documented somewhere.

jensens:
> If the user hasn't the `View` permission on the target object, should it be returned as a related object **at all**?

In Plone we have these two permissions for a reason. You might want or not want to show an object in listing. You might want to tell, look there is a relation to X even if you're not allowed to view X details. Using these two permissions this is possible. Usually both are mapped the same in workflow permission mappings. But, depends on the customization, not always.

davisagli:
Also consider: The serialization is also used when loading the item for editing. What should happen if a user has permission to edit an item that has references to items that they do not have permission to view? I am not sure what the right answer is, but I am sure that it is not silently removing the reference.

maethu:
Yes my bad... I should know this 🙈 The issue is in my workflow definition. 

But to clarify:
In the context if a relation, or multiple relations, the parent of a relation is the current object. 

So if doc1 has doc2 as related item. And the user does not have "Access Contents Information" on doc1.
Does this mean relatedItems should be empty?

I totally doe see the issue regarding: `Also consider: The serialization is also used when loading the item for editing. What should happen if a user has permission to edit an item that has references to items that they do not have permission to view?` 

This is no longer related to my issue, but I'm a bit confused what make sense and what not. 

I adapted the code to check for "Access content information" on the __parent__ of the relation. 

But I'm not sure if the helps a lot... or if it makes things more confusing. 
Feel free to close this PR, it's not really a bug in the end, more a integration issue depending on your application. 

I implemented some custom relation serializer on my end, which filters relations for anonymous users depending on the `View` permission. Because for my use case this makes the most sense.

src/plone/restapi/serializer/relationfield.py
jensens: For the sake of readability: No `else` needed when the `if` returns.

src/plone/restapi/serializer/relationfield.py
```diff
@@ -17,7 +18,12 @@
 @adapter(IRelationValue)
 @implementer(IJsonCompatible)
 def relationvalue_converter(value):
-    if value.to_object:
+    has_permission = api.user.has_permission(
```
jensens, line 21: @plone/restapi-team In a comment above @erral wrote `plone.api` usage is allowed in this package in tests only and asked if this is right. May one you confirm this?

src/plone/restapi/serializer/relationfield.py
```diff
@@ -17,7 +18,12 @@
...
+        "Access Contents Information",
+        obj=value.__parent__
```
jensens, line 23: I do not get why we now check on parent? This is is probably a misunderstanding and wrong.

jensens (changes requested):

src/plone/restapi/serializer/relationfield.py
maethu: @jensens 

Probably another miss-understanding on my end. 

I used the parent, since @davisagli wrote in the issue #1634 
`@maethu In general, permission to view an object's metadata in listings is controlled by the "Access contents information" permission on its container, not the View permission.`

I'm happy to change this!

src/plone/restapi/serializer/relationfield.py
tisto: We agreed that we can use plone.api in tests. Though, if I am not mistaken we also decided lately that plone.restapi can use plone.api everywhere, correct? Sorry for the late reply here. Just want to get this straight.

tisto:
@maethu @davisagli @jensens what's the status of this PR?

davisagli:
@tisto From my point of view, I am waiting for a proposal for how to resolve the problem I mentioned in https://github.com/plone/plone.restapi/pull/1635#issuecomment-1531626897

I think we need something like a way to include UIDs for items that the current user does not have permission to view. Probably this should be a querystring option. Then the edit view can ask for them and make sure they are not removed, but other use cases can leave them out.

Issue #1639, opened 2023-05-18
The plone.app.multilingual dependency is not declared


folix-01:
The plone.app.multilingual package is used by plone.restapi but we have no dependency for it

folix-01:
The other approach is to use the plone.app.multilingual conditionally

PR #1640, opened 2023-05-18
Add the plone.app.multilingual dependency to the setup.py

folix-01:
related to https://github.com/plone/plone.restapi/issues/1639


folix-01:
Related to https://github.com/plone/Products.CMFPlone/issues/3779

PR #1642, opened 2023-05-18
Update image scales in blocks data when serializing internal links

davisagli:
This adds `image_scales` while serializing blocks (in the same serialization transform that handles resolveuid URLs), in order to support rendering images with a `srcset` (https://github.com/plone/volto/pull/3337)

`image_scales` is added at the same level as the `url` or `@id` key. That way the only change to the existing data structure is adding a new key, which should be backwards compatible in most cases.


davisagli:
@pnicolli My worry about backwards compatibility is this: The resolveuid serializer will process any field with the name `url`, even in custom blocks. So we can break them if those blocks suddenly start getting objects instead of strings there, and aren't expecting it.

Options for avoiding that:
1. Put the image_scales beside the url instead of inside it:
```
{
    "url": "/path/to/image",
    "image_scales": { ... }
}
```
2. Put the image_scales inside a url object, but only for the image block (still not 100% backwards compatible, but at least the change is more constrained in scope)

davisagli:
@pnicolli I updated this to keep the `url` as a string and put `image_scales` at the same level -- but I haven't had time to try it with volto yet.

pnicolli:
@davisagli thanks! I will give it a try today.

pnicolli:
I tried pulling this branch and running my frontend branch against this, created an image block, saved it and image data is sent correctly by the rest api, so everything looks good!

pnicolli:
Also tried replacing the image in the Image object and I am getting the correct new scales 👍

davisagli:

davisagli:

pnicolli:
Tested it with https://github.com/plone/volto/pull/3337 and it works fine for me.

davisagli:

davisagli:
@tisto Okay with you if I go ahead and merge this? I tested it in the context of our big project and didn't see any regressions.

sneridagh (approved):

ericof (approved):

PR #1643, opened 2023-05-19
Use plone.app.multilingual conditionally

folix-01:


src/plone/restapi/services/configure.zcml
```diff
@@ -46,10 +46,12 @@
...
       />
-  <include
-      package=".multilingual"
-      zcml:condition="have plone-5"
-      />
+  <configure zcml:condition="installed plone.app.multilingual">
+    <include
+        package=".multilingual"
+        zcml:condition="have plone-5"
+        />
```
jensens, line 53: While at it, we can drop the outdated condition too.
```suggestion
    <include
        package=".multilingual"
        />
```

jensens (changes requested):

jensens:
Why does it touch many unrelated files?

erral:
Why is this needed?

plone.restapi master is already Plone 5.x only which already includes p.a.multilingual as a dependency.

folix-01:
> Why does it touch many unrelated files?

Because of black I think

folix-01:
> Why is this needed?
> 
> plone.restapi master is already Plone 5.x only which already includes p.a.multilingual as a dependency.

Because we need to remove il from the Products.CMFPlone

jensens:
We remove it as a dep of CMFPlone and add it as a dep on Plone. Thus it becomes a core-addon. Architecture wise this is favorable. At the end, for the user, nothing changed.

jensens (approved):

jensens:

Issue #1644, opened 2023-05-27
Startup error: plone.app.contentrules.ManageContentRules permission not found


mauritsvanrees:
See report by @1letter on the [community forum](https://community.plone.org/t/plone-6-0-5-soft-released/17490/3?u=mauritsvanrees).
`plone.restapi` uses a permission from `plone.app.contentrules`, but does not load its zcml.  Depending on the order in which the various packages get loaded, startup may fail.

Also, contentrules might become a "core add-on", and become a dependency of `Plone` instead of `Products.CFMPlone`.   While we are at it, we could load the code conditionally.

PR #1645, opened 2023-05-27
Fix possible startup error by loading plone.app.contentrules zcml

mauritsvanrees:
Also: only load code related to contentrules when this package is available.
Fixes https://github.com/plone/plone.restapi/issues/1644

I have a small doubt where some of these conditions should be on `plone.contentrules` instead of `plone.app.contentrules`.  But I guess this is one of those split packages where you either have none or both.  I doubt that anyone has written an own implementation on top of `plone.contentrules` that actually works with the current restapi code around it.  I think we do not need to worry.


mauritsvanrees:

davisagli (approved):
@mauritsvanrees Perfect, thanks!

PR #1647, opened 2023-05-30
Fix timezone handling in history endpoint and tests

### LINKED ISSUE #1617: Bug Report: Unexpected "Z" appended to API output
I updated `plone.restapi` version, I found the but for timezone on `@history` view. My env is JST(UTC+9). I look at this view as 9 hours later.

### Description: 
Currently, the plone.restapi package outputs datetime values with an appended "Z" at the end, indicating that the datetime is in UTC timezone. However, I think the original data is saved in local timezone as a Unix timestamp in milliseconds.

This behavior can be confusing for users because it is not clear whether the timestamp represents the local time or UTC. Therefore, it may be better to output the datetime value in a format that clearly indicates whether it is in local or UTC timezone.

### Steps to Reproduce: 
Make a request to the API that returns a datetime value.
Inspect the response body.
Expected behavior
The datetime value should be in a format that clearly indicates whether it is in local or UTC timezone.

### Actual behavior:
The datetime value has a "Z" appended at the end, which indicates that it is in UTC timezone. However, the original data is saved in local timezone as a Unix timestamp in milliseconds.

### Related PR:
https://github.com/plone/plone.restapi/pull/1531/files

### Environment
plone.restapi version: 8.35.2

davisagli:
Fixes #1617

Also fixes the StaticTime helper used in tests to specify that the fake times used in tests are UTC, which should help with the problem where the times produced in tests are different depending on the local timezone of the machine where they run.

(This causes many example responses to be updated to add `+00:00` to the serialized datetimes. This better matches the actual serialized datetimes that are produced when not running tests.)


davisagli:

jensens:

davisagli:

PR #1649, opened 2023-05-31
handle suffix with not traversable objects

cekk:
The use-case is a link that ends with "/@@download/file" for example (http://my-site.com/foo/@@download/file).

Previous implementation can't get the object and returned not modified url, because /file is not traversable with the object traverser, but only with the browser traverser.

Current implementation return "resolveuid/123456/@@download/file"


davisagli (approved):

davisagli:

jensens (approved):

davisagli:

PR #1654, opened 2023-06-06
Navigation remote url

iFlameing:


src/plone/restapi/services/navigation/get.py
```diff
@@ -146,6 +146,8 @@ def navtree(self):
                 "description": safe_text(brain.Description),
                 "review_state": json_compatible(brain.review_state),
                 "use_view_action_in_listings": brain.portal_type in types_using_view,
+                "remoteUrl": json_compatible(brain.getRemoteUrl),
```
tisto, line 149: @iFlameing can you elaborate on your use case? Why do we use "remoteURL" and not "remote_url"?

PR #1680, opened 2023-08-18
unregister handler even if test fails

davisagli:


davisagli:

PR #1690, opened 2023-08-24
WIP: TUS temporary file optimization

JeffersonBledsoe:


davisagli:
@JeffersonBledsoe NamedBlobFile uses IStorage adapters to process the value that is passed in (https://github.com/plone/plone.namedfile/blob/master/plone/namedfile/storages.py). It might make sense to implement this as an IStorage adapter for TUSUpload.

> it also bypasses DX validation of the blob as that seems to read this into memory for some reason. instead we should work out how to still validate but not read the data unless we really have to

Yeah, I don't think we can bypass validation entirely. If the entire file is read during validation that's surprising, have you identified where that is happening?

djay:
@davisagli I've put in the storage adapter and that works. I've also put in a test to show it works.

But the validation does actually read the whole file into memory (also shown in the test).

- MinMaxLen is being used with 0 as min by default and that is applied to the contents of the blob, not the NamedBlobFile.
- The validation being applied is against [INamedBlobFile](https://github.com/plone/plone.namedfile/blob/master/plone/namedfile/interfaces.py#L19) and the attribute '''data''' and getting this reads the whole file.
- but it actually doesn't matter. it will read the data in regardless - [get_schema_validation_errors](https://github.com/zopefoundation/zope.schema/blob/master/src/zope/schema/_bootstrapfields.py#L1043) 
- either we somehow make data be lazy loaded?
- OR we get rid of this call? https://github.com/plone/plone.namedfile/blob/master/plone/namedfile/field.py#L109
- OR maybe use something other than a bytes field in the INamedBlobFile schema taht doesn't do validation against file contents itself?

Not really the sure the right way to solve this yet. ideas?

djay:
I also suspect that some of the other storages like FileUploadStorage might also be be already buffered into a local file at least some of the time.  In that case, the file can be moved rather than read again improving performance. I haven't looked into this yet however.

djay:
@davisagli I had a go at changing the validation to avoid reading in the whole file in https://github.com/plone/plone.namedfile/pull/155

src/plone/restapi/services/content/tus.py
```diff
@@ -394,3 +394,14 @@ def expires(self):
...
         return formatdate(expiration, False, True)
+
+
+@implementer(IStorage)
+class TUSUploadStorable:
+    """Special storage to skip reading and writing of the upload since it's already on disk"""
+
+    def store(self, data, blob):
+        if not isinstance(data, TUSUpload):
+            raise NotStorable('Could not store data (not of "FileUpload").')
```
davisagli, line 405:
```suggestion
            raise NotStorable('Could not store data (not of type "TUSUpload").')
```

davisagli: @djay Now that I'm looking at things more carefully, I'm wondering if this special storage adapter is really needed.

Previously, an open file was passed into the NamedBlobFile constructor. This should end up using the [FileDescriptorStorable](https://github.com/plone/plone.namedfile/blob/master/plone/namedfile/storages.py#L62) which uses [blob.consumeFile](https://github.com/zopefoundation/ZODB/blob/891a8a6e056e58eedb010f8971a492a4b083f2e4/src/ZODB/blob.py#L227), which is already using rename_or_copy_blob.

src/plone/restapi/tests/test_tus.py
```diff
@@ -395,6 +394,60 @@ def test_tus_can_upload_text_file(self):
...
+
+        self.assertEqual(
+            blob_write,
+            1,
+            "Slow write to blob instead of os rename. Should be only 1 on init",
+        )
+        self.assertEqual(blob_read, 0, "Validation is reading the whole blob in memory")
+        # TODO: would be better test to patch file read instead and ensure its not called?
+
+        ZODB.blob.Blob.open = old_open
```
davisagli, line 449: This should go in a finally block in case there's an error during the test.

davisagli (changes requested):

PR #1701, opened 2023-08-28
Remove deprecated token parameter from vocabularies endpoint

tisto:


tisto:
@davisagli @sneridagh I removed the deprecated "token" parameter altogether in the PR now. I was wondering if we should return a 422 unprocessable entity (or something else) in case someone passes the "token" paramenter instead of "tokens". The idea would be to fail hard, so people notice it instead of just ignoring it (which might be hard to track).

The downside of this approach is that we have code that we might want to remove in plone.restapi 10 to clean up stuff. Opinions?

davisagli:
-1 from me, this breaks volto as it currently uses this endpoint

PR #1710, opened 2023-09-22
Drop Python 3.7 support

tisto:


README.rst
```diff
@@ -74,12 +74,10 @@ Install ``plone.restapi`` by adding it to your buildout.
 Python / Plone Compatibility
 ============================
 
-plone.restapi 8 requires Python 3 and works with Plone 5.2 and Plone 6.x.
-
-plone.restapi 8 does not officially support Python 3.6.
+plone.restapi 8 and 9 require Python 3 and work with Plone 5.2 and Plone 6.x.
```
tisto, line 77: @mauritsvanrees @davisagli FYI: I updated the README to reflect our "more strict" policy regarding old Python versions.

tisto:

README.rst
```diff
@@ -74,12 +74,15 @@ Install ``plone.restapi`` by adding it to your buildout.
 Python / Plone Compatibility
 ============================
 
-plone.restapi 8 requires Python 3 and works with Plone 5.2 and Plone 6.x.
+plone.restapi 9 requires Python 3 and works with Plone 5.2 and Plone 6.x.
 
-plone.restapi 8 does not officially support Python 3.6.
+plone.restapi 8 entered "maintenance" mode with the release of plone.restapi 9 (September 2023).
+It is not planned to backport any features to this version and we highly recommend to upgrade to plone.restapi 9.
+plone.restapi 8 will receive security releases (if they are any) for the next two years (until September 2025).
```
tisto, line 81: @mauritsvanrees are you ok with those lines? I had the feeling that we have to give people time to upgrade. I don't expect you or the security team to provide those fixes, I can backport stuff if necessary.

tisto:
I triggered a Jenkins job manually: https://jenkins.plone.org/job/pull-request-6.0-3.11/851/

news/1570.bugfix
```diff
@@ -1,2 +1 @@
-Use incoming request to produce location for @tus-upload
-[instification]
+Use incoming request to produce location for @tus-upload @instification
```
stevepiercy, line 1: To make it clear that the former is not a GitHub username, but an endpoint.
```suggestion
Use incoming request to produce location for `@tus-upload`. @instification
```

stevepiercy (changes requested):
Docs look good. Minor MyST syntax clarification. Thank you!

README.rst
mauritsvanrees: @tisto This triggers many thoughts.

This is longer than what we promise until now in the [release schedule](https://plone.org/download/release-schedule), but shorter than what we said last Friday, when talking about security supporting 6.0 for five years.

Because what we try to do until now, and what I managed yesterday, is to provide security fixes for **all minor versions of a package**.  You can see this best in the [`plone.namedfile` advisory](https://github.com/plone/plone.namedfile/security/advisories/GHSA-jj7c-jrv4-c65x), where I made fixes in version 5.6.1, 6.0.3, 6.1.3, 6.2.1.  Those last three are used in various Plone 6.0.x releases.  It is more work, but then: 5.6.1 and 6.2.1 were really needed, and then fixing the versions in between is not so hard.

Following this train of thought, for `plone.restapi` theoretically I should have made releases for 8.0.x, 8.1.x, 8.2.x, etc until 8.43.x.  Or at least from 8.32 onwards, as that is the release that was in Plone 6.0.0.  Of course I didn't do that.  Also, if we would have made a traditional hotfix package, there is no way I could have really tested compatibility with all these feature releases, except when the underlying code for the vulnerable part hasn't changed.

But I think we are moving towards saying that you should be on the last bugfix release of Plone 6.0.x to have security support.  Or at least that we will only fix one branch of a package per Plone minor version, and that we expect that you can use this also when you are still stuck on Plone 6.0.0.

With the inclusion of plone.restapi 9 in next Plone 6.0.8, we are saying you can use that version in 6.0.0 as well.  If we can't say this, then we should not include it.

End conclusion:
Since it *is* a major release of a package in a Plone bugfix release, it is good to promise some extra support for 8.x.x.
So I am fine with these lines.

mauritsvanrees (approved):

stevepiercy (approved):

PR #1714, opened 2023-09-27
Expose the sources + their target instead of just some sources in "breaches" attribute in @linkintegrity endpoint response

jaroel:
Needs https://github.com/plone/plone.app.linkintegrity/pull/95

Breaking: this changes the shape of an endpoint.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,6 +19,7 @@ Breaking changes:
 - Change the @linkintegrity endpoint to add `items_total`, the number of contained items which would be deleted. @davisagli, @danalvrz, @pgrunewald (#1636)
 - The default branch was renamed from `master` to `main`. @tisto, @davisagli (#1695)
 - Drop support for Python 3.7. Set python_requires to >= 3.8 @tisto (#1709)
+- Expose the sources + their target instead of just some sources in "breaches" attribute in @linkintegrity endpoint response. [jaroel]
 
 
 New features:
@@ -461,7 +462,7 @@ Bug fixes:
 
 
 - Added url field to Actions (#817)
-- Update statictime tests following changes to p.a.disucssion (see 
+- Update statictime tests following changes to p.a.disucssion (see
   https://github.com/plone/plone.app.discussion/pull/204) - [instification] (#1520)
 - Update @portrait endpoint to use sanitized user id [instification] (#1524)
 
--- a/src/plone/restapi/services/linkintegrity/get.py
+++ b/src/plone/restapi/services/linkintegrity/get.py
@@ -34,16 +34,21 @@ class LinkIntegrityGet(Service):
             item = uuidToObject(uid)
             item_path = "/".join(item.getPhysicalPath())
             links_info = item.restrictedTraverse("@@delete_confirmation_info")
-            breaches = links_info.get_breaches()
             data = getMultiAdapter((item, self.request), ISerializeToJsonSummary)()
-            data["breaches"] = []
-            for breach in breaches:
-                for source in breach.get("sources", []):
-                    # remove unwanted data
-                    source["@id"] = source["url"]
-                    del source["url"]
-                    del source["accessible"]
-                    data["breaches"].append(source)
+            data["breaches"] = [
+                {
+                    "target": result["target"],
+                    "sources": [
+                        {
+                            "@id": source["url"],
+                            "uid": source["uid"],
+                            "title": source["title"],
+                        }
+                        for source in result["sources"]
+                    ],
+                }
+                for result in links_info.get_breaches()
+            ]
             # subtract one because we don't want to count item_path itself
             data["items_total"] = len(catalog(path=item_path)) - 1
             result.append(data)
--- a/src/plone/restapi/tests/test_services_linkintegrity.py
+++ b/src/plone/restapi/tests/test_services_linkintegrity.py
@@ -20,7 +20,6 @@ import unittest
 
 
 class TestLinkIntegrity(unittest.TestCase):
-
     layer = PLONE_RESTAPI_BLOCKS_FUNCTIONAL_TESTING
 
     def setUp(self):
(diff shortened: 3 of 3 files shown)


sneridagh:
LGTM - Except for the breaking, I know this sucks, but I'd rather try not to trigger a breaking change. @tisto your call.

tisto:
@sneridagh if we have a breaking change, we have a breaking change. No harm in that. We will just do a plone.restapi 10 release...

If you are using semver you shouldn't be afraid to bump a major version :)

tisto:
@jaroel I did not really look into the code. Question up front: can we write the plone.restapi code in a way that it will still work if plone.app.querystring is not the latest version? I am ok with a breaking change release of plone.restapi. Though, I wouldn't want to introduce a hard dependency on a plone.app.querystring version if possible.

jaroel:
@tisto the shape of plone.app.linkintegrity's data has not changed, not a problem. You'll see some duplicate entries in the modal, but that's what the new stuff is for.

tisto:

src/plone/restapi/tests/http-examples/linkintegrity_get.resp
```diff
@@ -7,9 +7,18 @@ Content-Type: application/json
...
+                        "@id": "http://localhost:55001/plone/doc-1",
+                        "title": "First document",
+                        "uid": "SomeUUID000000000000000000000001"
+                    }
+                ],
+                "target": {
+                    "@id": "http://localhost:55001/plone/doc-2",
+                    "title": "Second document",
+                    "uid": "SomeUUID000000000000000000000002"
+                }
```
davisagli, line 21: @jaroel @tisto @sneridagh  I don't understand why it's necessary to change the shape of the response like this. The new `target` inside `breaches` is just duplicating information from the top level list (lines 6-7), isn't it? (Of course, it's simple to add the title and uid at the top level if they are needed.)

Maybe I am missing something?

jaroel: @davisagli The target is different for (also deleted) subitems.

https://github.com/plone/plone.restapi/blob/7baab0eb68207dd9abc3816e333e04eb9cdcd000/src/plone/restapi/tests/test_services_linkintegrity.py#L349-L352 gives an idea of that. Does this help?

davisagli: @jaroel Okay, that helps. But do the breaches for a subitem need to be listed under their parent? If not, we could keep the same shape (with all the targets at the top level for both items the user selected and subitems, and each one contains a list of its breaches/sources).

Just trying to avoid breaking compatibility if there isn't an important reason.

jaroel: @davisagli yep; I need to push it one level up and it should work as-is. 🥇

jaroel: @davisagli could have another look?

src/plone/restapi/services/linkintegrity/get.py
```diff
@@ -38,6 +38,9 @@ def reply(self):
             data = getMultiAdapter((item, self.request), ISerializeToJsonSummary)()
             data["breaches"] = []
             for breach in breaches:
+                if breach["target"]["uid"] != uid:
+                    uids.append(breach["target"]["uid"])
```
davisagli, line 42: Should only append it if it's not already there, right?

davisagli (commented):
@jaroel One question, and I think you need to update the sample response that goes in the docs, but yeah, that looks more promising.

jaroel:
> @jaroel One question, and I think you need to update the sample response that goes in the docs, but yeah, that looks more promising.

Afaict the request and response haven't changed?

davisagli:
> Afaict the request and response haven't changed?

@jaroel Uh right, I was thinking wrong.

Looks good to me now.

tisto:

tisto:
@jaroel p.a.linkintegrity has been release with a fix. Do you want to amend this PR before merging and releasing?

jaroel:
> @jaroel p.a.linkintegrity has been release with a fix. Do you want to amend this PR before merging and releasing?

@tisto Could you release plone.app.linkintegrity 3.x too (with the backport)? Then I can remove the skipUnless. We can also land this work first and then change after. Dealer's choice.

tisto:
@jaroel here you go:

https://pypi.org/project/plone.app.linkintegrity/3.6.2/

jaroel:

tisto (approved):

CHANGES.rst
```diff
@@ -461,7 +461,7 @@ Bug fixes:
 
 
 - Added url field to Actions (#817)
-- Update statictime tests following changes to p.a.disucssion (see 
+- Update statictime tests following changes to p.a.disucssion (see
```
stevepiercy, line 464:
```suggestion
- Updated ``statictime`` tests following changes to ``p.a.discussion`` (see
```

CHANGES.rst
```diff
@@ -461,7 +461,7 @@ Bug fixes:
...
   https://github.com/plone/plone.app.discussion/pull/204) - [instification] (#1520)
```
stevepiercy, line 465: Use GitHub `@` syntax.

```suggestion
  https://github.com/plone/plone.app.discussion/pull/204) - @instification (#1520)
```

CHANGES.rst
```diff
@@ -461,7 +461,7 @@ Bug fixes:
...
 - Update @portrait endpoint to use sanitized user id [instification] (#1524)
```
stevepiercy, line 466:
```suggestion
- Updated ``@portrait`` endpoint to use sanitized user id. @instification (#1524)
```

news/1714.bugfix
```diff
@@ -0,0 +1 @@
+Treat sub-items like items in @linkintegrity endpoint. [jaroel]
```
stevepiercy, line 1:
```suggestion
Treat sub-items like items in ``@linkintegrity`` endpoint. @jaroel
```

stevepiercy (changes requested):
Clean up change notes for reStructuredText syntax and rendering of GitHub username links.

CHANGES.rst
```diff
@@ -461,9 +461,9 @@ Bug fixes:
 
 
 - Added url field to Actions (#817)
-- Update statictime tests following changes to p.a.disucssion (see 
-  https://github.com/plone/plone.app.discussion/pull/204) - [instification] (#1520)
-- Update @portrait endpoint to use sanitized user id [instification] (#1524)
+- Updated ``statictime`` tests following changes to ``p.a.discussion`` (see
+  https://github.com/plone/plone.app.discussion/pull/204) - @instification (#1520)
```
stevepiercy, line 465: One more and it's good!

```suggestion
  https://github.com/plone/plone.app.discussion/pull/204). @instification (#1520)
```

stevepiercy (changes requested):

sneridagh (approved):

CHANGES.rst
tisto: @stevepiercy this is something I am always wondering about. In German, you do not use punctuation when using bullet points. Is the way to go here to always use "." at the end of each changelog entry? I am ok with it either way, I just want to have one recommended way. :)

stevepiercy (approved):
Thank you!

tisto:

CHANGES.rst
stevepiercy: Either is acceptable in English, depending on the context. Usually the change notes are in fact complete sentences, which should end in a period, regardless of whether it is in a list. When the bulleted list is an actual list of items and not a sentence, such as:

- Item1
- Item2
- Item3

...then you would not use a period at the end of the item.

See the Microsoft Style Guide for lists: https://learn.microsoft.com/en-us/style-guide/scannable-content/lists

tisto: @stevepiercy thanks for the clarification Steve! I will use a period from now on in changelog entries.

PR #1720, opened 2023-10-07
Cleanup robot tests

gforcada:
It was not running for quite a long time ago already.


news/1.bugfix
davisagli, line 1: This should be named 1720.bugfix, not 1.bugfix. With the latter, towncrier will put a link to #1 in the generated changelog.

davisagli (changes requested):

news/1.bugfix
davisagli: Actually I would call it 1720.internal

gforcada: Thanks, done :)

davisagli:
@gforcada Tests are failing because there are still imports from plone.app.robotframework

gforcada:
Actually not in `plone.restapi` but on `plone.app.event` and `plone.app.contenttypes` 😅

As this repository is still not `plone/meta` enabled, and it might take some time for it, at least parts of it, specially the testing part as this repo still supports Plone 5.2, it's a bit hard to fix.

We can keep the dependency here, to avoid the breakage, it does not feel clean, but at least works, hopefully 🍀

davisagli:
@gforcada I'm trying it with plone.restapi's `test` extra depending on `plone.app.contenttypes[test]`

tisto (approved):
Wow. I had no clue we still had this dependency in there. Thank you for cleaning up @gforcada!

davisagli:

davisagli (approved):

tisto:
@gforcada @davisagli did you folks run into this while trying Python 3.12 or rather the robotframework cleanup?

I just started a branch here:

https://github.com/plone/plone.restapi/pull/1722

and I was about to remove p.a.robotframework and friends because they cause trouble. I just want to avoid that we work on this at the same time...If that's the case I am happy to stop my efforts...

davisagli:
@tisto I will merge this PR as soon as the jenkins builds finish.

gforcada:
@davisagli thanks! 😄

PR #1723, opened 2023-10-15
fix: use pam layer in pam services

mamico:
This fix limits the use of multilingual services only if multilingual is actually installed, not only if it is present tar the 'eggs' (because in the new distribution of Plone, `plone.app.multilingual` is always present).


davisagli:

davisagli (approved):
LGTM, thanks

PR #1725, opened 2023-10-30
Remove regex from sphinx-copybutton config, now that linenos are excl…

stevepiercy:
…uded by default

See https://sphinx-copybutton.readthedocs.io/en/latest/use.html#automatic-exclusion-of-prompts-from-the-copies

See also https://github.com/plone/documentation/pull/1563


stevepiercy:

PR #1726, opened 2023-10-31
extractCredentials: do not fail when the request is too large to read.

mauritsvanrees:
This is a problem since at least Zope 5.8.4, introduced in Plone 6.0.7.
See https://github.com/plone/Products.CMFPlone/issues/3848 and https://github.com/zopefoundation/Zope/pull/1180.

This PR makes sure that a too big request body does not throw an error when extracting credentials.  This makes large image upload work again in ClassicUI with `plone.restapi` installed.

Uploading a file or image larger than 1MB using `plone.restapi` still likely fails, unless you have increased the `form-memory-limit` in `zope.conf`.  Fixing this may need a change [along these lines](https://github.com/zopefoundation/Zope/issues/1177#issuecomment-1784889996), but that would be a more far reaching change that needs careful consideration.

### DIFF
--- a/news/3848.bugfix
+++ b/news/3848.bugfix
@@ -1,0 +1,2 @@
+``extractCredentials``: do not fail when the request is too large to read.
+@maurits
--- a/src/plone/restapi/pas/plugin.py
+++ b/src/plone/restapi/pas/plugin.py
@@ -14,6 +14,7 @@ from Products.PluggableAuthService.interfaces.plugins import IAuthenticationPlug
 from Products.PluggableAuthService.interfaces.plugins import IChallengePlugin
 from Products.PluggableAuthService.interfaces.plugins import IExtractionPlugin
 from Products.PluggableAuthService.plugins.BasePlugin import BasePlugin
+from zExceptions import BadRequest
 from zope.component import getUtility
 from zope.interface import implementer
 
@@ -98,6 +99,8 @@ class JWTAuthenticationPlugin(BasePlugin):
         # is intended to change or update the logged in user.
         try:
             creds = deserializer.json_body(request)
+        except BadRequest:
+            pass
         except exceptions.DeserializationError:
             pass
         else:


mauritsvanrees:

mauritsvanrees:
As alternative, or as extra, we could do this before calling `creds = deserializer.json_body(request)`:

```
if request.get_header("Content-Type") != "application/json":
    return
```

This was suggested by @davisagli in https://github.com/plone/Products.CMFPlone/issues/3848#issuecomment-1769272646

In my case in ClassicUI during an image upload, the content type is something like `multipart/form-data; boundary=...`.  When just browsing the site, the content type is None.  But then, when I visit http://localhost:8080/Plone/++api++/ in the browser (so not called by javascript), the Content-Type request header is also None.  So that check might not be enough.

davisagli:
@mauritsvanrees In both of those cases, the request does not have a JSON body, and we don't want to waste effort in trying to parse it as JSON. The only problem would be if there is a client which is sending a JSON body but not sending the `Content-Type: application/json` header.

It should be a conditional around the code that tries to get login & password from the JSON body though, not a return out of the function. We still need to run the code after that looks for the token in the Authorization header, regardless of request content type.

davisagli:
Hmm, the whole chunk of code looking for login and password in the request body seems misplaced. Shouldn't that be done only in the login service?

mauritsvanrees:
> Hmm, the whole chunk of code looking for login and password in the request body seems misplaced. Shouldn't that be done only in the login service?

I think you are right.  Theoretically this chunk gets a dict with `login` and `password`, but would then be passed on to `authenticateCredentials` a few lines down, and this only [checks for a `token`](https://github.com/plone/plone.restapi/pull/1726/files#diff-41816888fc078e400f02c4aafbd2183a261e9cfcdc7f08720de50510843fb98aR126-R128).  So extracting login and password indeed seems useless.

mauritsvanrees:
Closed in favour of PR #1728.

PR #1727, opened 2023-10-31
extractCredentials: do not read json from the request.

mauritsvanrees:
The result was never used, and it may fail when the request is too large to read. This is a problem since at least Zope 5.8.4, introduced in Plone 6.0.7. See https://github.com/plone/Products.CMFPlone/issues/3848 and https://github.com/zopefoundation/Zope/pull/1180.

This PR is an alternative to https://github.com/plone/plone.restapi/pull/1726.  See discussion there.

### DIFF
--- a/news/3848.bugfix
+++ b/news/3848.bugfix
@@ -1,0 +1,3 @@
+``extractCredentials``: do not read json from the request.
+The result was never used, and it may fail when the request is too large to read.
+@maurits
--- a/src/plone/restapi/pas/plugin.py
+++ b/src/plone/restapi/pas/plugin.py
@@ -6,8 +6,6 @@ from datetime import datetime
 from datetime import timedelta
 from plone.keyring.interfaces import IKeyManager
 from plone.keyring.keyring import GenerateSecret
-from plone.restapi import deserializer
-from plone.restapi import exceptions
 from Products.CMFCore.permissions import ManagePortal
 from Products.PageTemplates.PageTemplateFile import PageTemplateFile
 from Products.PluggableAuthService.interfaces.plugins import IAuthenticationPlugin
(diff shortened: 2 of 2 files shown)


mauritsvanrees:
This [fails](https://github.com/plone/plone.restapi/actions/runs/6709273431/job/18231949574?pr=1727#step:10:232) because [in the `test_api_login_grants_zmi` test](https://github.com/plone/plone.restapi/blob/c43fd4fd16bc13c9e5cc919fe0f7d954eece9ee9/src/plone/restapi/tests/test_functional_auth.py#L100-L107) we get an Unauthorised.

What this test tries to do, and what no longer works with this PR:

* Do a POST to the `@login` endpoint with login and password.
* Request `/manage_workspace` on the Zope root, with the `__ac` cookie that is set.

When I instead request `/manage_workspace` on the *Plone* root, it *does* work.  The initial response of the `@login` POST *does* have an `__ac` cookie, and its path is `/`, so that should not be the problem.  But *something* more happened before, that no longer happens if we remove these lines.

It has to do with `ZODBUserManager.authenticateCredentials` which does not mind that it receives credentials from a different plugin, like ours.  It happily uses them, and the Zope root `acl_users` would authenticate us, as it finds the admin user.  But now that our `extractCredentials` no longer passes on these credentials, no user is found.

The [`@login` endpoint](https://github.com/plone/plone.restapi/blob/c43fd4fd16bc13c9e5cc919fe0f7d954eece9ee9/src/plone/restapi/services/auth/login.py) still does something that makes the user be authenticated in Plone, but it no longer works on the Zope root level.

So I guess this approach is not good enough.
What do you think @davisagli ?

davisagli:
@mauritsvanrees Okay, makes sense: we need to extract the credentials in a PAS extract credentials plugin so that they are available for use by other authenticateCredentials plugins.

I still kind of feel like this would belong more cleanly in a separate plugin, since it is not really directly related to authenticating a JWT token. But let's leave it for now.

PR #1728, opened 2023-10-31
jwt_auth plugin extractCredentials: check request content-type

davisagli:
Alternative to #1726 

We should only try to parse the request body as JSON if there's a request header saying it's JSON.

### DIFF
--- a/news/1728.bugfix
+++ b/news/1728.bugfix
@@ -1,0 +1,1 @@
+Fix jwt_auth extractCredentials plugin to only try to read credentials from the request body if there is a `Content-Type: application/json` header. @davisagli
--- a/src/plone/restapi/pas/plugin.py
+++ b/src/plone/restapi/pas/plugin.py
@@ -96,13 +96,14 @@ class JWTAuthenticationPlugin(BasePlugin):
         # Prefer any credentials in a JSON POST request under the assumption that any
         # such requested sent when a JWT token is already in the `Authorization` header
         # is intended to change or update the logged in user.
-        try:
-            creds = deserializer.json_body(request)
-        except exceptions.DeserializationError:
-            pass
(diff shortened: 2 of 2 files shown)


davisagli:

mauritsvanrees (approved):
LGTM.  This fixes it for the ClassicUI case: uploading large files works again.

I still would suggest to combine this with my PR #1726, which catches `BadRequest`.  This probably will not fix large uploads in Volto, but at least it would shift the error to the service that handles the uploads, instead of here in the PAS plugin.

But I did not try this in combination with Volto, so maybe there is no problem anymore.  It would be good to get that confirmed though.

mauritsvanrees:
For the record, I confirm that with this PR a POST to the `@login` endpoint still works, regardless of whether you set a Content-Type header or not, as long as you either set an Accept header or use the `++api++` namespace.

davisagli:
@mauritsvanrees The Accept header doesn't sound relevant to me. That tells the server what content-type the client hopes to receive in the response, not what content-type it is sending in the request.

I confirmed that uploading large files in Volto still triggers the error, but that's expected. I guess we can catch BadRequest here, but I think that just hides the problem.

To really fix it for Volto, we need to:
- make plone.restapi support multipart/form-data, like Dieter suggested
- OR improve plone.restapi's TUS upload endpoint to work without sharing a filesystem between instances
- make volto send uploads in the new format

I think this is a good path forward but it's not going to happen overnight.

mauritsvanrees:
This is good to go, I merge.
Would be good to have this in a release soon.  I may do that today, unless @tisto or @davisagli is earlier.

mauritsvanrees:
Before release, please approve and merge PR #1729.

PR #1729, opened 2023-11-02
Temporarily disable form memory limit checking for files and images.

mauritsvanrees:
This fixes a regression due to a low Zope form memory limit of 1MB used since Plone 6.0.7. You can use ``dos_protection`` settings in ``etc/zope.conf`` to change the limit. See https://github.com/plone/Products.CMFPlone/issues/3848 and https://github.com/zopefoundation/Zope/pull/1142.

### DIFF
--- a/src/plone/restapi/deserializer/__init__.py
+++ b/src/plone/restapi/deserializer/__init__.py
@@ -4,6 +4,9 @@ import json
 
 
 def json_body(request):
+    # TODO We should not read the complete request BODY in memory.
+    # Once we have fixed this, we can remove the temporary patches.py.
+    # See there for background information.
     try:
         data = json.loads(request.get("BODY") or "{}")
     except ValueError:
--- a/src/plone/restapi/patches.py
+++ b/src/plone/restapi/patches.py
@@ -1,0 +1,20 @@
+# TEMPORARY patch for low form memory limit introduced in Zope 5.8.4.
+# See https://github.com/plone/Products.CMFPlone/issues/3848
+# and https://github.com/zopefoundation/Zope/pull/1180
+# Should be removed once `plone.restapi.deserializer.json_body` no longer
+# reads the complete request BODY in memory.
+from ZPublisher import HTTPRequest
+
+import logging
+
+
+logger = logging.getLogger(__name__)
+_attr = "FORM_MEMORY_LIMIT"
+_limit = getattr(HTTPRequest, _attr, None)
+if _limit and _limit == 2**20:
+    setattr(HTTPRequest, _attr, 2**24)
+    logger.info(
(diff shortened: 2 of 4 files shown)


mauritsvanrees:

mauritsvanrees:

src/plone/restapi/patches.py
```diff
@@ -0,0 +1,20 @@
...
+
+import logging
+
+
+logger = logging.getLogger(__name__)
+_attr = "VALUE_LIMIT"
+_limit = getattr(ZopeFieldStorage, _attr, None)
+if _limit:
+    setattr(ZopeFieldStorage, _attr, None)
+    logger.info(
```
davisagli, line 16:
```suggestion
    logger.debug(
```
I don't feel strongly about this, but it feels like it's probably noise for most people.

src/plone/restapi/deserializer/__init__.py
```diff
@@ -4,6 +4,9 @@
 
 
 def json_body(request):
+    # TODO We should not read the complete request BODY in memory.
+    # Once we have fixed this, we can remove the temporary patches.py.
+    # See there for background information.
     try:
         data = json.loads(request.get("BODY") or "{}")
```
davisagli, line 11: Based on https://github.com/zopefoundation/Zope/blob/master/src/ZPublisher/HTTPRequest.py#L1061, I think we could do (untested):
```
body = request.get("BODYFILE")
data = {} if body is None else json.load(body)
```

This will not actually avoid reading the file into memory (https://pythonspeed.com/articles/json-memory-streaming/ makes it clear that json.load still does that) but would bypass the descriptor that enforces the VALUE_LIMIT.

davisagli (approved):
@mauritsvanrees Thanks for continuing to follow up on this.

I'm okay with this as is, but am also making a comment with a suggestion that may avoid the need for a monkeypatch.

src/plone/restapi/patches.py
mauritsvanrees: I don't feel strongly about this either, but I think it would be good to have this noticeable, because it may remind us developers that we still need to fix this instead of having this temporary patch.

src/plone/restapi/deserializer/__init__.py
mauritsvanrees: I thought `BODYFILE` would only be defined when an actual file is being uploaded, and not for example when you POST a login and password.  But that is not true.

It does not sound like a permanent solution, as this would still offer a way to potentially DOS the server.  But maybe very large uploads would still get stopped by one of the other limits.  And initial testing seems to work out.

Let me open a different PR, so we still have the current one in case there are problems.

Issue #1730, opened 2023-11-02
json_body should not read entire request BODY


mauritsvanrees:
So [this line](https://github.com/plone/plone.restapi/blob/9.1.1/src/plone/restapi/deserializer/__init__.py#L8) is bad:

    data = json.loads(request.get("BODY") or "{}")

With Zope 5.8.4+ this fails when the request BODY is larger than 1MB.
See discussion in https://github.com/plone/Products.CMFPlone/issues/3848 and https://github.com/zopefoundation/Zope/pull/1142.
See [my summary](https://github.com/plone/Products.CMFPlone/issues/3848#issuecomment-1790665336) of the history and the steps taken so far.

Temporarily fixed in https://github.com/plone/plone.restapi/pull/1729 by disabling a Zope form memory limit check.

gogobd:
I was patching this like this:

`…/plone/restapi/deserializer/__init__.py`
```
from plone.restapi.exceptions import DeserializationError

import json


def json_body(request):
    try:
        try:
            if hasattr(request, "BODYFILE"):
                data = request.get('BODYFILE').read()
            else:
                data = json.loads(request.get("BODY") or "{}")
        except:
            breakpoint()
    except ValueError:
        raise DeserializationError("No JSON object could be decoded")
    if not isinstance(data, dict):
        raise DeserializationError("Malformed body")
    return data


def boolean_value(value):
    """

    Args:
        value: a value representing a boolean which can be
               a string, a boolean or an integer
                   (usually a string from a GET parameter).

    Returns: a boolean

    """
    return value not in {False, "false", "False", "0", 0}
 
```
Buut as @mauritsvanrees pointed out: "We should not read the complete request BODY in memory." and so that's probably not helping. I wasn't able to come up with a test that would actually push a request through Zope's machinery to address this issue. I also think that a "fix" like the one I show here would not really eliminate the D.O.S. threat that potentially was the reason for introducing the limits in the first place. The only "good" fix would be to have every request body be a stream in all cases, but I wonder if that's feasible. Can we use a data structure that is a stream and behaves like "in memory data"?

gogobd:
Additionally we "fixed" the problem by upping the limits in zope.conf like this:

```
<dos_protection>
    form-memory-limit 1MB
    form-disk-limit 1GB
    form-memfile-limit 4KB
</dos_protection>
```

gogobd:
... maybe at one point we will have to live with "huge" data in memory.

[...]

PR #1731, opened 2023-11-02
Read json from request BODYFILE instead of BODY.

mauritsvanrees:
This fixes a regression due to a low Zope form memory limit of 1MB used since Plone 6.0.7. See [CMFPlone issue 3848](https://github.com/plone/Products.CMFPlone/issues/3848) and [Zope PR 1142](https://github.com/zopefoundation/Zope/pull/1142).

Fixes https://github.com/plone/plone.restapi/issues/1730

### DIFF
--- a/src/plone/restapi/deserializer/__init__.py
+++ b/src/plone/restapi/deserializer/__init__.py
@@ -17,7 +17,8 @@ def json_body(request):
             # Go back to the beginning.
             bodyfile.seek(0)
         try:
-            data = json.load(bodyfile)
+            body = bodyfile.read()
+            data = {} if not body else json.loads(body)
         except ValueError:
             raise DeserializationError("No JSON object could be decoded")
         finally:


mauritsvanrees:
@davisagli This might be the good direction, but lots of tests fail.
Locally (when running the tests from within the coredev 6.0 buildout) I even get an `OSError` because there are too many open files.

davisagli:
@mauritsvanrees Okay, let's go with https://github.com/plone/plone.restapi/pull/1729 for now and come back to this.

mauritsvanrees:
I have pushed a fix which helps: we need to make sure we read the bodyfile from the beginning.
I can't tell yet it the open files error goes away with this.

mauritsvanrees:
4 failures, 1 error.  Not bad, but still needs some work.  And it could mean just a few more test fixes, or it may mean possible breakage in third party code.  So even if we go for this, it may mean a major release.
So indeed perhaps first merge and release PR #1729, use that in Plone 6.0.8, and then come back to this one.
cc @tisto.

I stop for today.

mauritsvanrees:
The "too many open files" error is gone at least.

davisagli:
@mauritsvanrees Not sure if this is the reason for the remaining test failures, but I think what we need to do is not always seek back to 0, but make sure that we restore the file pointer to the position it was before reading. The ValueDescriptor in Zope does this: https://github.com/zopefoundation/Zope/blob/master/src/ZPublisher/HTTPRequest.py#L1377-1379

mauritsvanrees:
@davisagli I restore the file position now, but that was not it.
Problem is with the `@lock` endpoint, where the BODYFILE is empty.  `json.load` on an empty file gives a ValueError.  So now I read the BODYFILE before passing its contents on, and use a dictionary if the contents are empty.  That fixes the lock tests at least.

Yes, I said I would stop for today. ;-)

mauritsvanrees:

mauritsvanrees:
Okay, so this needs fixes in `plone.volto`.  Maybe only updates to the tests, I don't know.

Let's go with PR #1729 for now and use that in Plone 6.0.8.  It is approved already.  If no one merges and releases it, I intend to do so myself Monday, and make 6.0.8 final.

Afterwards we can continue on the current PR.

tisto:
@mauritsvanrees I merged #1729 and released it with plone.restapi 9.1.2. Feel free to ping me if there is anything else you would like me to do.

mauritsvanrees:
Thanks!

PR #1732, opened 2023-11-07
Does not test in Python 3.7

wesleybl:
Plone 5.2 no longer supports Python 3.7. See: https://github.com/plone/Products.CMFPlone/blob/5.2.14/setup.py#L25-L26


davisagli (approved):
Thanks! I see we already dropped support in the readme and setup.py classifiers, so this just looks like a remaining detail rather than a new change in policy.

PR #1733, opened 2023-11-07
Use plone.recipe.precompiler to generate mo files to test

wesleybl:
There are tests that need translation. If we started the instance for testing, the translation files were generated, causing the documentation tests to give a different result locally, since in CI the mo files did not exist.

### DIFF
--- a/base.cfg
+++ b/base.cfg
@@ -14,6 +14,7 @@ parts =
     deploy-to-heroku
     omelette
     zpretty
+    precompile
 develop = .
 sources-dir = extras
 auto-checkout =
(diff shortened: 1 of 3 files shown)


davisagli (approved):
Thanks, I had to do a similar fix for a project recently.

I think another way to solve this is to add
```
environment-vars =
    zope_i18n_compile_mo_files true
```

for the `test` section (and setting this environment variable also works when not using buildout). But this is also fine.

We should update the plone.restapi repository to use a pip-based installation instead of buildout, but that's a separate, bigger task.

PR #1736, opened 2023-11-24
add recordsproxy to controlpanel serializer so it handles schemas with new fields by returning field default

djay:


davisagli (commented):
It can be confusing when a default is calculated every time the field is read. For example if there's a field with a defaultFactory that returns the current userid, then it will look like a different value is stored depending on which user reads the value. For that reason, I would be inclined to require an explicit upgrade step if a new field is added that uses defaultFactory.

I would also suggest putting DefaultRecordsProxy in plone.registry so that it can be used by both plone.restapi and plone.api instead of duplicating it. That does make things a bit more complicated since plone.restapi needs to potentially handle old versions of Plone that don't have it.

djay:
> It can be confusing when a default is calculated every time the field is read. For example if there's a field with a defaultFactory that returns the current userid, then it will look like a different value is stored depending on which user reads the value. 

Well it does only fix the case of accessing hte control panel so the user will see their name and if they hit save then it persist.
If the registry information is used elsewhere like in another public api, then that code will have to make other decisions about how to handle missing information.
I'm also not sure this scenario is that common. do you know of an example @davisagli ?

> For that reason, I would be inclined to require an explicit upgrade step if a new field is added that uses defaultFactory.

Thats the case now that this is trying to solve. Control panels break until you upgrade your plugin which is confusing for admins. There is no nice explanation why, just a stack trace. And in addition its just extra work for the developer to go put in explicit upgrade steps that rerun your profile or add that specific missing data in the registry when they can just specify a default in the schema instead (the more intuitive thing to do)

A more complete solution is to have different versions of the schema interface linked to a particular installed version and use the old one until the upgrade step is run. Or maybe fields in a schema linked to a version? But all of this also adding more work for the developer.

Or we solve this by showing a nicer error to users asking them to upgrade the plugin?

> 
> I would also suggest putting DefaultRecordsProxy in plone.registry so that it can be used by both plone.restapi and plone.api instead of duplicating it. That does make things a bit more complicated since plone.restapi needs to potentially handle old versions of Plone that don't have it.

davisagli:
> Well it does only fix the case of accessing hte control panel so the user will see their name and if they hit save then it persist.

My concern is that they are seeing a value that is not actually persisted. If it's the value they want, they'll probably think they don't need to save.

It's not really a problem for a static default value, but more of a potential problem if there's a defaultFactory that returns different defaults at different times.

An unrelated question: It looks like the existing RecordsProxy already falls back to the field's `missing_value`. Is there a reason it doesn't work to use that when you add a new field?

djay:
> An unrelated question: It looks like the existing RecordsProxy already falls back to the field's `missing_value`. Is there a reason it doesn't work to use that when you add a new field?

You are right. the semantics of using this make more sense that using the default. Not sure developers would know to use missing_value but either way documentation would be needed.
The check=False would still need to be part of this change otherwise it won't work.

davisagli:
@djay Hmm, no, it's not that simple. It'll use the `missing_value` when reading, but throw an exception when trying to write the field if it hasn't been registered yet: https://github.com/plone/plone.registry/blob/master/plone/registry/recordsproxy.py#L46

The registry design stores a persistent field definition as part of a record; I expect it will be an uphill battle to get things to work correctly without an upgrade step to add records for the new fields.

djay:
> @djay Hmm, no, it's not that simple. It'll use the `missing_value` when reading, but throw an exception when trying to write the field if it hasn't been registered yet: https://github.com/plone/plone.registry/blob/master/plone/registry/recordsproxy.py#L46
> 

You are right again. my test doesn't probably test setting since the value is actually there.
This check could be change to check against the schema however, instead of the registry?

> The registry design stores a persistent field definition as part of a record; I expect it will be an uphill battle to get things to work correctly without an upgrade step to add records for the new fields.

It does? I wasn't aware. if so why can't the recordsproxy use that to check against instead of the current Interface? Then it will be checking against the schema that was there at the time it was installed which would also solve the problem I think?

djay:
@davisagli my test did delete the record so does show that the value can be set again in that case. I've changed it to use the missing value and the test passes.
So is that the solution we want?

PR #1746, opened 2024-01-30
Improve the block transforms to include also anchors when resolving uid back and forth

sneridagh:
Now it covers the use case if an anchor is present.

### DIFF
--- a/news/1746.bugfix
+++ b/news/1746.bugfix
@@ -1,0 +1,1 @@
+Improve the block transforms to include also anchors when resolving uid back and forth @sneridagh
--- a/news/1746.bugfix
+++ b/news/1746.bugfix
@@ -1,1 +1,1 @@
-Improve the block transforms to include also anchors when resolving uid back and forth @sneridagh
+Fixed the logic for converting public URLs to and from internal UID-based URLs. Now if the URL includes a fragment, it is preserved. @sneridagh
--- a/src/plone/restapi/deserializer/utils.py
+++ b/src/plone/restapi/deserializer/utils.py
@@ -25,11 +25,13 @@ def path2uid(context, link):
         )
 
     # handle edge-case when we have non traversable path like /@@download/file
-    if "/@@" in path:
-        path, suffix = path.split("/@@", 1)
-        suffix = "/@@" + suffix
-    else:
-        suffix = ""
+    SUFFIXES = ["/@@", "#"]
+    suffix = ""
+    for suffix_separator in SUFFIXES:
+        if suffix_separator in path:
+            path, suffix = path.split(suffix_separator, 1)
+            suffix = suffix_separator + suffix
+
     obj = portal.unrestrictedTraverse(path, None)
     if obj is None or obj == portal:
         return link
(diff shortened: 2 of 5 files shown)


sneridagh:

src/plone/restapi/deserializer/utils.py
```diff
@@ -25,11 +25,13 @@ def path2uid(context, link):
...
-    if "/@@" in path:
-        path, suffix = path.split("/@@", 1)
-        suffix = "/@@" + suffix
-    else:
-        suffix = ""
+    SUFFIXES = ["/@@", "#"]
+    suffix = ""
+    for suffix_separator in SUFFIXES:
+        if suffix_separator in path:
+            path, suffix = path.split(suffix_separator, 1)
```
davisagli, line 32: @sneridagh You are overwriting `suffix` here. So if both separators are present, only one of the suffixes will be kept.

src/plone/restapi/serializer/utils.py
```diff
@@ -23,13 +23,15 @@ def resolve_uid(path):
...
         return path, None
 
-    uid, suffix = match.groups()
+    uid, suffix, anchor = match.groups()
     brain = uuidToCatalogBrain(uid)
     if brain is None:
         return path, None
     href = brain.getURL()
     if suffix:
         return href + "/" + suffix, brain
```
davisagli, line 32: What if there is a suffix and an anchor?

davisagli (changes requested):

sneridagh:
I know, it’s one or the other, never both. Do you have a use case in mind?

Víctor Fernández de Alba
Github/Twitter: sneridagh

sneridagh:
@davisagli I think not, because a suffix is something that matches a `/@@` BrowserView. A BrowserView with an anchor, in RESTAPI world? Could be in ClassicLand, but not in RESTAPI. We could make it work, but I think that without a use case, it's premature optimization, could be that there's never a use case for that.

davisagli:
@sneridagh Although this is plone.restapi, the URLs that an editor can enter are not API URLs. It could be a browser view that renders HTML including anchors, sure. Here we need to be careful and pass through unchanged anything that is not the path to a content item.

src/plone/restapi/serializer/utils.py
sneridagh: Done

src/plone/restapi/deserializer/utils.py
sneridagh: It seemed wrong, but it's fine, I added a test. Ported the pattern to the serializer too.

news/1746.bugfix
```diff
@@ -0,0 +1 @@
+Improve the block transforms to include also anchors when resolving uid back and forth @sneridagh
```
davisagli, line 1:
```suggestion
Fixed the logic for converting public URLs to and from internal UID-based URLs. Now if the URL includes a fragment, it is preserved. @sneridagh
```

davisagli (changes requested):
Uh, the tests are failing...

davisagli (approved):
Okay, let's see what jenkins says...

davisagli:

PR #1748, opened 2024-02-15
Fix serializer block with anchors, remove trailing slash before the anchor

sneridagh:
@davisagli we overlooked this one :( the roundtrip was bugged because of the trailing slash. Please take a look at the regex, if it's fine like this.

### DIFF
--- a/news/1748.bugfix
+++ b/news/1748.bugfix
@@ -1,2 +1,1 @@
-Fix blocks resolveuid serializer with suffixes, remove trailing slash added before anchors
-[sneridagh]
+Fixed an edge case in the blocks resolveuid transforms with a trailing slash before a fragment. @sneridagh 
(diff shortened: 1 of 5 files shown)


tisto (approved):
LGTM. However I'd like @davisagli to have a second look at it.

news/1748.bugfix
```diff
@@ -0,0 +1,2 @@
+Fix blocks resolveuid serializer with suffixes, remove trailing slash added before anchors
+[sneridagh]
```
davisagli, line 2:
```suggestion
Fixed an edge case in the blocks resolveuid transforms with a trailing slash before a fragment. @sneridagh 
```

davisagli:
I added a related fix for the deserializer too. Checking what jenkins says...

davisagli (approved):

tisto:
@davisagli @sneridagh released with https://pypi.org/project/plone.restapi/9.4.2/#section-1.

PR #1749, opened 2024-02-16
Add the site timezone info to the @site endpoint

folix-01:


folix-01:
@jensens @mauritsvanrees  Hi, here I have the only test run that fails, it's the test for a python39. 
The problem is that the restapi response and the expected result have a different dictionary key-value order.
I don’t really understand why it happens, any idea? 
Screenshot 2024-02-16 at 17 27 48

wesleybl:
@folix-01  did you run the tests locally and `plone.site_timezone` was `Europe/Berlin` in the resp file?

stevepiercy:
@folix-01 the [CI's error message is above your screenshot with what happened and how to solve it](https://github.com/plone/plone.restapi/actions/runs/7933333589/job/21661857837#step:11:17).

folix-01:
> @folix-01 did you run the tests locally and `plone.site_timezone` was `Europe/Berlin` in the resp file?

I didn't run the tests locally so as I can't make a buildout

folix-01:

src/plone/restapi/services/site/get.py
```diff
@@ -38,6 +39,7 @@ def __call__(self, expand=False):
                 "plone.site_logo": site_settings.site_logo and getSiteLogo() or None,
                 "plone.robots_txt": site_settings.robots_txt,
                 "plone.allowed_sizes": image_settings.allowed_sizes,
+                "plone.site_timezone": default_timezone(self.context),
```
erral, line 42: I would say that we should get this setting from registry and not from the `default_timezone` function.

According to [its implementation](https://github.com/plone/plone.app.event/blob/master/plone/app/event/base.py#L449-L455) the function is designed to return the correct timezone even per-user, so if we call this endpoint with different users, we could get different results.

I would say that if we add the site_timezone setting, this should be the one set by the `plone.portal_timezone` setting.

I am OK on returning also the users own timezone, but I would return that in a different setting and do not call it "site_timezone" but "user_timezone".

erral (changes requested):

src/plone/restapi/services/site/get.py
stevepiercy: Renaming the key does not help. The implementation ambiguously returns either the user or site timezone. It would be better to name the key as `plone.timezone` because it is totally ambiguous from this implementation.

Thus I would go with @erral's suggestion. Get the value from the registry. However, I would return the user's timezone from the user endpoint, not the site endpoint.

src/plone/restapi/services/site/get.py
```diff
@@ -38,11 +42,31 @@ def __call__(self, expand=False):
                 "plone.site_logo": site_settings.site_logo and getSiteLogo() or None,
                 "plone.robots_txt": site_settings.robots_txt,
                 "plone.allowed_sizes": image_settings.allowed_sizes,
+                "plone.timezone": self.plone_timezone(),
```
erral, line 45: When adding the other settings (site_logo, robots.txt, etc) we kept the registry key name in the output JSON. I would do the same here and use `plone.portal_timezone` instead of `plone.timezone`

folix-01: changed as required

src/plone/restapi/services/site/get.py
folix-01: Ok, I agree, changed as required

stevepiercy (commented):
LGTM, but I defer to @erral as he has greater familiarity with this code than I.

erral (approved):

folix-01:
@erral @stevepiercy  How is the contribution workflow working? Do I have to do the merge personally or wait the release manager to do it?

erral:
nope, @tisto or @davisagli will look at it and merge

folix-01:

folix-01:

folix-01:

news/1749.feature
```diff
@@ -0,0 +1 @@
+Add the site tiemezone to the @site endpoint return result. [folix-01]
```
davisagli, line 1:
```suggestion
Add the site timezone to the @site endpoint return result. @folix-01
```

davisagli (approved):
LGTM. Thanks!

PR #1751, opened 2024-02-16
Give Site Administrator permission to add users to groups

### LINKED ISSUE #1750: Site Administrator cannot add user to group
In #1712, the ability for the Site Administrator to manage users was added. But I noticed that he is unable to add a user to a group. He receives an `Unauthorized` error.

The groups service uses the addMember method:

https://github.com/plone/plone.restapi/blob/16325f20b927f28ff03762687862dcaea9944da5/src/plone/restapi/services/groups/update.py#L125

This method checks if the user has the `Manage Users` permission:

https://github.com/plone/Products.PlonePAS/blob/d148e7eb50e52cddf954406c06026fb5e60f3da7/src/Products/PlonePAS/tools/groupdata.py#L223

But he doesn't.

@davisagli  informed me that it is not a good idea to give `Manager Users` to the Site Administrator. See:

https://github.com/plone/plone.restapi/pull/1712#discussion_r1377057545

So the ideal is to do the same thing that was done in the other PR and do it with Plone Classic, correct?

Plone Classic use `addPrincipalToGroup`:

https://github.com/plone/Products.CMFPlone/blob/f3096371d80c3ca8a5829c12976bd0ab2d2c08a4/Products/CMFPlone/controlpanel/browser/usergroups_groupmembership.py#L44

So we should use it here too.

wesleybl:
The `addMember` method requires the user to have the `Manage Users` permission but the Site Administrator does not have this permission. Then we use the `addPrincipalToGroup` method, which does not require this permission.

fixes #1750

### DIFF
--- a/news/1750.bugfix
+++ b/news/1750.bugfix
@@ -1,0 +1,1 @@
+Give Site Administrator permission to add users to groups. @wesleybl
--- a/news/1750.bugfix
+++ b/news/1750.bugfix
@@ -1,1 +1,1 @@
-Give Site Administrator permission to add users to groups. @wesleybl
+Fixed the permission check for adding users to groups and removing users from groups, so that it is allowed for users with the Site Administrator role. @wesleybl
--- a/src/plone/restapi/services/groups/add.py
+++ b/src/plone/restapi/services/groups/add.py
@@ -71,7 +71,7 @@ class GroupsPost(Service):
         # Add members
         group = gtool.getGroupById(groupname)
         for userid in users:
-            group.addMember(userid)
+            gtool.addPrincipalToGroup(userid, groupname)
 
         self.request.response.setStatus(201)
         self.request.response.setHeader(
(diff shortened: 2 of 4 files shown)


wesleybl:

jensens:
While this fixes the immediate problem, I would question question why the two methods have different permissions - and if it would not be better to fix it there than working around here?

wesleybl:
@jensens this is the same solution used in the Classic UI. See @davisagli 's opinion on using `Manage Users`:

https://github.com/plone/plone.restapi/pull/1712#discussion_r1377057545

jensens:
> @jensens this is the same solution used in the Classic UI. See @davisagli 's opinion on using `Manage Users`:
> 
> [#1712 (comment)](https://github.com/plone/plone.restapi/pull/1712#discussion_r1377057545)

got it, indeed.

jensens (approved):

davisagli (approved):
Thanks!

news/1750.bugfix
```diff
@@ -0,0 +1 @@
+Give Site Administrator permission to add users to groups. @wesleybl
```
davisagli, line 1:
```suggestion
Fixed the permission check for adding users to groups and removing users from groups, so that it is allowed for users with the Site Administrator role. @wesleybl
```

PR #1752, opened 2024-02-20
Support 'not' queries with @search endpoint.

maethu:
This PR adds support to exclude certain items from the result set using `not` for DateIndex, FieldIndex and KeywordIndex. 

Example request:
```
GET /Plone/@search?Subject.not=important&path.query=%2FPlone&path.depth=-1 HTTP/1.1
```

Returns all items not having the subject "important".


davisagli (commented):
@maethu This is already supported by the `@querystring-search` endpoint using the `isNot` and `none` operations (see https://github.com/plone/plone.app.querystring/pull/110). Is there a reason it's needed for the `@search` endpoint as well?

maethu:
> @maethu This is already supported by the `@querystring-search` endpoint using the `isNot` and `none` operations (see [plone/plone.app.querystring#110](https://github.com/plone/plone.app.querystring/pull/110)). Is there a reason it's needed for the `@search` endpoint as well?

Thanks very much, @davisagli, I'm aware of that. On my project, I use the `@search` endpoint for my search form because the form state management requires much less code via `@search` endpoint than it does via `@querystring-search`. I also don't have to update a deeply nested state. On top, it's a way more readable URL, mostly because it's shorter. 

No hard feelings if it does not make sense in plone.restapi. I can keep my patch in my project. I just thought maybe it would make sense to give this upstream.

davisagli:
@maethu Thanks, I'll take another look and also ask @tisto for his opinion. If we do decide to add it, it will need documentation.

tisto:
@davisagli let's discuss this in the next days. @maethu it would be awesome to have our docs updated to reflect this change. Without proper docs, we can not merge any PR.

PR #1757, opened 2024-02-28
add new @login endpoint to return available external login options

erral:
This PR adds a generic `GET /@login` endpoint that would expose links to additional login providers in a Plone site.

This way the endpoints added in packages like [pas.plugins.authomatic](https://github.com/collective/pas.plugins.authomatic/blob/main/src/pas/plugins/authomatic/services/configure.zcml#L8-L14) and [pas.plugins.oidc](https://github.com/collective/pas.plugins.oidc/blob/main/src/pas/plugins/oidc/services/login/configure.zcml#L7-L13) could be exchanged with adapter registrations and allow both products to be installed in the same Plone site for its use.

I am open to discuss whether the adapter needs to adapt just the IPloneSiteRoot object or perhaps may be a multiadapter including the request (perhaps to register the adapter for a given IThemeSpecific interface) or any other things.


erral:

news/1757.feature
```diff
@@ -0,0 +1,2 @@
+Add a @login endpoint to get external login services' links
```
stevepiercy, line 1:
```suggestion
Add a @login endpoint to get external login services' links. @erral
```

news/1757.feature
```diff
@@ -0,0 +1,2 @@
...
+[erral]
```
stevepiercy, line 2:
```suggestion
```

stevepiercy (changes requested):
This needs tests and documentation, similar to other endpoints. https://6.docs.plone.org/plone.restapi/docs/source/endpoints/index.html

erral:
> This needs tests and documentation, similar to other endpoints. https://6.docs.plone.org/plone.restapi/docs/source/endpoints/index.html

Yes, I am preparing that.

stevepiercy (changes requested):

erral:

stevepiercy (approved):
I decided it would be faster, easier, and less frustrating to make the changes directly instead of a proper review. I hope that is OK.

erral:
> I decided it would be faster, easier, and less frustrating to make the changes directly instead of a proper review. I hope that is OK.

Thanks! :heart:

erral:
I am investigating why the output for the documentation is empty.

stevepiercy:
@erral what output? The preview looks good: https://deploy-preview-1757--plone-restapi.netlify.app/endpoints/login.html

erral:
@stevepiercy the last output, this one:

```
HTTP/1.1 200 OK
Content-Type: application/json

{
    "options": []
}
```

That should not be empty :/

src/plone/restapi/tests/test_documentation.py
```diff
@@ -785,6 +811,12 @@ def test_documentation_jwt_logout(self):
         )
         save_request_and_response_for_docs("jwt_logout", response)
 
+    def test_documentation_external_doc_links(self):
+        response = self.api_session.get(
+            f"{self.portal.absolute_url()}/@login",
+        )
+        save_request_and_response_for_docs("external_authentication_links", response)
```
stevepiercy, line 818: With my rename of the docs file, should this be changed as well?

```suggestion
        save_request_and_response_for_docs("login", response)
```

stevepiercy (commented):
Question for @erral

stevepiercy:
It's not an issue with the docs, because the docs only pull from this file:

https://github.com/plone/plone.restapi/blob/erral-login-options/src/plone/restapi/tests/http-examples/external_authentication_links.resp

Can you show what the content of that file should be?

docs/source/endpoints/login.md
```diff
@@ -0,0 +1,68 @@
...
+
+In the adapter, the add-on needs to return the list of external links and some metadata, including the `id`, `title`, and name of the `plugin`.
+
+An example adapter would be the following, in a file named {file}`adapter.py`:
+
+```python
+from zope.component import adapts
+from zope.interface import implementer
+
+@adapts(IPloneSiteRoot)
```
davisagli, line 31: Should be adapter, not adapts

docs/source/endpoints/login.md
```diff
@@ -0,0 +1,68 @@
...
+                "plugin": "github",
+                "url": "https://some.example.com/login-authomatic/github",
+            },
+        ]
+```
+
+With the corresponding ZCML stanza, in the corresponding {file}`configure.zcml` file:
+
+```xml
+<adapter factory=".adapter.MyExternalLinks" />
```
davisagli, line 54: It needs to be a named adapter, otherwise it will conflict with the same adapter from another addon.

src/plone/restapi/interfaces.py
```diff
@@ -240,3 +240,14 @@ class IBlockVisitor(Interface):
...
+
+class IExternalLoginProviders(Interface):
+    """An interface needed to be implemented by providers that want to be listed
+    in the @login endpoint
+    """
+
+    def get_providers():
+        """
+        return a list of login providers, with its id, title, plugin and url
+        """
```
davisagli, line 253: Maybe this should be defined in plone.base instead of plone.restapi, so that an addon doesn't need to depend on plone.restapi just to register this adapter and provide information about what providers it supports?

On the other hand, that would make it harder to backport support for this feature to older versions of Plone.

I don't feel strongly about it but in general, plone.restapi's purpose is to provide API access to features that exist at a lower level and might also be accessed via other interfaces.

davisagli (commented):
I like this idea in general.

Things I'm wondering:
- What about a site that wants to let the user choose between login from external providers and Plone's standard login?
- What if the integrator wants more control over the order of different login options?

So I'm thinking that maybe there should be a registry setting with a list of login providers to show, and this endpoint could use the info from the registry setting plus the details from the IExternalLoginProviders adapters. But -- this can probably be added later, so I'm not necessarily asking you to do it as a condition for adding this endpoint.

src/plone/restapi/interfaces.py
erral: the whole idea of the endpoint is to avoid duplicated `@login` endpoints such the ones we have in pas.plugins.authomatic and pas.plugins.oidc to support volto-authomatic

I agree that perhaps this should go in plone.base somewhere and integrate it with Plone's standard login page.

Perhaps we can open an issue in Plone and work on that later on?

src/plone/restapi/tests/test_documentation.py
erral: No, this is for the test-generated request and response file. I would leave this as it is, because we already have `login.req` file in the http-examples folder

erral:
> It's not an issue with the docs, because the docs only pull from this file:
> 
> https://github.com/plone/plone.restapi/blob/erral-login-options/src/plone/restapi/tests/http-examples/external_authentication_links.resp
> 
> Can you show what the content of that file should be?

I have already fixed this. There was a bug in my code :smiling_imp:

erral:

stevepiercy (approved):
Preview of docs with update look good: https://deploy-preview-1757--plone-restapi.netlify.app/endpoints/login.html

erral:

erral:

stevepiercy:
@erral is there any reason not to merge this PR?

davisagli:
@erral @stevepiercy Is there any prototype of using this API from a frontend? That would provide some confidence that we've implemented the right API. It's easier to make changes before we release something.

I'm not sure if you plan to update Volto to use this, or just use it in specific projects. If Volto will use it, then we also need to coordinate with @robgietema to support it in Nick.

erral:
Right know this `GET` operation over the `@login` endpoint is implemented by [pas.plugins.authomatic](https://github.com/collective/pas.plugins.authomatic/blob/main/src/pas/plugins/authomatic/services/configure.zcml#L8-L14) but only returns the plugins available there (see https://github.com/collective/pas.plugins.authomatic/blob/main/src/pas/plugins/authomatic/services/login.py).

The endpoint is consumed by [volto-authomatic](https://github.com/collective/volto-authomatic/blob/main/src/actions/auth/authOptions.js) to list the login options.

But this implementation works *only* with pas.plugins.authomatic.

I had the requirement to use pas.plugins.authomatic and pas.plugins.oidc in the same site. Moreover I had the requirement to use 2 different oidc plugins. 

So first I generalized pas.plugins.oidc to support [multiple oidc based plugins](https://github.com/collective/pas.plugins.oidc/pull/48).

Then I needed volto-authomatic to support both plugins, so I brought the `@login` endpoint implementation here, and prepare PRs for both pas.plugins.authomatic and pas.plugins.oidc to support the adapter lookup implemented here (see https://github.com/collective/pas.plugins.authomatic/pull/73 and https://github.com/collective/pas.plugins.oidc/pull/47).

So, the idea is:

- plone.restapi to provide a way for other add-ons to regiser external login services, as a central point.
- volto-authomatic is the consumers of such endpoint

davisagli:
@erral Thanks for the explanation and pointers. It makes sense to me.

I could imagine that in the future, maybe Volto itself could check this endpoint and show the login providers on the default login page. But it's good enough to be able to use it in add-ons for now.

I don't see a problem with Nick support. Nick doesn't support external auth providers yet, so it should be very easy to implement the endpoint which returns an empty list :)

I'll add a review with a couple minor comments.

src/plone/restapi/interfaces.py
```diff
@@ -240,3 +240,14 @@ class IBlockVisitor(Interface):
 
     def __call__(self, block):
         """Return an iterable of sub-blocks found inside `block`."""
+
+
+class IExternalLoginProviders(Interface):
```
davisagli, line 245: @erral Maybe we should just call this ILoginProviders? An add-on could also add a custom login form that is implemented within Plone rather than an external service, and want to include it in the list

docs/source/endpoints/login.md
```diff
@@ -0,0 +1,71 @@
...
+
+When you install one of these add-ons, it modifies the login process, directing the user to third party services.
+
+To expose the links provided by these add-ons, `plone.restapi` provides an adapter based service registration.
+It lets those add-ons know that the REST API can use those services to authenticate users.
+This will mostly be used by frontends that need to show the end user the links to those services.
+
+To achieve that, third party products need to register one or more adapters for the Plone site root object, providing the `plone.restapi.interfaces.IExternalLoginProviders` interface.
+
+In the adapter, the add-on needs to return the list of external links and some metadata, including the `id`, `title`, and name of the `plugin`.
```
davisagli, line 23: What's the difference between `id` and `plugin`? Why do we need both?

erral: This is on purpose to support the case of a plugin added twice. Think of 2 `oidc` plugins in a site. They would be different in `id` but same in `plugin`, like a way to mark the "plugin type" in `plugin`.

erral: That's what I do, for instance, in my PR on pas.plugins.oidc: https://github.com/collective/pas.plugins.oidc/blob/use-adapter/src/pas/plugins/oidc/services/login/get.py#L23-L26

davisagli: @erral Got it. I didn't understand from the docs what `plugin` meant. I'll update the example below to show "pas.plugins.authomatic" as the plugin to make it a bit clearer.

docs/source/endpoints/login.md
```diff
@@ -0,0 +1,71 @@
...
+class MyExternalLinks:
+    def __init__(self, context):
+        self.context = context
+
+    def get_providers(self):
+        return [
+            {
+                "id": "myprovider",
+                "title": "Provider",
+                "plugin": "myprovider",
```
davisagli, line 42:
```suggestion
                "plugin": "pas.plugins.authomatic",
```

docs/source/endpoints/login.md
```diff
@@ -0,0 +1,71 @@
...
+                "url": "https://some.example.com/login-url",
+            },
+            {
+                "id": "github",
+                "title": "GitHub",
+                "plugin": "github",
```
davisagli, line 48:
```suggestion
                "plugin": "pas.plugins.authomatic",
```

davisagli (approved):

PR #1762, opened 2024-03-14
Bump all the versions in GitHub workflows

stevepiercy:
flake8 is not ready for Python 3.12, so it stays at 3.11 for now, unless another tool replaces it.

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -7,15 +7,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.8]
+        python-version: [3.12]
 
     steps:
       # git checkout
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       # python setup
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
           cache: pip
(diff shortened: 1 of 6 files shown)


stevepiercy:

jensens:
As this was already before this PR the case its not about this PR: Having a matrix with just one Python version is ... pointless?

stevepiercy:
@jensens please have another look. The matrix is preserved for tests, but not for things that only need to run once on a single version of Python, such as black, docs, and so on.

jensens:
I.e. for black there is a matrix with one entry. It does not make much of a difference - so lets keep it -, but it is pointless.

screenshot_2024-03-14_18:18:04_selection

stevepiercy:
Oh, I understand your point now. Well, maybe `matrix` is a poor choice of key, and [`env`](https://docs.github.com/en/actions/learn-github-actions/variables#defining-environment-variables-for-a-single-workflow) would be better. Either way, it allows to set it in one place and reuse it in multiple places.

- https://github.com/plone/plone.restapi/blob/main/.github/workflows/black.yml#L17
- https://github.com/plone/plone.restapi/blob/main/.github/workflows/black.yml#L20

I can change it to use `env`, if you prefer. Please let me know.

davisagli (approved):
@jensens I see your point, but also the matrix key affects the name of the workflow check, which has implications for how the branch protection is applied. Let's keep it as is, or at least address it separately.

PR #1766, opened 2024-03-27
In Plone 6, uses plone.textindexer to add block texts to the SearchableText index

### LINKED ISSUE #1744: IBlocks SearchableText indexer bypasses plone.textindexer behavior
The `SearchableText_blocks` indexer combines searchable text from blocks with the result of calling the base searchable text extractor, which is hardcoded to be `plone.app.contenttypes.indexers.SearchableText`. However, if the `plone.textindexer` behavior is active, it will be bypassed by this, so `IDynamicTextIndexExtender` adapters won't be used.

Given the way the component architecture handles adapter lookups, there may be no clean solution to call "whatever would have been called for a searchable text indexer if the `IBlocks` one wasn't in place". Another solution would be to turn the logic upside-down and implement the `IBlocks` indexer as an `IDynamicTextIndexExtender` adapter, which would require the `plone.textindexer` behavior to be implied by using Blocks.

I'm not sure what the best or cleanest solution might be. However, Blocks competing with `plone.textindexer` for the actual indexer feels unexpected and possible solutions feel hacky.

wesleybl:
This ensures that behaviors can add fields to SearchableText with `plone.textindexer`.

`IBlocks` now inherits from `IDexterityTextIndexer`, so there is no need to add the `plone.textindexer` behavior to the types to enable indexing. The content simply needs to have the `IBlocks` behavior.

fixes #1744

### DIFF
--- a/news/1744.bugfix
+++ b/news/1744.bugfix
@@ -1,0 +1,1 @@
+In Plone 6, uses ``plone.textindexer`` to add block texts to the SearchableText index, instead of ``plone.indexer``. This ensures that behaviors can add fields to SearchableText with ``plone.textindexer``. @wesleybl
--- a/src/plone/restapi/behaviors.py
+++ b/src/plone/restapi/behaviors.py
@@ -1,18 +1,19 @@
 from plone.autoform.interfaces import IFormFieldProvider
 from plone.restapi import _
-from plone.restapi import HAS_PLONE_6
 from plone.schema import JSONField
 from plone.supermodel import model
+from zope.interface import Interface
 from zope.interface import provider
 
 import json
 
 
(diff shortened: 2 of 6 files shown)


wesleybl:

wesleybl:

wesleybl:
@davisagli is it necessary to remove the Plone Jenkins CI job - pull-request-6.1-3.8 from the repository configuration?

davisagli (approved):
@wesleybl This looks good. I made a couple simplifications to the conditional imports -- see what you think.

davisagli:

wesleybl:
> I made a couple simplifications to the conditional imports -- see what you think.

@davisagli better. Thanks!

Issue #943, opened 2020-05-08
Password reset fails when using email as login


AlexBueckig:
If you use the email as username to login, the password reset fails saying that the `old_password` provided is wrong.

On further inspection I noticed that the username used here https://github.com/plone/plone.restapi/blob/dcd8587bcc8f90e0f445aa4888b9c9d042166eea/src/plone/restapi/services/users/add.py#L324-L326 is always the plone username no matter what authentication option is enabled. Using `target_user.getUser().getUserName()` , which is declared here, https://github.com/plone/plone.restapi/blob/dcd8587bcc8f90e0f445aa4888b9c9d042166eea/src/plone/restapi/services/users/add.py#L254 instead of the created variable username solves the issue since it always returns the correct loginname, etiher email or plone username, regarding to what option you choose in the settings.

PR #1767, opened 2024-03-28
Fix password reset when using email as login

Hrittik20:
Fixes: #943

### DIFF
--- a/src/plone/restapi/services/users/add.py
+++ b/src/plone/restapi/services/users/add.py
@@ -341,7 +341,7 @@ class UsersPost(Service):
                 )
 
             check_password_auth = pas.authenticate(
-                username, old_password.encode("utf-8"), self.request
+                target_user, old_password.encode("utf-8"), self.request
             )
             if not check_password_auth:
                 return self._error(
--- a/src/plone/restapi/services/users/add.py
+++ b/src/plone/restapi/services/users/add.py
@@ -341,7 +341,9 @@ class UsersPost(Service):
                 )
 
             check_password_auth = pas.authenticate(
-                target_user.getUser().getUserName(), old_password.encode("utf-8"), self.request
+                target_user.getUser().getUserName(),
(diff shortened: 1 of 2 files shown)


src/plone/restapi/services/users/add.py
```diff
@@ -341,7 +341,7 @@ def update_password(self, data):
                 )
 
             check_password_auth = pas.authenticate(
-                username, old_password.encode("utf-8"), self.request
+                target_user.getUser().getUserName(), old_password.encode("utf-8"), self.request
```
stevepiercy, line 344: This needs to pass lint, specifically black is failing: https://github.com/plone/plone.restapi/actions/runs/8472294504/job/23214126598?pr=1767

Also I don't know whether this continues to allow use of the username, as well as the email. In fact, I don't think this is the correct fix for this, as the previous line requires a user to be authenticated.

stevepiercy (dismissed):

stevepiercy (commented):
This still needs review by a core contributor.

I wonder whether we should have a unit test for both email and username? But let's see what happens in review.

src/plone/restapi/services/users/add.py
stevepiercy: I am not entirely sure. I think we need another endpoint, or adapt an existing one, or work with the permissions, to do what we need in Volto, because this endpoint requires authentication. I haven't thought through the details.

davisagli (approved):
Looks right to me.

You haven't followed the instructions from the bot to start the full Plone test suite with this change. I'll do that now.

davisagli:
> You haven't followed the instructions from the bot to start the full Plone test suite with this change.

Not your fault, the bot didn't post the message it usually does for some reason. Maybe because this PR was submitted from a fork instead of from a branch inside the plone/plone.restapi repository.

davisagli:

PR #1772, opened 2024-04-03
image_scales in serializer are returned as json_compatible format

cekk:
This lead to problems when for example we have a custom service that serialize some blocks and do not force a json_compatible formatting to the returned data.

I have a service that take some blocks, serialize them and return the result.
But image_scales are a PersistentMapping instance and breaks json formatting.

For content-types serialization or other services like @search this is not a problem because there is something that force json compatibility.

### DIFF
--- a/news/1772.bugfix
+++ b/news/1772.bugfix
@@ -1,0 +1,1 @@
+image_scales in serializer are returned as json_compatible format. @cekk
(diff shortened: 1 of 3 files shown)


cekk:

davisagli (approved):
@cekk Thanks, LGTM

PR #1776, opened 2024-04-30
Fix event dates in docs

### LINKED ISSUE #1773: Documentation of event content type does not return a timezone
The Event content type response returns `start` and `end` datetimes with a timezone:

```
  "end": "2023-12-31T12:00:00+00:00", 
  "start": "2023-01-01T11:00:00+00:00", 
```

However, in the documentation of the Event content type response, the `start` and `end` datetimes do not display a timezone indicator.

https://6.docs.plone.org/plone.restapi/docs/source/endpoints/content-types.html#event

```
    "end": "2013-01-01T12:00:00",
    "start": "2013-01-01T10:00:00",
```

@davisagli explained that this is due to a mock in the tests to avoid grabbing the current datetime that would result in a failure.

The documentation should reflect the actual response, not the mock.

davisagli:
Fixes #1773


stevepiercy (approved):
Thank you! I wish the rest of the timezone stuff were this simple.

PR #1789, opened 2024-05-17
Add endpoint that uses plone.volto `block_types` indexer (wip)

jnptk:
# ToDo
- [ ] Add page title data to response


src/plone/restapi/services/blocktypes/get.py
```diff
@@ -0,0 +1,39 @@
...
+    def reply(self):
+        catalog = api.portal.get_tool(name="portal_catalog")
+        request_body = self.request.form
+        result = {}
+
+        if request_body.get("blocktypes") != "":
+            blocktypes = request_body.get("blocktypes").split(",")
+
+            for blocktype in blocktypes:
+                brains = catalog.searchResults(block_types=blocktype)
```
jensens, line 17: I am sure this can be optimized to one query with removal of the for loop.

src/plone/restapi/services/blocktypes/get.py
```diff
@@ -0,0 +1,39 @@
...
+                result[blocktype] = Counter()
+
+                for brain in brains:
+                    obj = brain.getObject()
```
jensens, line 21: The above query could ask the `object_provides` index for the (marker) interface of the `volto.blocks` behavior. This would several an object wake ups.

davisagli (commented):
@jonaspiterek Is there a compelling reason to add a new service for this? It should be possible to query the block_types index using the existing @querystring-search service (we might need to add a criterion for it in the querystring registry settings)

src/plone/restapi/services/blocktypes/get.py
jonaspiterek: I just tried the `object_provides` index for the marker interface of the `volto.blocks` behavior like so: `brains = catalog(object_provides=IBlocks.__identifier__)` which returns brains just like the `block_types` index. And so I am asking my self: Doesn't this make the `block_types` index completely unnecessary? Because from the brains I get the object and from there I get the block values.

/cc @davisagli

davisagli: @jonaspiterek You can do that but then you end up loading the full object for all the content items that have blocks, and loading the full object is slower than loading the metadata that is stored as part of the catalog. The block_types index exists for the use case where you want to find all the items that have a particular type of block, without actually loading the full objects.

Also the block_types index [properly finds blocks that are nested inside other blocks](https://github.com/plone/plone.volto/blob/main/src/plone/volto/indexers.py#L44), which your code doesn't do yet.

davisagli: @jonaspiterek Btw if you just want to find out how many items have each type of block, you can also get that efficiently from the index like this (if I recall correctly): `api.portal.get_tool("portal_catalog").Indexes["block_types"].uniqueValues(withLengths=1)`

jonaspiterek:
@davisagli The blocktypes_indexer does not return as much information as I want to have so using the @querystring endpoint doesn't do the trick. I used functions from the indexer to make the endpoint similar to the indexer.

Making a request like: localhost:8080/Plone/@blocktypes?blocktypes=slate,image,introduction,separator,girdBlock

returns:

```
{
    "image": {
        "Event": {},
        "News Item": {
            "/Plone/news-item": 1
        },
        "Page": {
            "/Plone/page": 1
        },
        "Site": {}
    },
    "introduction": {
        "Event": {},
        "News Item": {},
        "Page": {
            "/Plone/page": 2
        },
        "Site": {}
    },
    "separator": {
        "Event": {},
        "News Item": {},
        "Page": {
            "/Plone/page": 4
        },
        "Site": {}
    },
    "slate": {
        "Event": {
            "/Plone/event": 1
        },
        "News Item": {
            "/Plone/news-item": 13
        },
        "Page": {
            "/Plone/page": 11
        },
        "Site": {
            "/Plone": 3
        }
    }
}
```

Note: the structure is not final, also I'm not sure if those are all information I want to display on the controlpanel.

jnptk:
is continued in https://github.com/plone/plone.volto/pull/197, closing.

PR #1796, opened 2024-07-09
fix: get schemata defined in the context

mamico:


plone-6.0.x.cfg
```diff
@@ -20,3 +20,4 @@ robotframework-assertion-engine = 2.0.0
 robotframework-debuglibrary = 2.3.0
 robotframework-pythonlibcore = 4.2.0
 grpcio-tools = 1.59.0
+docutils = 0.21.2
```
stevepiercy, line 23: See https://github.com/pypa/readme_renderer/blob/main/CHANGES.rst

```suggestion
readme_renderer <= 43.0
```

stevepiercy (changes requested):

plone-6.0.x.cfg
```diff
@@ -20,3 +20,4 @@ robotframework-assertion-engine = 2.0.0
 robotframework-debuglibrary = 2.3.0
 robotframework-pythonlibcore = 4.2.0
 grpcio-tools = 1.59.0
+readme_renderer = 43.0
```
stevepiercy, line 23: Meh. Still fails. I found that it is a dependency of [`zest.releaser`](https://zestreleaser.readthedocs.io/en/latest/changelog.html#a2-2023-07-19). See https://github.com/zestsoftware/zest.releaser/blob/master/pyproject.toml#L18. You could pin `zest.releaser` to `< 9.0.0` or `= 8.0.0` and see if that avoids the issue, although it won't fix it. At some point, it will need to be addressed.

stevepiercy (changes requested):
Fingers crossed!

plone-6.0.x.cfg
mamico: solved with last main branch merge

mamico:

mamico:

news/1796.bugfix
```diff
@@ -0,0 +1,2 @@
+Fix getting schemata for behaviors defined in the context
```
stevepiercy, line 1:
```suggestion
Get schemata for behaviors defined in the context. @mamico
```

news/1796.bugfix
```diff
@@ -0,0 +1,2 @@
...
+[mamico]
```
stevepiercy, line 2:
```suggestion
```

stevepiercy (changes requested):
Fixes for news items only. This PR needs a review from a maintainer.

stevepiercy (approved):
LGTM, but this still needs a review from a maintainer. Thank you!

src/plone/restapi/types/utils.py
```diff
@@ -229,7 +229,10 @@ def get_jsonschema_for_fti(fti, context, request, excluded_fields=None):
         fieldsets = ()
         additional_schemata = ()
     else:
-        additional_schemata = tuple(getAdditionalSchemata(portal_type=fti.id))
+        if fti.id == context.portal_type:
+            additional_schemata = tuple(getAdditionalSchemata(context=context))
+        else:
+            additional_schemata = tuple(getAdditionalSchemata(portal_type=fti.id))
```
davisagli, line 235: @mamico Can you help me understand the logic here? What is the scenario where we can't or shouldn't pass the context?

mamico: @davisagli thanks for asking for clarification and putting me de doubts...

The api is called to know the schema at the creation of a new CT or to know the schema at the modification of a content. In the first case I expected it to always be called from the root of the portal (but I'm wrong), in the second case it is called from the context of the object, and it would be useful to request the schema by passing the context.

So currently it is called (ie by Volto) as:
`/++api++/@types/Document` creation in the root
`/++api++/folder/@types/Document` creation in a folder
`/++api++/folder/doc1/@types/Document` modification of `/folder/doc1`

... but also ...

`/++api++/folder/doc1/@types/Document` to the creation of a new document inside `doc1` 

... here the modification I proposed behaves incorrectly (`doc1.portal_type` is equal to `Document` but the correct schema is the generic schema, not the contextual).

I am putting the PR back in draft because it needs to be rethought in some way.

Suggestions?

davisagli: @mamico It's important to get the schema for new content items in the context of its container, because some vocabularies might be context-dependent. (i.e. a list of keywords that depends on what section you are in)

So maybe we need a parameter so that the schema can be requested with or without context-dependent behaviors. Then they could be included by default, but the add form could request it without this.

mamico: @davisagli Your position is clear, and I absolutely agree with you. Do you have any suggestions for the parameter name to use? Something like `.../@types/Document?contextual=1` ?

PR #1798, opened 2024-07-27
Clean up contributing docs and use RTD for pull request previews

stevepiercy:
This PR cleans up contributing documentation for Plone 6 and switches from Netlify to Read the Docs for pull request previews.

Here's the changes to preview on a separate branch:

https://plonerestapi.readthedocs.io/en/docs-contributing/contributing/index.html

[Pull request previews](https://app.readthedocs.org/projects/plonerestapi/builds/?version__slug=&state=&version__type=external) won't build until merged to `main`.


stevepiercy:

stevepiercy:
I'm stuck. [docutils dropped support for Python 3.8](https://www.docutils.org/RELEASE-NOTES.html#release-0-21-2024-04-09). Zope pins it at 0.18.1, but we need 0.21.2 to use modern Sphinx and themes. How should I handle this?

I could create a new `plone-6.0.x-python3.8.cfg` file without the pin, and put a conditional in `tests.yml` for the `run` command.

```yml
run: if [ "${{ matrix.plone-version }}" == "6.0" ] && [ ${{ matrix.python-version }} == '3.8' ]; then buildout -t 10 -c plone-6.0.x-python3.8.cfg; else buildout -t 10 -c plone-${{ matrix.plone-version }}.x.cfg; fi;
```

I'll try that.

stevepiercy:

.readthedocs.yaml
```diff
@@ -5,25 +5,21 @@
...
+    python: "3.12"
+  commands:
+    # Cancel building pull requests when there aren't changes in the docs directory or YAML file.
+    # You can add any other files or directories that you'd like here as well,
+    # like your docs requirements file, or other files that will change your docs build.
+    #
+    # If there are no changes (git diff exits with 0) we force the command to return with 183.
+    # This is a special exit code on Read the Docs that will cancel the build immediately.
+    - |
+      if [ "$READTHEDOCS_VERSION_TYPE" = "external" ] && git diff --quiet origin/6.0 -- docs/ .readthedocs.yaml requirements-docs.txt requirements.txt;
```
mauritsvanrees, line 21: This should not compare with `origin/6.0` but with `origin/main` I would say.

And isn't the logic wrong?  Now if `git diff --quiet` exits with zero, the condition will be false, so `exit 183` is not reached.

mauritsvanrees (commented):
Two inline comments, for the rest this seems fine.  Thanks!

.readthedocs.yaml
stevepiercy: Fixed in 5f2017aa4502cfc242b8f7e7ad96f52f7c2856cf.

The logic is correct, yet tricky in bash. If there are no changes in the specified files, then `git diff` exits with `0`, which is "success" in bash and interpreted as `true`. Non-zero is interpreted as `false`. Also RTD told me to do it, and it works in Volto and Documentation. 😁

plone-6.0.x-python3.8.cfg
```diff
@@ -0,0 +1,22 @@
+[buildout]
+extends =
+    https://dist.plone.org/release/6.0.11.1/versions.cfg
+    base.cfg
+
+[buildout:python37]
```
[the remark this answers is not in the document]
stevepiercy: Fixed in d6bd7804ffacf07ce18647b84e9f07e9e7b0124b.

stevepiercy:

.readthedocs.yaml
mauritsvanrees: Ouch, so the GitHub conditions syntax is confusing, when you are used to how bash (or any shell) does it...
Thanks, the change is good.

mauritsvanrees (approved):

stevepiercy:
Thanks for the review! Merging!

davisagli:
I'm not happy with overriding a Zope pin here. Like I said in https://github.com/zopefoundation/Zope/issues/1220#issuecomment-2254348296, I think a better solution is to pin readme-renderer to a version that's compatible with the version of Sphinx that we use.

stevepiercy:
@davisagli there's a ton of changes coming through with Sphinx 8, MyST Parser 4.0, plone-sphinx-theme, and other packages, all of which depend on docutils > 21. Agreed, this is not ideal, but I need to start moving forward and not wait around.

Zope pins docutils for the sole reason to use RTD Sphinx Theme. Personally I believe that is not a good reason to accept its choice. Other Sphinx themes have leapt far beyond RTD Sphinx Theme. I'll ask if they want to move to a modern Sphinx theme.

davisagli:
@stevepiercy Sure, I support moving forward with a newer version of docutils. But we have to do it in the right order and test it first with plone.restapi's dependencies. If changing this pin here breaks something in another Plone package, we would have no way of knowing. In Plone docutils is not used only for documentation, but also for transforms between rst and html within Plone.

The main thing I'm saying is that we should update the pin first in buildout.coredev. That will allow us to see if it breaks tests in any Plone package. If not, then we can continue like this.

stevepiercy:
> In Plone docutils is not used only for documentation, but also for transforms between rst and html within Plone.
> 
> The main thing I'm saying is that we should update the pin first in buildout.coredev.

TIL.

I'll create a PR for CI tests, and post back here.

stevepiercy:
Well, I'm stuck. Neither Plone 6 or 5 docs say how to pin a package in `buildout.coredev`. I made some incorrect guesses. Which of the various files should I edit?

Oh, and should I run either `./bin/buildout` or `./bootstrap.sh` after a change, correct?

davisagli:
The pins are in versions.cfg. You don’t have to do anything locally, you
can create a PR with the update via GitHub and let CI run…

stevepiercy:
Done https://github.com/plone/buildout.coredev/pull/942. Thanks for the assist @davisagli. So far things are passing, and I just started Jenkins for the two builds on 6.1.

Should I do another PR against 6.0?

stevepiercy:
Opened similar PR, but had to use docutils 0.20.1 to get Python 3.8 support. 😞

mauritsvanrees:
If it helps anywhere, I could update the docutils pin in https://dist.plone.org/release/6.1-dev/ and https://dist.plone.org/release/6.0-dev/
Let me know if you see CI tests failing in Plone packages that would be fixed by such a change.

stevepiercy:
@mauritsvanrees can you explain how it might help? I sincerely don't know the details. My guess is that it would give developers an opportunity to test a build using a `*-dev` release before trying a full 6.x release.

I want to do the wisest thing, including reverting pins and putting them in the right places. What do you recommend?

We now have three pins referenced in this PR. Should I remove the one that I did in this PR, now that the other two were merged?

Please let me know. Thank you!

stevepiercy:
@mauritsvanrees @davisagli see https://github.com/zopefoundation/Zope/pull/1221/files where they switched to the Furo theme and use pinned versions of docutils according to the versions of Python they support.

Is there anything I should undo from the three PRs?

- https://github.com/plone/plone.restapi/pull/1798
- https://github.com/plone/buildout.coredev/pull/942
- https://github.com/plone/buildout.coredev/pull/943

mauritsvanrees:
The idea would be that if there is another Plone package that would benefit from the new docutils version for its docs, then it may help if that new docutils version is on `dist.plone.org/release/{6.0,6.1}-dev`.  `plone.restapi` here is extending the 6.0.11.1 versions/constraints, so it does not benefit.

Anyway, I have updated those two dev release directories with new docutils and two other packages.

Nothing to undo for now.  At one point when Plone coredev starts using a new Zope release with the updated version, then we can remove the overrides.

stevepiercy:
> At one point when Plone coredev starts using a new Zope release with the updated version, then we can remove the overrides.

Should we create an issue in the appropriate repo, whichever that is, to remind us?

mauritsvanrees:
No, once there is a new version of Zope released, I will adjust the coredev-buildouts and then I always look if we have set any overrides.  You correctly added the pin under the `OVERRIDES` headers, so that will work fine.

PR #1799, opened 2024-08-12
Include next/previous only if enabled

### LINKED ISSUE #1797: No way to determine whether next/ previous should be used from a content response
If `Enable next previous navigation` is unchecked or the `Next previous navigation` is disabled, the `next` and `previous` keys are still included in the response. This means that currently the only way to know whether the `Next/ Previous` should be used is to get the schema information for the content type and lookup the information there, but this requires a logged-in user.

## Proposed solutions

- A `Next/ previous enabled` key is added  to the content response which indicated whether the next/ previous should be used or not.

## Other options

- The 'Next' and 'Previous' keys are not included if the behaviours aren't enabled or the toggle behaviour is on but the toggle is unselected.
  - Unfavourable due to this being a breaking change in the response

JeffersonBledsoe:
Fixes #1797

### DIFF
--- a/news/1799.bugfix
+++ b/news/1799.bugfix
@@ -1,0 +1,4 @@
+Fixed Dexterity content serializer:
+Return an empty object for `next_item` and `previous_item`
+unless the parent has next/previous support enabled.
+@JeffersonBledsoe, @davisagli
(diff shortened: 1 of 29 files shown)


davisagli:

PR #1807, opened 2024-09-04
In @vocabulary endpoint, call the vocabulary factory with the 'title'…

### LINKED ISSUE #1806: Vocabularies should be called with query, if it exists
https://github.com/plone/plone.restapi/blob/aec08ce8f9d787694142a76b16f3147fef5d628c/src/plone/restapi/services/vocabularies/get.py#L89

Calling vocabularies can take a query:

https://github.com/plone/plone.app.vocabularies/blob/e228d563f4c3fd8fd68ea25f0a86dc9562cfea7e/plone/app/vocabularies/principals.py#L215

Calling `getVocabulary` action in Volto allows a query, too:

https://github.com/plone/volto/blob/41da257edaba21dd479f385b9d53d688283d093e/packages/volto/src/actions/vocabularies/vocabularies.js#L25

And there's a widget that supports this type of searching with vocabularies, already:

https://github.com/plone/volto/blob/41da257edaba21dd479f385b9d53d688283d093e/packages/volto/src/components/manage/Widgets/SelectAutoComplete.jsx#L192

tiberiuichim:
… field from the request form

Fixes #1806 

----
📚 Documentation preview 📚: https://plonerestapi--1807.org.readthedocs.build/

### DIFF
--- a/src/plone/restapi/services/vocabularies/get.py
+++ b/src/plone/restapi/services/vocabularies/get.py
@@ -86,7 +86,8 @@ class VocabulariesGet(Service):
                 404, "Not Found", f"The vocabulary '{vocabulary_name}' does not exist"
             )
 
-        vocabulary = factory(self.context)
+        query = self.request.form.get("title", "")
+        vocabulary = factory(self.context, query=query)
         serializer = getMultiAdapter(
             (vocabulary, self.request), interface=ISerializeToJson
         )
--- a/src/plone/restapi/services/vocabularies/get.py
+++ b/src/plone/restapi/services/vocabularies/get.py
@@ -86,8 +86,14 @@ class VocabulariesGet(Service):
                 404, "Not Found", f"The vocabulary '{vocabulary_name}' does not exist"
             )
 
+        # some vocabularies (such as plone.app.vocabularies.Principals)
+        # allow passing a query parameter to their factory
         query = self.request.form.get("title", "")
-        vocabulary = factory(self.context, query=query)
+        if query:
+            vocabulary = factory(self.context, query=query)
+        else:
+            vocabulary = factory(self.context)
+
         serializer = getMultiAdapter(
             (vocabulary, self.request), interface=ISerializeToJson
         )


src/plone/restapi/services/vocabularies/get.py
```diff
@@ -86,7 +86,8 @@ def reply(self):
                 404, "Not Found", f"The vocabulary '{vocabulary_name}' does not exist"
             )
 
-        vocabulary = factory(self.context)
+        query = self.request.form.get("title", "")
+        vocabulary = factory(self.context, query=query)
```
davisagli, line 90: @tiberiuichim This won't work with many vocabularies, which is why a lot of tests broke.

The IVocabularyFactory interface specifies that a vocab factory is called with a context, but not other parameters: https://github.com/zopefoundation/zope.schema/blob/master/src/zope/schema/interfaces.py#L1087

The addition of the `query` parameter seems to be something that was made up for the vocabulary factories in plone.app.vocabularies.principals without following any formal interface. This makes it hard to detect here how the factory is supposed to be called.

To really sort this out, we need to define a new interface like IQueryableVocabularyFactory. Then plone.app.vocabularies can define that these vocabulary factories implement that interface, and plone.restapi can be updated to pass the query to factories which provide that interface.

davisagli (changes requested):

tiberiuichim:
@davisagli what's your feeling on the low-tech approach, only pass the query parameter if it exists in the request?

With this change I can use the SelectAutocomplete widget to search and select users. I can understand your feeling about "use a custom endpoint", but it would be a lot more work than just these 3 changed lines, which are legitimate in filling a gap in Plone - Volto compatibility.

davisagli:
> what's your feeling on the low-tech approach, only pass the query parameter if it exists in the request?

Is there some mechanism to make sure that it is only sent in the request for the vocabularies that support it?

tiberiuichim:
No, the simple guard doesn't work, we have to implement your approach. I was hoping to limit the "damage" to plone.app.restapi, but it looks like it will need work in plone.app.vocabularies as well

davisagli:
@tiberiuichim My approach is probably possible to do in plone.restapi only, if it imports and applies a custom interface to these vocabularies at import time, and checks for it at runtime.

tiberiuichim:
I'll try a different approach. I'll implement my own custom vocabulary, based on the p.a.v.Principals, which would properly use the serializer API approach in plone.restapi. This would mean that this PR wouldn't be needed anymore

davisagli:
@tiberiuichim Okay, I'll close this for now

PR #1810, opened 2024-09-12
Safely convert items to IContentListingObject

### LINKED ISSUE #1809: Error when serializing a Collection that includes Events with recurrences
When you list events in a Collection, and try to serialize it through the REST API, you get the following error

```json
{
  "message": "('Could not adapt', <Occurrence at /Plone/eu/agenda/pentsio-duinak-eskatzeko-elkarretaratzea/2024-09-16>, <InterfaceClass plone.app.contentlisting.interfaces.IContentListingObject>)",
  "type": "TypeError"
}
```

The point is that plone.app.event's `Ocurrence` objects are not `IContentish` and thus are not adaptable to `IContentListingObject`, so [this line](https://github.com/plone/plone.restapi/blob/main/src/plone/restapi/serializer/summary.py#L86) breaks.

The solution on plone.restapi's side should be to defensively call `IContentListingObject` and in plone.app.event's side to define an adapter for `IContentListingObject`.

PRs to come

erral:
Fixes #1809 

----
📚 Documentation preview 📚: https://plonerestapi--1810.org.readthedocs.build/

### DIFF
--- a/src/plone/restapi/serializer/summary.py
+++ b/src/plone/restapi/serializer/summary.py
@@ -83,7 +83,10 @@ class DefaultJSONSummarySerializer:
         self.blocklisted_attributes = metadata["blocklisted_attributes"]
 
     def __call__(self):
-        obj = IContentListingObject(self.context)
+        try:
+            obj = IContentListingObject(self.context)
+        except TypeError:
+            return {}
 
         summary = {}
         for field in self.metadata_fields():
--- a/news/1810.bugfix
+++ b/news/1810.bugfix
@@ -1,0 +1,2 @@
+Safely convert objects to IContentListingObject
+[erral]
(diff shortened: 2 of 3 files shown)


erral:

src/plone/restapi/serializer/summary.py
```diff
@@ -83,7 +83,10 @@ def __init__(self, context, request):
         self.blocklisted_attributes = metadata["blocklisted_attributes"]
 
     def __call__(self):
-        obj = IContentListingObject(self.context)
+        try:
+            obj = IContentListingObject(self.context)
+        except TypeError:
+            return {}
```
davisagli, line 89: Doesn't this just make it less obvious what the problem is? As a consumer of the API, I would find it surprising to see an empty object in a list of items, and I would wonder what happened. Better to throw the error that happens now, I think.

davisagli (changes requested):

PR #1816, opened 2024-09-19
fix show_excluded_items in @navigation api

mamico:
The `show_ecluded_items` setting does not work for the `@navigation` API due to an incorrect condition.

----
📚 Documentation preview 📚: https://plonerestapi--1816.org.readthedocs.build/


mamico:
In Plone 5.2 test is broken with and without the fix.
I can't figure out if it's the wrong test or what else

mamico:

src/plone/restapi/tests/test_services_navigation.py
```diff
@@ -124,6 +124,35 @@ def test_dont_broke_with_contents_without_review_state(self):
         )
         self.assertIsNone(response.json()["items"][1]["items"][3]["review_state"])
 
+    def test_show_exclude_items(self):
```
stevepiercy, line 127: Typo.

```suggestion
    def test_show_excluded_items(self):
```

news/1816.bugfix
```diff
@@ -0,0 +1,2 @@
+Fix incorrect condition for show_ecluded_items setting in the @navigation API.
```
stevepiercy, line 1: Add reST syntax and fix typo.

```suggestion
Fix incorrect condition for ``show_excluded_items`` setting in the ``@navigation`` API.
```

src/plone/restapi/tests/test_services_navigation.py
```diff
@@ -124,6 +124,35 @@ def test_dont_broke_with_contents_without_review_state(self):
...
+            "Folder",
+            id="excluded-subfolder",
+            title="Excluded SubFolder",
+            exclude_from_nav=True,
+        )
+        transaction.commit()
+        response = self.api_session.get(
+            "/folder/@navigation", params={"expand.navigation.depth": 2}
+        )
+        self.assertNotIn(
```
stevepiercy, line 139: Is `settings.show_excluded_items==True` at this point? That could explain the test failure.

stevepiercy (changes requested):
Couple of typos and a question for the failing test.

src/plone/restapi/tests/test_services_navigation.py
stevepiercy: @mamico I noticed that there are differences between the the Python 3.8/Plone 5.2 and all other runs. There are two extra lines, excluding any deprecation warnings, between the test setup and test results: https://github.com/plone/plone.restapi/actions/runs/10954244520/job/30415848888?pr=1816#step:10:216

```console
*** Relations rebuild. flush: False
*** Relations rebuild. flush: True
```

Perhaps that is a clue? 

Also do we care about Plone 5.2/Python 3.8 for this bugfix? I don't think it is a security issue, but a maintenance fix, and Plone 5.2 is out of maintenance support.

mamico: @stevepiercy probably your first suggestion is the right one. It seems that Plone 5.2 and Plone 6.0 have different defaults for `show_ecluded_items` (?). Now I explicitly set it in the test. I am waiting for the results of the CI...

mamico:

stevepiercy (approved):
Wow, I actually guessed right! Yay, team!

We should have one more core developer review before merging.

davisagli (approved):
Thanks!

PR #1818, opened 2024-09-27
Fix invalid values in RelationListFieldSerializer.

### LINKED ISSUE #1745: relatedItems field returns None in the serializer if the related item is deleted
## Steps to reproduce

1. Create a content 1
2. Create a content 2
3. Edit content 1 and fill in the `relatedItems` field with content 2
4. Delete content 2
5. See content 1 serialization
6. `relatedItems` will be: `[ None ]`

This doesn't make sense to me. In the example, I think it's better to return an empty list. However, to do this, we would have to test the type of each element in a list [here](https://github.com/plone/plone.restapi/blob/f90329dfe49aa483517861a610b5175d68e0b6be/src/plone/restapi/serializer/converters.py#L100), to know if the value is a `RelationValue`. Or create a new adapter for the `IRelationList` field

Or should you handle this in the frontend?

Faakhir30:
Closes #1745 

Filtered `NoneType` items from the resulting list at `RelationListFieldSerializer`.

I guess we don't need to update this also in `RelationChoiceFieldSerializer`,  as it is returning a single `RelationValue` or `None` value, so it should be fine this way.

----
📚 Documentation preview 📚: https://plonerestapi--1818.org.readthedocs.build/

### DIFF
--- a/src/plone/restapi/serializer/relationfield.py
+++ b/src/plone/restapi/serializer/relationfield.py
@@ -7,7 +7,6 @@ from plone.restapi.serializer.dxfields import DefaultFieldSerializer
 from z3c.relationfield.interfaces import IRelationChoice
 from z3c.relationfield.interfaces import IRelationList
 from z3c.relationfield.interfaces import IRelationValue
-from z3c.relationfield import RelationValue
 from zope.component import adapter
 from zope.component import getMultiAdapter
 from zope.globalrequest import getRequest
@@ -48,6 +47,4 @@ class RelationListFieldSerializer(DefaultFieldSerializer):
         )
         if not value:
             return []
-        if isinstance(value, RelationValue):
-            return [value]
         return [el for el in value if el.to_object]
--- a/src/plone/restapi/serializer/relationfield.py
+++ b/src/plone/restapi/serializer/relationfield.py
@@ -42,9 +42,7 @@ class RelationListFieldSerializer(DefaultFieldSerializer):
         Returns:
             list: List of RelationValues
         """
-        value = getattr(
-            self.field.interface(self.context), self.field.__name__, default
-        )
+        value = super().get_value()
         if not value:
             return []
-        return [el for el in value if el.to_object]
+        return [el for el in value if el.to_id]
--- a/src/plone/restapi/serializer/relationfield.py
+++ b/src/plone/restapi/serializer/relationfield.py
@@ -7,6 +7,7 @@ from plone.restapi.serializer.dxfields import DefaultFieldSerializer
 from z3c.relationfield.interfaces import IRelationChoice
 from z3c.relationfield.interfaces import IRelationList
 from z3c.relationfield.interfaces import IRelationValue
+from z3c.relationfield import RelationValue
 from zope.component import adapter
 from zope.component import getMultiAdapter
 from zope.globalrequest import getRequest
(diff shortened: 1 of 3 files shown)


src/plone/restapi/serializer/relationfield.py
```diff
@@ -33,4 +34,20 @@ class RelationChoiceFieldSerializer(DefaultFieldSerializer):
...
+        Returns:
+            list: List of RelationValues
+        """
+        value = getattr(
+            self.field.interface(self.context), self.field.__name__, default
+        )
+        if not value:
+            return []
+        if isinstance(value, RelationValue):
+            return [value]
```
wesleybl, line 52: I think this case will not occur since it is a RelationList. I think it can be removed.

src/plone/restapi/serializer/relationfield.py
```diff
@@ -33,4 +34,20 @@ class RelationChoiceFieldSerializer(DefaultFieldSerializer):
...
+        return [el for el in value if el.to_object]
```
wesleybl, line 53: I think it's better to use `el.to_id` instead of `el.to_object`. `el.to_object` will wake up the objects.

wesleybl (changes requested):

src/plone/restapi/serializer/relationfield.py
Faakhir30: @wesleybl Ive tested this locally, turns otu el.to_id exists even if referenced to_id content has been deleted, so it ends up returning `[None]` as el.to_id couldnt filter it, so i guess awaking object is the only solution here.

Faakhir30:

src/plone/restapi/serializer/relationfield.py
wesleybl: > Ive tested this locally, turns otu el.to_id exists even if referenced to_id content has been deleted

It exists but it returns None for the deleted content. So the item will not be added to the list.

wesleybl: I saw that your test fails when `el.to_id` is used. This is because you missed calling `z3c.relationfield.event._setRelation`. For example:

```python
        rel1 = RelationValue(intids.getId(self.portal.doc1))
        rel2 = RelationValue(intids.getId(self.portal.doc2))
        self.portal.doc1.test_relationlist_field = [
            rel1,
            rel2,
        ]
        _setRelation(self.portal.doc1, "test_relationlist_field", rel1)
        _setRelation(self.portal.doc1, "test_relationlist_field", rel2)

```

Faakhir30:

wesleybl (approved):
LGTM

Faakhir30:

src/plone/restapi/serializer/relationfield.py
```diff
@@ -33,4 +33,18 @@ class RelationChoiceFieldSerializer(DefaultFieldSerializer):
...
+
+        Args:
+            default (list, optional): Default field value. Defaults to empty list.
+
+        Returns:
+            list: List of RelationValues
+        """
+        value = getattr(
+            self.field.interface(self.context), self.field.__name__, default
+        )
```
davisagli, line 47: I would use `super` to call the method inherited from DefaultFieldSerializer here, instead of copying its code: 

```suggestion
        value = super().get_value(default)
```

davisagli (changes requested):

davisagli:

davisagli (commented):
LGTM. We can merge it once the builds are green.

davisagli (approved):

Faakhir30:

PR #1813, opened 2024-09-17
added upload and download aliases in csv format.

Faakhir30:
Refers #1812 

Closes #1793 

Using content negotiation for PUT, GET aliases service.
 - send the "Accept" HTTP header "text/csv" for get request to download aliases as csv. Then REST API would return the CSV file.
 - send the "Content-Type" HTTP header "text/csv" for post request to add aliases from csv.

Tasks breakdown:

- [x] Handle `text/csv` content-type to upload aliases in bulk.
- [x] Handle `text/csv` content-type to download aliases.
- [x] Add new http-examples
- [x] Add aliases unit/integration tests to `plone.restapi`.
- [x] Add/update documentation

----
📚 Documentation preview 📚: https://plonerestapi--1813.org.readthedocs.build/

PR #1821, opened 2024-09-30
Fix logger initialization at add Allias by csv

Faakhir30:
Patch to #1813 

- Fixed logger's initialization.
- Add check to check content-type of uploading file is CSV.

----
📚 Documentation preview 📚: https://plonerestapi--1821.org.readthedocs.build/


src/plone/restapi/services/aliases/add.py
```diff
@@ -94,11 +94,19 @@ def _reply_csv(self):
         form = self.request.form
         if not form.get("file"):
             raise BadRequest("No file uploaded")
+
+        file = form["file"]
+
+        if not hasattr(file, "headers"):
+            raise BadRequest("Uploaded file is not a valid file")
```
davisagli, line 101: HTTPRequest only creates a FileUpload object if there are headers (see https://github.com/zopefoundation/Zope/blob/master/src/ZPublisher/HTTPRequest.py#L549) so I don't think this check is necessary.

news/1821.bugfix
```diff
@@ -0,0 +1 @@
+Add check on Content-Type of uploaded file at aliases bulk insertion by a CSV. @Faakhir30
```
stevepiercy, line 1:
```suggestion
Added a check for ``Content-Type`` of an uploaded CSV file for aliases bulk addition. @Faakhir30
```

stevepiercy (changes requested):

news/1821.bugfix
davisagli: @stevepiercy This is a tiny fix for a PR that was just merged and not released yet. I would prefer to remove this file and not add a new release note, since it's really part of the change that was already noted in https://github.com/plone/plone.restapi/pull/1813

I can bypass the towncrier check and merge once @Faakhir30 reviews my comment

davisagli (approved):

PR #1824, opened 2024-10-10
fix: @querystring shouldn't list userids

### LINKED ISSUE #1777: @querystring shouldn't list userids
# Problem

Restapi @querystring doesn’t require authentication.

This lists the users vocabulary which can list all users id and full name.  Depending how plone is configured, the email could be the id.

e.g.

``` sh
curl https://www.plone-demo.info/++api++/@querystring
```

``` json
      "sortable": true, 
      "title": "Creator", 
      "values": {
        "admin2": {
          "title": ""
        }, 
        "editor": {
          "title": ""
        }, 
        "reader": {
          "title": ""
        }, 
        "siteadmin": {
          "title": ""
        }
      }, 
      "vocabulary": “plone.app.vocabularies.Users"
```

The group names is also information that you might not want to make public.

Volto uses this for search blocks but I don’t think it needs to be public as it should only be needed for the search block editing UI. 
The editor could add a filter that uses that vocabulary but I don't think it makes sense to have a filter on users on a public site.

If you set “many users” then it works around the issue but this is not going to be obvious that this is a work

# Proposed solution
- don't allow users or groups to be filters and don't include those vocab in the querystring results

# Other solutions considered
- seperate out the vocab calls into different api calls so permissions can be applied per vocab?
- make @querystring an protected editor only call and use a different method to get the vocab for the filters.

askadityapandey:
This Pull Request fixes #1777 , lemme know if I need to make changes!

### DIFF
--- a/src/plone/restapi/services/querystring/configure.zcml
+++ b/src/plone/restapi/services/querystring/configure.zcml
@@ -4,48 +4,27 @@
     xmlns:plone="http://namespaces.plone.org/plone"
     xmlns:zcml="http://namespaces.zope.org/zcml"
     >
-  
-  <!-- Editor endpoint -->
+
   <plone:service
       method="GET"
-      factory=".get.QuerystringEditorGet"
-      for="Products.CMFPlone.interfaces.IPloneSiteRoot"
-      permission="cmf.ModifyPortalContent"
-      name="@querystring"
-      />
-  <plone:service
-      method="GET"
-      factory=".get.QuerystringEditorGet"
-      for="Products.CMFCore.interfaces.IContentish"
-      permission="cmf.ModifyPortalContent"
-      name="@querystring"
-      />
-      
-  <!-- Public endpoint -->
-  <plone:service
-      method="GET"
-      factory=".get.QuerystringPublicGet"
+      factory=".get.QuerystringGet"
       for="Products.CMFPlone.interfaces.IPloneSiteRoot"
       permission="zope2.View"
-      name="@querystring-public"
-      />
-  <plone:service
-      method="GET"
-      factory=".get.QuerystringPublicGet"
-      for="Products.CMFCore.interfaces.IContentish"
-      permission="zope2.View"
-      name="@querystring-public"
+      name="@querystring"
       />
 
-  <!-- Maintain caching for both endpoints -->
+  <plone:service
+      method="GET"
+      factory=".get.QuerystringGet"
+      for="Products.CMFCore.interfaces.IContentish"
+      permission="zope2.View"
+      name="@querystring"
+      />
+
   <cache:ruleset
-      for=".get.QuerystringEditorGet"
+      for=".get.QuerystringGet"
       ruleset="plone.content.dynamic"
       zcml:condition="have plone-app-caching-3"
       />
-  <cache:ruleset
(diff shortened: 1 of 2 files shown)


davisagli (changes requested):
Thanks for working on this. I'd like a few things to be different in the implementation:

1. We should not have 2 different endpoints. We should still have one endpoint, `@querystring`, but it should fiilter its output based on the current user's permissions.
2. Instead of a hardcoded list of sensitive vocabs, we should check for the user's permission to use the vocab in the same way that the `@vocabularies` endpoint already does: https://github.com/plone/plone.restapi/blob/f68fab7ea3209a096423d8567fa7f70a93786c7a/src/plone/restapi/services/vocabularies/get.py#L31
3. There should be at least one test for this filtering.
4. There should be a note about the permission-based filtering in the docs for this endpoint.

docs/source/endpoints/querystring.md
```diff
@@ -21,6 +21,8 @@ The `operators` property contains additional metadata about each operation.
 If an index uses a vocabulary, the vocabulary values are included in the `values` property.
 The vocabulary is resolved in the same context where the `/@querystring` endpoint is called (requires `plone.app.querystring >= 2.1.0`).
 
+**Note:** This endpoint returns the querystring configuration, filtering out vocabularies that the current user does not have permission to access. The filtering is done based on the same permission checks as the `@vocabularies` endpoint, ensuring that sensitive information, such as user or group data, is not exposed to unauthorized users.
```
davisagli, line 24: We can be a bit more concise:

```suggestion
If an index uses a vocabulary that the current user does not have permission to access, that index will not be included in the results.
```

src/plone/restapi/services/querystring/configure.zcml
```diff
@@ -4,27 +4,48 @@
...
-      for=".get.QuerystringGet"
+      for=".get.QuerystringEditorGet"
+      ruleset="plone.content.dynamic"
+      zcml:condition="have plone-app-caching-3"
+      />
+  <cache:ruleset
+      for=".get.QuerystringPublicGet"
       ruleset="plone.content.dynamic"
       zcml:condition="have plone-app-caching-3"
       />
```
davisagli, line 50: This configuration is currently broken, because the QuerystringPublicGet and QuerystringEditorGet classes no longer exist. That is why the tests failed.

I don't actually want any changes in this file. We're trying to update the existing endpoint to do some additional filtering, not change anything about how the endpoint is registered.

davisagli (changes requested):

davisagli:
@askadityapandey I would like to merge this, but I am waiting for you to remove the changes in configure.zcml. They are wrong and are the reason the tests are not passing.

davisagli:
@askadityapandey Please look at the automated checks that failed:
- "Black" -> this means you should autoformat the code by running `make black`
- "Tests" -> the tests are failing with a 500 error
- "Changelog verifier" -> you need to add an entry for the release notes, see https://6.docs.plone.org/contributing/index.html#contributing-change-log-label

davisagli:
Closing this because it had problems that were never corrected. I opened https://github.com/plone/plone.app.querystring/pull/172 to check the vocab permission as the querystring registry data is read, instead of filtering it afterward.

PR #1828, opened 2024-10-25
Handle parentheses search queries

tedw87:
This PR introduces a fix to handle parentheses  in the search functionality.
There is an error when query contains parantheses.
`ParseError
Token <Token:EOF> required, '(' found`

----
📚 Documentation preview 📚: https://plonerestapi--1828.org.readthedocs.build/

### DIFF
--- a/news/1828.bugfix
+++ b/news/1828.bugfix
@@ -1,1 +1,1 @@
-Fix error when search query contains parantheses. @tedw87
+Removed parentheses from search query. @tedw87
--- a/news/1828.bugfix
+++ b/news/1828.bugfix
@@ -1,0 +1,1 @@
+Fix error when search query contains parantheses. @tedw87
--- a/news/1828.bugfix
+++ b/news/1828.bugfix
@@ -1,1 +1,1 @@
-Removed parentheses from search query. @tedw87
+`@search` service: Remove parentheses from search query. @tedw87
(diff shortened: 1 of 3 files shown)


news/1828.bugfix
```diff
@@ -0,0 +1 @@
+Fix error when search query contains parantheses. @tedw87
```
stevepiercy, line 1:
```suggestion
Removed parentheses from search query. @tedw87
```

src/plone/restapi/search/handler.py
```diff
@@ -75,6 +75,10 @@ def _constrain_query_by_path(self, query):
             path = "/".join(self.context.getPhysicalPath())
             query["path"]["query"] = path
 
+    def quote_chars(self, query):
+        # Escape parentheses by adding backslashes before them
```
stevepiercy, line 79: This is not escaping, but removing.

```suggestion
        # Remove parentheses from the query
```

stevepiercy (changes requested):
Use correct terminology.

stevepiercy (approved):
LGTM, but this PR needs a technical review. Thank you!

davisagli (changes requested):
Please replace parentheses with a space rather than an empty string. a query like `foo(bar)` should be sanitized to `foo bar` to keep the terms separate, not `foobar`.

I would also like to have a test for this scenario.

news/1828.bugfix
```diff
@@ -0,0 +1 @@
+Removed parentheses from search query. @tedw87
```
davisagli, line 1:
```suggestion
`@search` service: Remove parentheses from search query. @tedw87
```

davisagli (approved):

davisagli:

PR #1836, opened 2024-11-04
Change login name after changing email in sites where "use email as login" is enabled

### LINKED ISSUE #1835: Email change does not change login name
When a site has "enable self registration" and "use email as login" options set to true, when a user changes their email the login name is not updated, and thus they can't log in with the new e-mail, and they have to log in with the old one.

erral:
Fixes #1835 

----
📚 Documentation preview 📚: https://plonerestapi--1836.org.readthedocs.build/

### DIFF
--- a/news/1835.bugfix
+++ b/news/1835.bugfix
@@ -1,0 +1,2 @@
+Fix log in after changing email when "email as login" is enabled
+[erral]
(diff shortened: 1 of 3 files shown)


jensens (approved):

jensens:

davisagli (commented):
@erral @jensens Thanks.

It looks like PAS can raise a ValueError if there was a problem changing the login, such as no plugin for that or a duplicate with an existing login. We should make sure this returns an error message to the client in a useful way.

PR #1845, opened 2024-11-20
Do not change request during serialization of relation fields

cekk:
Serializer should not change current request.
I cloned it to allow adding UID to the data and not touching original request.

----
📚 Documentation preview 📚: https://plonerestapi--1845.org.readthedocs.build/

### DIFF
--- a/news/1845.bugfix
+++ b/news/1845.bugfix
@@ -1,0 +1,2 @@
+Do not change request during relation fields serialization
+[cekk]


cekk:

src/plone/restapi/tests/test_dxfield_serializer.py
```diff
@@ -325,6 +325,19 @@ def test_relationlist_field_serialization_returns_list(self):
...
+        doc2 = self.portal[
+            self.portal.invokeFactory(
+                "DXTestDocument",
+                id="doc2",
+                title="Referenceable Document",
+                description="Description 2",
+            )
+        ]
+        self.serialize("test_relationchoice_field", doc2)
+        self.assertEqual(self.request.form["metadata_fields"], ["foo", "bar"])
```
davisagli, line 339: It'd be slightly preferable to use real metadata names and check that they are present in the serialized object (in other words, test the result, not the implementation)

davisagli (approved):
Thanks for the fix.

It would be nice if the `__call__` method of the ISerializeToJsonSummary adapter took an _optional_ parameter to override the metadata_fields, so that it would not need to be passed via the request. But, this is not easily backwards compatible. So I'll accept your fix.

PR #1847, opened 2024-11-26
Register IObjectPrimaryFieldTarget for Link objects

cekk:
This could be useful for example when you link a Link in slate.
Anonymous users automatically have resolved url in text link without redirects.

I choose to do it for anonymous users because otherwise this could change stored value in the slate data: for example if an editor of the current page can't edit the linked Link, he could override slate link data with the resolved url and not with the resolveuid link to Link object.

btw for this kind of use-cases probably could solve our problems a flag "is in edit-mode" or "is in view mode" that Volto sends to the backend and let us do different things in serializer based on that flag.

----
📚 Documentation preview 📚: https://plonerestapi--1847.org.readthedocs.build/


cekk:

davisagli:
> This could be useful for example when you link a Link in slate.

@cekk Can you remind me where/how does that end up using the IObjectPrimaryFieldTarget adapter?

cekk:
https://github.com/plone/plone.restapi/blob/main/src/plone/restapi/serializer/dxcontent.py#L137
and 
https://github.com/plone/plone.restapi/blob/main/src/plone/restapi/serializer/utils.py#L34

The first one add an attribute to the serialized json.
The second one is called evertime a block need to convert an internal link with resolveuid into a resolved url.

In my usecase when you refers to a Link object and call resolve_uid with its UID, i'd like to return:
- the Link absolute_url if the use is authenticated, because can potentially edit it
- the remoteUrl value of the Link object if the user is anonymous, to avoid an intermediate redirect (when the anon goes to the Link path and then will be redirected to its remoteUrl address)

davisagli:
@cekk Thanks. I agree with the use case; it matches how we handle Links in other situations. I just needed the reminder that the resolveuid code uses this.

davisagli (approved):

news/1847.feat
davisagli, line 1: .feat is not one of the recognized file extensions. This should be .feature

davisagli (changes requested):

PR #1848, opened 2024-11-26
Handle aliases in path2uid

cekk:
It's me..again ;)

This time, my edge-case is:

I have a parent Document (/aaa) with some contents inside it (/aaa/foo).
The parent have some text with an internal link to /aaa/foo that is stored in block data as a resolveuid string (correct).
Now, edit parent changing its id (from "aaa" to "bbb") and add some random text.
When you save the form, Volto sends to restapi the blocks with the link to foo like this: http://localhost:3000/aaa/foo.
restapi deserializer now try to resolve all links and convert them in resolveuid-like, but it is called after parent rename, so the path "/aaa/foo" is invalid and lead to store in the parent blocks a link like "http://localhost:3000/aaa/foo" that is potentially broken.

My pr add an additional check in path2uid method, to check also aliases for a given path.

----
📚 Documentation preview 📚: https://plonerestapi--1848.org.readthedocs.build/

### DIFF
--- a/news/1848.bugfix
+++ b/news/1848.bugfix
@@ -1,0 +1,1 @@
+Fix resolving paths in deserializer if the target was moved in the same request. @cekk
--- a/news/1848.feature
+++ b/news/1848.feature
@@ -1,0 +1,1 @@
+Handle aliases in path2uid to resolve also paths of renamed contents. @cekk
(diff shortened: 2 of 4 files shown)


cekk:

davisagli (approved):
@cekk That's quite a nasty edge case, and a clever solution. Thanks.

PR #1849, opened 2024-12-03
Do not break slate linkintegrity if block is not a slate block

cekk:
These are some compatibility fixes when blocks are not proper slate blocks.

We registered some linkintegrity adapters that extends slate one, but in some cases (we have some old installations that does not have slate yet) some methods breaks.

This pr tries to graceful handle them.

----
📚 Documentation preview 📚: https://plonerestapi--1849.org.readthedocs.build/

### DIFF
--- a/news/1849.bugfix
+++ b/news/1849.bugfix
@@ -1,0 +1,1 @@
+Do not break slate linkintegrity if block is not a slate block. @cekk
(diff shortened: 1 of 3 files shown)


cekk:

news/1849.bugfix
```diff
@@ -0,0 +1 @@
+Do not break slate linkintegrity if block is not a slate block. @cekk
```
davisagli, line 1:
```suggestion
Make slate block linkintegrity checking more robust in case data isn't in the expected format. @cekk
```

davisagli (approved):

PR #1851, opened 2024-12-12
Make primary field target adapter return early.

mauritsvanrees:
- When there is no primary field, do not go through all fields.
- First compare the field name, then the permission: comparing two strings is cheaper.
- If the field name matches, stop iterating over other fields: you have already found the field.

----
📚 Documentation preview 📚: https://plonerestapi--1851.org.readthedocs.build/

### DIFF
--- a/news/1851.bugfix
+++ b/news/1851.bugfix
@@ -1,0 +1,2 @@
+Make primary field target adapter return early.
+[maurits]
--- a/news/1851.bugfix
+++ b/news/1851.bugfix
@@ -1,2 +1,1 @@
-Make primary field target adapter return early.
-[maurits]
+Optimized performance of DexterityObjectPrimaryFieldTarget adapter. @maurits
--- a/src/plone/restapi/serializer/dxcontent.py
+++ b/src/plone/restapi/serializer/dxcontent.py
@@ -228,7 +228,6 @@ class DexterityObjectPrimaryFieldTarget:
             field = getFields(schema).get(primary_field_name)
             if field is None:
                 continue
-            print(primary_field_name)
             if not self.check_permission(
                 read_permissions.get(primary_field_name),
                 self.context,
--- a/src/plone/restapi/serializer/dxcontent.py
+++ b/src/plone/restapi/serializer/dxcontent.py
@@ -225,19 +225,22 @@ class DexterityObjectPrimaryFieldTarget:
         for schema in iterSchemata(self.context):
             read_permissions = mergedTaggedValueDict(schema, READ_PERMISSIONS_KEY)
 
-            for name, field in getFields(schema).items():
-                if name != primary_field_name:
-                    continue
+            field = getFields(schema).get(primary_field_name)
+            if field is None:
+                continue
+            print(primary_field_name)
+            if not self.check_permission(
+                read_permissions.get(primary_field_name),
+                self.context,
+            ):
+                return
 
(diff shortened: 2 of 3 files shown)


src/plone/restapi/serializer/dxcontent.py
```diff
@@ -220,23 +220,24 @@ def __init__(self, context, request):
 
     def __call__(self):
         primary_field_name = self.get_primary_field_name()
+        if not primary_field_name:
+            return
         for schema in iterSchemata(self.context):
             read_permissions = mergedTaggedValueDict(schema, READ_PERMISSIONS_KEY)
 
             for name, field in getFields(schema).items():
```
ale-rt, line 228: Would:
```
field = getFields(schema).get(primary_field_name)
```
work, rather than looping on all the schema fields?

mauritsvanrees:

news/1851.bugfix
```diff
@@ -0,0 +1,2 @@
+Make primary field target adapter return early.
+[maurits]
```
davisagli, line 2:
```suggestion
Optimized performance of DexterityObjectPrimaryFieldTarget adapter. @maurits
```

davisagli (approved):
This looks okay as is. The suggestion from @ale-rt also makes sense. @mauritsvanrees I'll let you decide whether to adjust it.

src/plone/restapi/serializer/dxcontent.py
mauritsvanrees: I did not want to change too much, but testing it seems to work, so yes, let's go for this.  Change made.

mauritsvanrees:
I now have a totally unrelated error only on [Plone 6.0 Python 3.12](https://github.com/plone/plone.restapi/actions/runs/12362761511/job/34503168178#step:10:2321):

```
  File "/home/runner/work/plone.restapi/plone.restapi/eggs/Products.PlonePAS-8.0.4-py3.12.egg/Products/PlonePAS/config.py", line 1, in <module>
    from PIL import Image
  File "/home/runner/work/plone.restapi/plone.restapi/eggs/Pillow-9.5.0-py3.12-linux-x86_64.egg/PIL/Image.py", line 103, in <module>
    from . import _imaging as core
ImportError: libtiff.so.5: cannot open shared object file: No such file or directory
```

The same run last week was fine.  Difference: last week this used [Ubuntu 22.04](https://github.com/plone/plone.restapi/actions/runs/12308576300/job/34354182307#step:1:4).  Now it uses [24.04](https://github.com/plone/plone.restapi/actions/runs/12362761511/job/34503168178#step:1:4).

In the upcoming Plone 6.0.14 I am upgrading Pillow to 11.0.0.  That could magically solve it. :-/
Or should I change the workflow to use ubuntu 22 explicitly?  That may be best for now.  I will do so.

mauritsvanrees:

src/plone/restapi/serializer/dxcontent.py
```diff
@@ -220,23 +220,27 @@ def __init__(self, context, request):
...
-                    (field, self.context, self.request), IPrimaryFieldTarget
-                )
-                if target_adapter:
-                    target = target_adapter()
-                    if target:
-                        return target
+            field = getFields(schema).get(primary_field_name)
+            if field is None:
+                continue
+            print(primary_field_name)
```
davisagli, line 231:
```suggestion
```

davisagli (approved):

PR #1855, opened 2025-01-13
Handle bad input as BadRequest instead of ValueError

djay:
useful so it can be more easily ignored

----
📚 Documentation preview 📚: https://plonerestapi--1855.org.readthedocs.build/

### DIFF
--- a/news/1855.bugfix
+++ b/news/1855.bugfix
@@ -1,0 +1,1 @@
+Changed bad int inputs into 500 Exceptions to 400 BadRequest so they can be filtered out of logs more easily. - djay
--- a/news/1855.bugfix
+++ b/news/1855.bugfix
@@ -1,1 +1,1 @@
-Changed bad int inputs into 500 Exceptions to 400 BadRequest so they can be filtered out of logs more easily. - djay
+Changed bad int inputs into 500 Exceptions to 400 BadRequest so they can be filtered out of logs more easily. @djay
--- a/src/plone/restapi/services/navigation/get.py
+++ b/src/plone/restapi/services/navigation/get.py
@@ -17,6 +17,7 @@ from zope.component.hooks import getSite
 from zope.i18n import translate
 from zope.interface import implementer
 from zope.interface import Interface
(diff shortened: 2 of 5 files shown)


djay:

src/plone/restapi/services/navigation/get.py
```diff
@@ -29,7 +30,10 @@ def __init__(self, context, request):
 
     def __call__(self, expand=False):
         if self.request.form.get("expand.navigation.depth", False):
-            self.depth = int(self.request.form["expand.navigation.depth"])
+            try:
+                self.depth = int(self.request.form["expand.navigation.depth"])
+            except ValueError as e:
```
davisagli, line 35:
```suggestion
            except (ValueError, TypeError) as e:
```

(Unlikely, but could be TypeError if the querystring had something like `expand.navigation.depth:list=asdf`)

src/plone/restapi/services/navigation/get.py
```diff
@@ -29,7 +30,10 @@ def __init__(self, context, request):
...
+                raise BadRequest(e)
```
davisagli, line 36: Could you write a test for this case please? The BadRequest gets serialized by plone.rest and I can't remember if you're allowed to pass another exception instance instead of a string.

davisagli (changes requested):

src/plone/restapi/services/navigation/get.py
djay: done

jensens (approved):

news/1855.bugfix
```diff
@@ -0,0 +1 @@
+Changed bad int inputs into 500 Exceptions to 400 BadRequest so they can be filtered out of logs more easily. - djay
```
davisagli, line 1:
```suggestion
Changed bad int inputs into 500 Exceptions to 400 BadRequest so they can be filtered out of logs more easily. @djay
```

davisagli (approved):
@djay Thanks!

PR #1857, opened 2025-01-14
Handle more BadRequests

djay:
----
📚 Documentation preview 📚: https://plonerestapi--1857.org.readthedocs.build/

### DIFF
--- a/news/1857.bugfix
+++ b/news/1857.bugfix
@@ -1,1 +1,1 @@
-Handle ValueError on querystringsearch as BadRequest - djay
+Handle TypeError on querystringsearch as BadRequest. @djay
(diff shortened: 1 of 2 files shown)


davisagli (commented):
@djay Maybe we should add a parse_int helper so we can avoid repeating the details?

djay:
@davisagli can but this was the last one I found turning up in my logs. But if you think it's worth it...

davisagli:
@djay That was me talking out loud while I reviewed. Not a blocker.

news/1857.bugfix
```diff
@@ -0,0 +1 @@
+Handle ValueError on querystringsearch as BadRequest - djay
```
davisagli, line 1:
```suggestion
Handle TypeError on querystringsearch as BadRequest. @djay
```

davisagli (approved):

PR #1858, opened 2025-01-15
Add parse_int to handle all cases of BadRequest from ints passed in

djay:
- **Another should be BadRequest found in the wild**
- **add changelog**
- **add parse_int util**
- **use parse_int for navigation also**
- **improve bad int message**

----
📚 Documentation preview 📚: https://plonerestapi--1858.org.readthedocs.build/

### DIFF
--- a/news/1858.bugfix
+++ b/news/1858.bugfix
@@ -1,1 +1,1 @@
-add parse_int to handle all cases of BadReqests from ints passed in. @djay
+Add parse_int to handle all cases of BadRequests from ints passed in. @djay
--- a/news/1858.bugfix
+++ b/news/1858.bugfix
@@ -1,0 +1,1 @@
+add parse_int to handle all cases of BadReqests from ints passed in. @djay
--- a/src/plone/restapi/deserializer/__init__.py
+++ b/src/plone/restapi/deserializer/__init__.py
@@ -44,4 +44,4 @@ def parse_int(data, prop, default):
     try:
         return int(data.get(prop, default))
     except (ValueError, TypeError):
-        raise BadRequest("Invalid %s" % prop)
+        raise BadRequest(f"Invalid {prop}: Not an integer")
--- a/src/plone/restapi/deserializer/__init__.py
+++ b/src/plone/restapi/deserializer/__init__.py
@@ -1,4 +1,5 @@
 from plone.restapi.exceptions import DeserializationError
+from zExceptions import BadRequest
 
 import json
 
@@ -28,3 +29,19 @@ def boolean_value(value):
 
     """
     return value not in {False, "false", "False", "0", 0}
+
+
+def parse_int(data, prop, default):
+    """
+    Args:
+        data: dict from a request
+        prop: name of a integer paramater in the dict
+        default: default if not found
+
+    Returns: an integer
+    Raises: BadRequest if not an int
+    """
+    try:
+        return int(data.get(prop, default))
+    except (ValueError, TypeError):
+        raise BadRequest("Invalid %s" % prop)
--- a/src/plone/restapi/services/discussion/conversation.py
+++ b/src/plone/restapi/services/discussion/conversation.py
@@ -2,7 +2,7 @@ from datetime import datetime
 from plone.app.discussion.browser.comment import EditCommentForm
 from plone.app.discussion.browser.comments import CommentForm
 from plone.app.discussion.interfaces import IConversation
-from plone.restapi.deserializer import json_body
+from plone.restapi.deserializer import json_body, parse_int
 from plone.restapi.interfaces import ISerializeToJson
 from plone.restapi.services import Service
 from plone.restapi.services.discussion.utils import can_delete
@@ -38,7 +38,7 @@ class CommentsGet(Service):
 
     def publishTraverse(self, request, name):
         if name:
-            self.comment_id = int(name)
+            self.comment_id = parse_int(dict(comment_id=name), "comment_id", None)
         return self
 
     def reply(self):
@@ -57,7 +57,7 @@ class CommentsAdd(Service):
 
     def publishTraverse(self, request, name):
         if name:
-            self.comment_id = int(name)
+            self.comment_id = parse_int(dict(comment_id=name), "comment_id", None)
             request["form.widgets.in_reply_to"] = name
         return self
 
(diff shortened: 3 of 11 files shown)


djay:
@davisagli sorry. I changed my mind. it does make the code cleaner.

djay:
@davisagli one other possible change is that I get a DesearlisationError from parse_body from plugins like collective.honeypot instead of BadRequest. I could change parse_body to raise a BadRequest instead and fix all such plugins... but it felt like changing the api... on the other hand, it's not like in that case they were properly using that exception.
or I could add a param to parse_body of "raise_badrequest=True"?

djay:
@davisagli or DeserealisationError could inherit from badrequest...

davisagli:
@djay I think it's fine the way you have it, catching DeserializationError and raising a BadRequest. It's conceivable that some endpoints might want to handle a DeserializationError by falling back to some other source of data instead of returning a 400.

news/1858.bugfix
```diff
@@ -0,0 +1 @@
+add parse_int to handle all cases of BadReqests from ints passed in. @djay
```
davisagli, line 1:
```suggestion
Add parse_int to handle all cases of BadRequests from ints passed in. @djay
```

src/plone/restapi/services/discussion/conversation.py
```diff
@@ -38,7 +38,7 @@ class CommentsGet(Service):
 
     def publishTraverse(self, request, name):
         if name:
-            self.comment_id = int(name)
+            self.comment_id = parse_int(dict(comment_id=name), "comment_id", None)
```
davisagli, line 41: This is okay, but a bit awkward. Maybe the first parameter of parse_int should be the actual value rather than a dict which contains the value? Not a deal breaker though.

davisagli (approved):

davisagli:

djay:
@davisagli I like the idea of changing the base exception. Because then I don't hanve to get another change into collective.honeypot and who know what other plugins have this problem.

```

class DeserializationError(BadRequest):
    """An error happened during deserialization of content."""

    def __init__(self, msg):
        self.msg = msg

    def __str__(self):
        return repr(self.msg)


class QueryParsingError(BadRequest):
    """An error happened while parsing a search query."""
```

davisagli:
@djay I'd rather not change the base exception. It's tying an exception that's about parsing too closely to one that's about HTTP responses, and they aren't always done together.

src/plone/restapi/services/discussion/conversation.py
djay: then you need to repeat the prop anyway to get a nicer error message in the exception...
or have parse_int and parse_int_dict.

davisagli: Yeah. It can stay like this.

djay:
@davisagli fair enough. from that perspective parse_body is linked to requests so this seems reasonable?

```
def json_body(request):
    # TODO We should not read the complete request BODY in memory.
    # Once we have fixed this, we can remove the temporary patches.py.
    # See there for background information.
    try:
        data = json.loads(request.get("BODY") or "{}")
    except ValueError:
        raise BadRequest("No JSON object could be decoded")
    if not isinstance(data, dict):
        raise DeserializationError("Malformed body")
    return data

```

davisagli:
@djay I don't understand yet why anything needs to change in json_body. You're already handling the DeserializationError it can raise...

djay:
@davisagli because then this is not needed. I've no idea how long this would take to get approved and other plugins might also have similar problems - https://github.com/collective/collective.honeypot/pull/26

davisagli:
@djay I can sort of see the argument for making json_body raise BadRequest which means it would return a 400 by default if the developer doesn't explicitly handle it. Let me sleep on it.

src/plone/restapi/services/discussion/conversation.py
djay: or I can do this. Just a little less explicit

```
def parse_int(data, prop, default):
    """
    Args:
        data: dict from a request to get str, or str itself
        prop: name of a integer paramater in the dict
        default: default if not found

    Returns: an integer
    Raises: BadRequest if not an int
    """
    value = data if type(data) in (str, int) else data.get(prop, default)
    try:
        return int(value)
    except (ValueError, TypeError):
        raise BadRequest(f"Invalid {prop}: Not an integer")
```

src/plone/restapi/deserializer/__init__.py
```diff
@@ -11,7 +11,8 @@ def json_body(request):
     try:
         data = json.loads(request.get("BODY") or "{}")
     except ValueError:
-        raise DeserializationError("No JSON object could be decoded")
+        # Often from spam/pentests which we want to ignore
+        raise BadRequest("No JSON object could be decoded")
```
davisagli, line 15: @djay I want to go back to DeserializationError here. Any code that calls `json_body` _should_ already be handling DeserializationError. I don't want to force existing code to be updated in cases where it actually wants to handle the error and not return a 400 response.

davisagli (changes requested):

src/plone/restapi/deserializer/__init__.py
djay: @davisagli turns out DeserializationError is a 400 response. But also turns out collective.sentry sends 400 responses.... 
Now I could change collective.sentry to ignore any 400 status but how then do I sift out in sentry the difference between an empty or rubbish request to any endpoint vs a bug? DeserializationError is used all over for all sorts of problems with request. It's for this reason why I'm still thinking it makes sense to differentiate between completely unparsable request vs another kind of bug. 
but I take your point other code might be handling this problem but if they are what else are they going to do but say it was a badrequest?
whats your thoughts?

davisagli: @djay I'm prioritizing backwards-compatibility here. Currently it's the responsibility of the code which calls json_body to handle any errors and return an appropriate response. It might be okay to change that but I'm not 100% confident. There is code in the restapi PAS plugin which catches DeserializationError and then falls back to authenticating from a different source. There are a lot of results on a github search for DeserializationError.

> Now I could change collective.sentry to ignore any 400 status but how then do I sift out in sentry the difference between an empty or rubbish request to any endpoint vs a bug?

You can't, with or without your suggested change here. A rubbish request could be from a bot or from a bug in a legitimate client. I think the server should return a 400 response either way, and it's not worth highlighting that in error monitoring. The legitimate client can report the 400 response in a way that makes it obvious something was wrong.

If this were new code I'd be okay with making json_body raise BadRequest in both cases, but it's existing code that's used in many places, and I'm not going to accept this change now.

djay:
@davisagli I removed those commits

davisagli (approved):
@djay Thank you!

davisagli:

Issue #1834, opened 2024-11-01
Drop support for Plone 5.2 and Python 3.8


ericof:
We should plan the next major release of plone.restapi (10.x) to drop support for Python 3.8, already end of life, and Plone 5.2 (will be end of life with the release of Plone 6.1.)

mauritsvanrees:
5.2 is end of life since end of October. I have updated the release schedule yesterday.

PR #1863, opened 2025-01-20
Prevent deprecation warnings.

mauritsvanrees:


davisagli (commented):
For what it's worth, I think it's time to work on a new major version of plone.restapi which drops support for Plone 5.2 (see also #1834)

davisagli (approved):
Merging as is, and I'll come back to dropping support for Plone 5.2 in a separate PR.

PR #1864, opened 2025-01-21
fix: contextnavigation when a file has an unknown type

mamico:
Fix the context navigation bug that occurs when a file type is unknown.

In my case, the file had a content type of `application/pkcs7` when I encountered the bug

----
📚 Documentation preview 📚: https://plonerestapi--1864.org.readthedocs.build/


mamico:

news/1864.bugfix
```diff
@@ -0,0 +1 @@
+Fix the context navigation bug that occurs when a file type is unknown. @mamico
```
davisagli, line 1:
```suggestion
In the `@contextnavigation` endpoint, return `"icon": null` for Files with a mimetype not found in the `content_type_registry`, instead of raising `TypeError`. @mamico
```

davisagli (approved):
Thanks!

PR #1865, opened 2025-01-22
Update Plone test versions, and pin twine

davisagli:
There was a new release of twine today which requires packaging 24.0, which is not in our KGS. /cc @mauritsvanrees 

----
📚 Documentation preview 📚: https://plonerestapi--1865.org.readthedocs.build/

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -10,8 +10,6 @@ jobs:
         include:
           - python-version: "3.8"
             plone-version: "5.2"
-          - python-version: "3.8"
-            plone-version: "6.0"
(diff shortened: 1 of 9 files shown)


mauritsvanrees:
> There was a new release of twine today which requires packaging 24.0, which is not in our KGS. /cc @mauritsvanrees

Plone 6.0.13 had `packaging` 23.2, but in Plone 6.0.14 and 6.1.0b2 we pin `packaging` 24.2, so that should be fine.

Also, in 6.0 the twine pin is `versions-extra.cfg`, so we could also include that one.

PR #1875, opened 2025-01-31
Handle block deserializers in control panels 

JeffersonBledsoe:


davisagli (commented):
@JeffersonBledsoe It looks plausible to me, but I'd really like to see a test confirming that the resolveuid transforms are applied. Also don't we probably need something similar for the serializer?

JeffersonBledsoe:
> @JeffersonBledsoe It looks plausible to me, but I'd really like to see a test confirming that the resolveuid transforms are applied. Also don't we probably need something similar for the serializer?

Yep, I'll get round to those soon hopefully! The current status is what was needed for a project, I'll add the tests and the serializer part of the story soon

PR #1880, opened 2025-02-12
Do not hardcode show_inactive in search

### LINKED ISSUE #1879: search should not hardcode default behavior
In testing https://github.com/plone/Products.CMFPlone/pull/4099 I find, that the plone.restapi search implementation hardcodes the search parameter `show_inactive`. 

https://github.com/plone/plone.restapi/blob/bc29368009dfc0719adab723cfddd9675a3c79ea/src/plone/restapi/search/handler.py#L124

which is already the default:

https://github.com/plone/Products.CMFPlone/blob/4a765ec624f5b93e6388c2057a8caced954ae958/Products/CMFPlone/CatalogTool.py#L406

But then plone.restapi tests rely on the buggy behavior described in https://github.com/plone/Products.CMFPlone/issues/4098 and actually expect inactive documents to be returned in the search, for example here:

https://github.com/plone/plone.restapi/blob/bc29368009dfc0719adab723cfddd9675a3c79ea/src/plone/restapi/tests/test_search.py#L746

The fix is to remove the hardcoded search parameter and rely on CMFPlone doing the right thing, which is both backward compatible with current CMFPlone master and forward compatible with https://github.com/plone/Products.CMFPlone/pull/4099.

Fix forthcoming.

gyst:
Fixes #1879.

This fix is both backward compatible, and forward compatible with https://github.com/plone/Products.CMFPlone/pull/4099

See also https://github.com/plone/Products.CMFPlone/issues/4098

### DIFF
--- a/news/1879.bugfix
+++ b/news/1879.bugfix
@@ -1,1 +1,1 @@
-Do not hardcode show_inactive in search; let CMFPlone handle that.
+Do not hardcode ``show_inactive`` in search; let ``Products.CMFPlone`` handle that. @gyst
--- a/news/1879.bugfix
+++ b/news/1879.bugfix
@@ -1,0 +1,1 @@
+Do not hardcode show_inactive in search; let CMFPlone handle that.
--- a/src/plone/restapi/search/handler.py
+++ b/src/plone/restapi/search/handler.py
@@ -120,9 +120,6 @@ class SearchHandler:
             types = types["query"]
         query["portal_type"] = self.filter_types(types)
 
-        # respect effective/expiration date
-        query["show_inactive"] = False
-
         # respect navigation root
         if "path" not in query:
             query["path"] = {"query": get_navigation_root(self.context)}


gyst:

news/1879.bugfix
```diff
@@ -0,0 +1 @@
+Do not hardcode show_inactive in search; let CMFPlone handle that.
```
stevepiercy, line 1: Sprinkle some reStructuredText syntax for `CHANGES.rst`, and give yourself credit.

```suggestion
Do not hardcode ``show_inactive`` in search; let ``Products.CMFPlone`` handle that. @gyst
```

stevepiercy (changes requested):
Minor tweak on the change log.

stevepiercy (approved):
Change log looks good, but this still needs a technical review from someone who knows more than I do.

gyst:
Yeah I've asked David, who already reviewed and approved the main PR for this in Products.CMFPlone.

PR #1884, opened 2025-03-08
Added `is_multilingual` to @site endpoint

sneridagh:
----
📚 Documentation preview 📚: https://plonerestapi--1884.org.readthedocs.build/


sneridagh:
@davisagli the builds are failing for 6.0, no clue why :( Same happens locally.

I still have to add the documentation updates. 
I guess you have it already in mind, but the auto documentation generator  is still pointing for using 6.0, and the `Makefile` command is missing for 6.1.

I added some comments, I had doubts, doing it inline.

src/plone/restapi/services/site/get.py
```diff
@@ -54,6 +59,11 @@ def __call__(self, expand=False):
             }
         )
 
+        if HAS_MULTILINGUAL and IPloneAppMultilingualInstalled.providedBy(self.request):
+            result["site"]["plone.is_multilingual"] = True
```
sneridagh, line 63: @davisagli should we use the prefix in this one? I assume we left the others because they are mapped like this to the registry values.

src/plone/restapi/services/site/get.py
```diff
@@ -54,6 +59,11 @@ def __call__(self, expand=False):
...
+        else:
+            result["site"]["plone.is_multilingual"] = False
+
```
sneridagh, line 66: Which was your intention for the `features` dict? Should we put this one on it?

davisagli: @sneridagh I would prefer to add it to the `features` dict.

I don't like that we used the `plone` prefix here; let's not make the problem worse by using it for things that don't even come from a registry setting with the same name.

And, this is exactly what the features dict is for: when the frontend needs to be able to check if some feature is provided by the backend.

davisagli (approved):

sneridagh:
Thanks @davisagli ! It would be great to have a release soon, so we can continue with Volto's related PR. I think @mauritsvanrees is thinking in cutting a Plone one soon.

PR #1885, opened 2025-03-10
Restore Vale and use latest version

stevepiercy:


stevepiercy:

styles/config/vocabularies/Base/accept.txt
davisagli, line 1: Can we put these inside the docs folder instead of adding a new top-level folder? It's not at all obvious that "styles" is related to Vale.

davisagli (changes requested):
How is this meant to be used? I would expect either it should be called from some CI job (if it's supposed to be automatic) or there should be a mention of it in the README (if it's supposed to be done manually)

styles/config/vocabularies/Base/accept.txt
stevepiercy: Done https://github.com/plone/plone.restapi/pull/1885/commits/4dedaed70c2f7bb3186ff219bb91412366bf88e5

stevepiercy:

PR #1886, opened 2025-03-12
Apply the resolveuid transform to all JSON fields, not only blocks

### LINKED ISSUE #1842: JSON fields in content types - Improve the overall story
It is quite common to use in Volto content type fields that are stored in JSON using `JSONField` field type. This is useful to store structured data (as in the old Datagridfield). We would like to be able to apply the some of the transforms in that data structures, and being able to process resolveIds and image scales, etc in there too, like in the blocks.

Also, being able to apply them in a "smart" way, so that transforms apply to some fields that are marked as so. For example, we have a small use case working in the blocks transforms already:

```json
{
  "blocks": {
    "asdadasd": {
      "@type": "URL",
       "value": "https://myserver/foo/bar"
    }
  }
}
```

but it would be great to expand this to other use cases, like images or for text indexing.

See also: https://github.com/plone/volto/issues/2252

davisagli:
Fixes #1842 

This makes it possible to use a non-blocks JSONField that stores internal links as resolveuid URLs.

----
📚 Documentation preview 📚: https://plonerestapi--1886.org.readthedocs.build/

### DIFF
--- a/news/1886.feature
+++ b/news/1886.feature
@@ -1,1 +1,2 @@
-Convert internal URLs to resolveuid URLs in JSON fields. @davisagli
+Apply block serialization and deserialization transforms also to JSON fields.
+This includes converting internal URLs to resolveuid URLs. @davisagli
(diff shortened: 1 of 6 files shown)


davisagli:

sneridagh (approved):
🚀 🕺 🎉

Issue #1888, opened 2025-03-13
Self registration broken


MrTango:
If i enable self registration in Plone and try to use the /register form to register a user, it only works when I'm already logged-in. For anonymous users, an error occurs:

```
https://tagung.plone.de/++api++/@users
[HTTP/3 400  271ms]

error    Object { message: "Error in fields. Property 'Eine Bestätigungs-E-Mail versenden mit einem Link, um das Passwort zu setzen.' is not allowed.", type: "WrongParameterError", errors: […] }
errors    [ {…} ]
0    Object { field: "sendPasswordReset", message: "Property 'Eine Bestätigungs-E-Mail versenden mit einem Link, um das Passwort zu setzen.' is not allowed." }
field    "sendPasswordReset"
message    "Property 'Eine Bestätigungs-E-Mail versenden mit einem Link, um das Passwort zu setzen.' is not allowed."
message    "Error in fields. Property 'Eine Bestätigungs-E-Mail versenden mit einem Link, um das Passwort zu setzen.' is not allowed."
type    "WrongParameterError"
```

in my opinion, we have to also allow sending `sendPasswordReset` when self registration is enabled:

```
        if security.enable_self_reg:
            allowed.append("sendPasswordReset")
```

MrTango:
Volto is currently always sending sendPasswordReset = true, this is why self registration is not working at all at the momen.

shivarawart:
I am a contributor

shivarawart:
I am preparing for proposal for specific project of phone.

PR #1889, opened 2025-03-13
Allow sending `sendPasswordReset` if self registration is enabled.

MrTango:
----
📚 Documentation preview 📚: https://plonerestapi--1889.org.readthedocs.build/

### DIFF
--- a/news/1888.bugfix
+++ b/news/1888.bugfix
@@ -1,0 +1,1 @@
+Allow sending `sendPasswordReset` if self registration is enabled. @MrTango
--- a/news/1888.bugfix
+++ b/news/1888.bugfix
@@ -1,1 +1,1 @@
-Allow sending `sendPasswordReset` if self registration is enabled. @MrTango
+In the `@users` add endpoint, ignore the `sendPasswordReset` parameter if self registration is enabled. @MrTango
--- a/src/plone/restapi/services/users/add.py
+++ b/src/plone/restapi/services/users/add.py
@@ -70,6 +70,9 @@ class UsersPost(Service):
             required.append("username")
             allowed.append("username")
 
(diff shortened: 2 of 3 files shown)


stevepiercy (approved):

MrTango:
this fixes #1888

MrTango:

MrTango:
@davisagli @sneridagh I don't really understand, why we have this `sendPasswordReset` on client side.
And since Volto is sending it always in the register form, the form does not work as expected.
As I see it, when `security_settings.enable_self_reg = True`, we should either allow sending this flag or don't send it at all from the register form.

davisagli (commented):
I'll accept this as a workaround -- but note that the sendPasswordReset value is forced to True on line 203 unless the user has permission to manage users or the enable_user_pwd_choice setting is enabled. So I think it's a bit misleading to accept the parameter when it will just be ignored.

In my opinion the better fix would be in Volto: If enable_user_pwd_choice is False, the self registration form should create a user without passing password or sendPasswordReset (and the code here will assume sendPasswordReset=True).

(Currently Volto doesn't try to support enable_user_pwd_choice=True at all. The form will always break in that case because it doesn't include inputs for the user to enter a password.)

news/1888.bugfix
```diff
@@ -0,0 +1 @@
+Allow sending `sendPasswordReset` if self registration is enabled. @MrTango
```
davisagli, line 1: This is a bit more accurate (and in plone.restapi release notes we always try to name the endpoint to give some context):

```suggestion
In the `@users` add endpoint, ignore the `sendPasswordReset` parameter if self registration is enabled. @MrTango
```

davisagli (approved):

PR #1898, opened 2025-03-21
Add test case for failure in search call

tiberiuichim:


tiberiuichim:
Tests fail with:

```
Error in test test_expand (plone.restapi.tests.test_search_utils.TestUnflattenDottedDict)
  File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/unittest/case.py", line 60, in testPartExecutor
    yield
  File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/unittest/case.py", line 676, in run
    self._callTestMethod(testMethod)
  File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/unittest/case.py", line 633, in _callTestMethod
    method()
  File "/home/runner/work/plone.restapi/plone.restapi/src/plone/restapi/tests/test_search_utils.py", line 31, in test_expand
    unflatten_dotted_dict(dct),
  File "/home/runner/work/plone.restapi/plone.restapi/src/plone/restapi/search/utils.py", line 35, in unflatten_dotted_dict
    inner = reduce(create_or_get, [result] + key_segments[:-1])
  File "/home/runner/work/plone.restapi/plone.restapi/src/plone/restapi/search/utils.py", line [28](https://github.com/plone/plone.restapi/actions/runs/13998116091/job/39198164883?pr=1898#step:6:29), in create_or_get
    return dct.setdefault(key, {})
AttributeError: 'str' object has no attribute 'setdefault'
```

davisagli:
The search endpoint doesn't support expanders. The use of them here suggests that maybe the frontend is trying to call the `@search` endpoint as a content route when it shouldn't.

The use of unflatten_dotted_dict here is meant for deserializing structured queries like `path.depth=1`. ~~I'll adjust it to raise a BadRequest instead of an AttributeError if it encounters a key that was already set to a string.~~ On second thought, I'll make it transfer the str value to the "query" key of the new dict, which is what actually makes sense in the context of search queries.

davisagli (approved):

PR #1902, opened 2025-03-26
Add tests for @inherit service endpoint 

### LINKED ISSUE #1897: Add a test of a behavior with fields to `@inherit`
> @erral @stevepiercy I considered the documentation tests to be a fairly adequate functional test. They confirm that the endpoint exists, can be called without an error, and returns data in the expected format -- and they'll break if that format changes unexpectedly.
> 
> As this issue revealed, they didn't cover the case of a behavior that actually includes fields. That was me being lazy because there wasn't an obvious built-in behavior that would be a good test case. It would be good to add a test that does this with a new behavior created for the purposes of the test. I'm drowning in tasks and probably won't get to this anytime soon. 

_Originally posted by @davisagli in [#1895](https://github.com/plone/plone.restapi/issues/1895#issuecomment-2734196614)_

ujsquared:
Fixes #1897 

here's my contribution for the Bucharest Sprint! 🎉

----
📚 Documentation preview 📚: https://plonerestapi--1902.org.readthedocs.build/

### DIFF
--- a/news/1897.feature
+++ b/news/1897.feature
@@ -1,0 +1,1 @@
+Add tests for inherit service. @ujsquared
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,7 @@ import pathlib
 import sys
 
 
-version = "9.13.2"
+version = "9.13.3.dev0"
 
 if sys.version_info.major == 2:
     raise ValueError(
--- a/src/plone/restapi/tests/test_services_inherit.py
+++ b/src/plone/restapi/tests/test_services_inherit.py
@@ -171,7 +171,7 @@ class TestServiceInherit(unittest.TestCase):
         response = self.api_session.get(url)
 
         self.assertEqual(response.status_code, 200)
-        data = response.json()["inherit"]
+        data = response.json()
         behavior_data = data["plone.testbehavior.ITestBehavior"]
 
         self.assertEqual(behavior_data["data"]["test_field"], "Inherited Value")
@@ -184,7 +184,7 @@ class TestServiceInherit(unittest.TestCase):
     def test_inherit_service_invalid_behavior(self):
         url = f"{self.child.absolute_url()}/@inherit?expand.inherit.behaviors=invalid.behavior"
         response = self.api_session.get(url)
-        self.assertNotIn("invalid.behavior", response.json()["inherit"])
+        self.assertNotIn("invalid.behavior", response.json())
 
     def test_inherit_multiple_behaviors(self):
         self.register_behavior(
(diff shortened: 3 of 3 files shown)


davisagli:
@ujsquared Wow, thanks!

There are 5 tests which have errors, you can see them here: https://github.com/plone/plone.restapi/actions/runs/14088663165/job/39459417111?pr=1902

I need to look at why the test-matrix builds were not triggered for your PR from your own fork. We intend for this to work...

ujsquared:
> @ujsquared Wow, thanks!
> 
> There are 5 tests which have errors, you can see them here: https://github.com/plone/plone.restapi/actions/runs/14088663165/job/39459417111?pr=1902
> 
> I need to look at why the test-matrix builds were not triggered for your PR from your own fork. We intend for this to work...



> @ujsquared Wow, thanks!
> 
> There are 5 tests which have errors, you can see them here: https://github.com/plone/plone.restapi/actions/runs/14088663165/job/39459417111?pr=1902
> 
> I need to look at why the test-matrix builds were not triggered for your PR from your own fork. We intend for this to work...

no worries! I'll take a look.

davisagli:
@ujsquared You might need to update your branch. I made a recent change to fix the response of the `@inherit` endpoint in https://github.com/plone/plone.restapi/pull/1900 (It's supposed to have a top-level "inherit" key when it's used as an expander inside `@components`, but not when it's called as a standalone endpoint)

ujsquared:
> @ujsquared You might need to update your branch. I made a recent change to fix the response of the `@inherit` endpoint in #1900 (It's supposed to have a top-level "inherit" key when it's used as an expander inside `@components`, but not when it's called as a standalone endpoint)

😮
should've git pulled in the updated branch.

davisagli (approved):
@ujsquared I am happy to merge this as is -- but I'll give some suggestions on what I would do different, and let you decide if you want to change anything.

1. I would avoid using the methods `register_behavior` and `create_folder_structure`, and make the tests do this setup directly, even though there is more repetition. When I was learning how to write unit tests, I was given the advice that this is not the place to factor things out to avoid repeating code. It makes it harder to read what scenario is being tested.
2. I don't think it's necessary to include IValidationBehavior. Validation is only done when writing to a field, and this service is read-only.

Let me know if I should just go ahead and merge -- it's fine if you don't want to spend more time on this.

davisagli:
@ujsquared Thanks!

ujsquared:
>     1. I would avoid using the methods `register_behavior` and `create_folder_structure`, and make the tests do this setup directly, even though there is more repetition. When I was learning how to write unit tests, I was given the advice that this is not the place to factor things out to avoid repeating code. It makes it harder to read what scenario is being tested.

I was fed up at that point with rewriting the registration and folder creation parts due to errors. Made the code a little less redundant I thought. But it was just two instances. Earlier versions I wrote had more occurrences, so I assumed more than 2 uses.    


>     2. I don't think it's necessary to include IValidationBehavior. Validation is only done when writing to a field, and this service is read-only.

Okay, will keep in mind for future.

Thanks for merging!

PR #1903, opened 2025-03-31
WIP: Option to transform primary field target to provide information about the target rather than just it's URL

JeffersonBledsoe:


JeffersonBledsoe:
@davisagli I've been having some issues trying to complete this and wanted to check it made sense as it would be a breaking change without the environment variable. The functionality mostly works well, but as you can see from the [latest test runs](https://github.com/plone/plone.restapi/actions/runs/18165012286/job/51704870842), the `@id` field is also being expanded, when this should really still be just a string to the URL of the referenced object. I see the following ways of moving forward with this PR:

- Find a way to check whether we're converting an `@id` field and ensure we don't run the expansion then (unsure how to do this currently)
- Change approach completely. We could have an additional field of "expanded objects" where the URLs that get created from the `resolveuid` expansions are the keys and the values contain information from the objects.
- I've misunderstood the primary file target field along the way and there's another way to get information about a file field without making another request.

davisagli:
@JeffersonBledsoe I think the idea is sound in general; the API is not providing enough information here.

In terms of backwards-compatibility, I'd suggest following this path:
1. add it on main but only when the environment variable is present
2. test volto with the environment variable present and make any necessary updates to handle the new structure
3. make a future major release of plone.restapi that has it enabled by default

About the extra expansion of the `@id`, I think the problem is that https://github.com/plone/plone.restapi/blob/f441068fd6be8e282c24bc9d24a2a47ec27a3b34/src/plone/restapi/serializer/blocks.py#L54 is really expecting resolve_uid to return a str, but now you've made a change that makes it return a dict in this case.

Possible solutions:
1. Update ResolveUIDSerializerBase to handle a return value that is a dict, and merge it into the existing dict (but maybe we'll run into other things that are expecting resolve_uid to always return a str)
2. Move your code to do extra expansion into ResolveUIDSerializerBase: make it check if the resolved str URL includes @@download, and if so add the other fields. This would be a bit similar to the special case it already has for image_scales.

PR #1912, opened 2025-04-23
Don't include teaser target data if user doesn't have permission

### LINKED ISSUE #1911: Teaser block serializer should not include content the current user cannot view
1. Create a private page.
2. Create another page which has a teaser that links to the private page.
3. Publish the second page.
4. Log out.
5. View the published page.

Expected: The teaser is not shown, because the user doesn't have permission to view that content.
Actual: The teaser is shown, OR there is an Unauthorized error. (Because the serializer uses the summary serialization of the item, and depending on the configured fields it can ends up using getObject, which checks the permission)

davisagli:
Fixes #1911 

----
📚 Documentation preview 📚: https://plonerestapi--1912.org.readthedocs.build/

### DIFF
--- a/news/1911.bugfix
+++ b/news/1911.bugfix
@@ -1,0 +1,1 @@
+In the teaser block serializer, don't include data from the target item if the current user doesn't have View permission on it. @davisagli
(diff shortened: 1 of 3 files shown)


davisagli:

ericof (approved):

sneridagh (approved):

Issue #1913, opened 2025-04-24
Error serialising empty reference (Plone 4 Archetypes)


mauritsvanrees:
I am using `collective.exportimport` to export Archetypes content from a Plone 4 site.  I get errors for a few items because they have references that are None:

```
Traceback (most recent call last):
  File "/srv/pw/buildout-cache/eggs/collective.exportimport-1.14-py2.7.egg/collective/exportimport/export_content.py", line 390, in export_content
    item = serializer()
  File "/srv/pw/buildout-cache/eggs/plone.restapi-7.9.0-py2.7.egg/plone/restapi/serializer/atcontent.py", line 76, in __call__
    result[name] = serializer()
  File "/srv/pw/buildout-cache/eggs/plone.restapi-7.9.0-py2.7.egg/plone/restapi/serializer/atfields.py", line 129, in __call__
    return [json_compatible(r.absolute_url()) for r in refs]
AttributeError: 'NoneType' object has no attribute 'absolute_url'
```

I will create a PR.

PR #1915, opened 2025-04-24
Fix error serializing empty reference from Archetypes content. [7.x.x]

mauritsvanrees:
Fixes https://github.com/plone/plone.restapi/issues/1913

### DIFF
--- a/news/1913.bugfix
+++ b/news/1913.bugfix
@@ -1,0 +1,2 @@
+Fix error serializing empty reference from Archetypes content.
+[maurits]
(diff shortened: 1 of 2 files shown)


mauritsvanrees:
@davisagli The tests fail due to using no longer available action versions and Ubuntu 20.
Can we ignore that for this small fix?

I would rather make an internal release than put any time in fixing the CI for this old version.

davisagli (approved):
@mauritsvanrees Looks okay to me.

mauritsvanrees:
Thanks for the review.
I have released `plone.restapi` 7.9.1 with this fix.

PR #1887, opened 2025-03-12
Add `@inherit` service to get inherited behavior values

davisagli:
This adds a new service, `@inherit`, to explicitly get values from behaviors assigned to ancestors in the object hierarchy.

(It's Acquisition, but without the bad magic.)

----
📚 Documentation preview 📚: https://plonerestapi--1887.org.readthedocs.build/en/1887/endpoints/inherit.html

PR #1916, opened 2025-04-24
Fix permission check in schema serializer

maethu:
Env:
Plone == 6.1.1
Python == 3.12.8
plone.restapi == 9.13.3

Issue: Exposing of protected information

Reproduce:

Protect any field with a permission not given to an anonymous user
Make request with Manager account, which has the permission --> Field is in response
Make a request as an anonymous user to same object, WITHOUT the permission --> Field is still there!!
Reboot the instance and make a new request as an anonymous user. --> Field is gone.
I'm not sure if this is intentional, or not.
But in my case, it exposes information I don't want to.

It got implemented here: [Add `@inherit` service to get inherited behavior values (#1887) · plone/plone.restapi@763ce18 · GitHub](https://github.com/plone/plone.restapi/commit/763ce189400c4640f7b05ebbc1c3ffcefd5c7eae#diff-74ad4dd30dcaee75ebfef37954d68ac69de8d5abf4a83e02b52d525fd3756407R41-R49)

Which clearly is meant to be stored on the object, but no committed (volatile).

Before it was "cached" only in the serializer instance: [Add `@inherit` service to get inherited behavior values (#1887) · plone/plone.restapi@763ce18 · GitHub](https://github.com/plone/plone.restapi/commit/763ce189400c4640f7b05ebbc1c3ffcefd5c7eae#diff-faaee0c689417efa563809ccb96137e21ce38e7f38c3700dcb57fdf59116f0b5L78) Which is IMHO the correct way to do this.

Fix:
Store it again on the serializer instance

----
📚 Documentation preview 📚: https://plonerestapi--1916.org.readthedocs.build/

### DIFF
--- a/news/1916.bugfix
+++ b/news/1916.bugfix
@@ -1,0 +1,2 @@
+Fix exposing protected fields in schema serializer.
+[maethu]
--- a/src/plone/restapi/deserializer/dxcontent.py
+++ b/src/plone/restapi/deserializer/dxcontent.py
@@ -7,13 +7,13 @@ from plone.restapi import _
 from plone.restapi.deserializer import json_body
 from plone.restapi.interfaces import IDeserializeFromJson
 from plone.restapi.interfaces import IFieldDeserializer
+from plone.restapi.serializer.schema import _check_permission
 from plone.supermodel.utils import mergedTaggedValueDict
 from z3c.form.interfaces import IDataManager
 from z3c.form.interfaces import IManagerValidator
 from zExceptions import BadRequest
 from zope.component import adapter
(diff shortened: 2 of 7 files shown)


maethu:

davisagli (commented):
@maethu Thanks for the analysis and fix. You're right, I forgot that volatile attributes can last longer than a single transaction if the persistent object is not changed.

I need to make some adjustments, because `check_permission` is also called from the dxcontent serializer.

davisagli:

davisagli:
@maethu Released in plone.restapi 9.13.5

PR #1926, opened 2025-05-27
Optimize `@navroot` endpoint (#1841)

### LINKED ISSUE #1841: Improve `@navroot` endpoint by removing nested data from the root object
The `@navroot` endpoint now serializes the navroot object in full, including some fields that are superfluous or not needed, some that can cause a big chunk of unneeded nested data like `items` or the `components` information.

We would like to improve it and polish what it returns.

pgrunewald:
fields `items` and `components` are no longer computed/returned.

Fixes #1841 

----
📚 Documentation preview 📚: https://plonerestapi--1926.org.readthedocs.build/

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,23 @@
 
 <!-- towncrier release notes start -->
 
+## 9.15.0 (2025-06-10)
+
+
+### New features:
+
+- Include all summary fields when serializing `next_item` and `previous_item`. @davisagli 
+
+
+### Bug fixes:
+
+- In the `RelationListFieldSerializer`, fix `list` used a default argument. @ale-rt #1922
+
+
+### Documentation:
+
+- Fix contributor documentation. @davisagli 
+
 ## 9.14.0 (2025-05-12)
 
 
--- a/news/+1925.documentation
+++ b/news/+1925.documentation
@@ -1,1 +1,0 @@
-Fix contributor documentation. @davisagli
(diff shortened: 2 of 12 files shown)


pgrunewald:

news/1841.bugfix
```diff
@@ -0,0 +1 @@
+Optimize `@navroot` endpoint: fields `items` and `components` are no longer computed/returned. @pgrunewald
```
davisagli, line 1:
```suggestion
Optimize `@navroot` endpoint: fields `items` and `@components` are no longer computed/returned. @pgrunewald
```

src/plone/restapi/services/navroot/get.py
```diff
@@ -28,7 +28,7 @@ def __call__(self, expand=False):
         result["navroot"]["navroot"] = getMultiAdapter(
             (portal_state.navigation_root(), self.request),
             ISerializeToJson,
-        )()
+        )(include_items=False, include_expansion=False)
```
davisagli, line 31: @pgrunewald There's a non-zero chance someone has registered a custom serializer which does not accept these keyword args. So we should probably be careful and first try this way, then fall back to calling it with no parameters.

davisagli (changes requested):

news/1841.bugfix
pgrunewald: Corrected. Thanks

src/plone/restapi/services/navroot/get.py
pgrunewald: Good catch. I added also a test with a custom serializer w/o keyword args to see, if it works.

PR #1930, opened 2025-05-30
Fix a redirect to RTD after running linkcheck

stevepiercy:


stevepiercy:
This PR is just a maintenance task and doesn't need a change log entry cluttering up history.

PR #1936, opened 2025-07-04
Fix tests for updated rolemap

petschki:
Test together with https://github.com/plone/Products.CMFPlone/pull/4042

### DIFF
--- a/news/+rolemap_testfix.tests.md
+++ b/news/+rolemap_testfix.tests.md
@@ -1,0 +1,1 @@
+Testfix for https://github.com/plone/Products.CMFPlone/pull/4042. @petschki
--- a/src/plone/restapi/tests/test_search.py
+++ b/src/plone/restapi/tests/test_search.py
@@ -660,27 +660,6 @@ class TestSearchFunctional(unittest.TestCase):
         ).json()
         self.assertEqual(response["items_total"], 1)
 
-        # not admin users can't see expired items
-        self.api_session.auth = ("editoruser", TEST_USER_PASSWORD)
-
-        response = self.api_session.get("/@search", params={}).json()
-        if HAS_PLONE_6:
-            # Since Plone 6 the Plone site is indexed ...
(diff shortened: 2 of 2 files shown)


news/+rolemap_testfix.tests.md
```diff
@@ -0,0 +1 @@
+Testfix for https://github.com/plone/Products.CMFPlone/pull/4042. @petschki
```
stevepiercy:
```suggestion
Fixed test to allow Contributor and Editor to access inactive portal content. See https://github.com/plone/Products.CMFPlone/pull/4042. @petschki
```

stevepiercy (changes requested):

src/plone/restapi/tests/test_search.py
```diff
@@ -718,6 +697,16 @@ def test_respect_access_inactive_permission(self):
...
 
+        # unauthorized users can not access expired content
+        self.api_session.auth = None
+        response = self.api_session.get("/@search", params={}).json()
+        if HAS_PLONE_6:
+            # Since Plone 6 the Plone site is indexed ...
+            self.assertEqual(response["items_total"], 1)
+        else:
+            # ... before it was not
+            self.assertEqual(response["items_total"], 0)
```
davisagli, line 708: The expired items need to be published to make sure they are filtered out here because of their expiration date and not because of allowedRolesAndUsers

davisagli:
Also, keep in mind plone.restapi still supports Plone 5 and 6.0. So this will need to be a test that runs conditionally based on the Plone version.

stevepiercy (approved):

petschki:
> Also, keep in mind plone.restapi still supports Plone 5 and 6.0. So this will need to be a test that runs conditionally based on the Plone version.

Does this rolechange gets backported to Plone 6? otherwise I'll check for the new roles in 6.1/2 separately ...

PR #1946, opened 2025-09-14
dummy change

gforcada:
- **chore: reorder MANIFEST.in**
- **Add news entry**

-----

Closes #

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,7 +1,19 @@
 graft src/plone/restapi
 graft docs
+
 include *.md
 include *.txt
+include pyproject.toml
+include netlify.toml
+
+global-exclude *.pyc
+recursive-exclude performance *
+prune docs/_build
+prune docs/styles/Microsoft
+prune src/plone/restapi/tests/images
+prune news
+prune ideas
+
 exclude *.cfg
 exclude *.yml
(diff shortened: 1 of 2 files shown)


gforcada:
Aha, so indeed, this `Uncommitted doc changes` fails here as well. Might this be that some PRs got merged without adding these changes?

I see that there are some locking related information and some context variable being added on some outputs 🤷🏾

stevepiercy:
I couldn't find the commit where it went sideways, if that's what it actually did. What does  this new check actually do? It appears that something changed with translations. I'm not clear, so I'd defer to @erral.

davisagli:
@gforcada @stevepiercy tl;dr this is normal but I need to make a commit to update the test response files.

When we run the plone.restapi tests, it captures the responses from a bunch of API calls and updates files which are included in the docs. This check makes sure that those aren't changing unexpectedly. If it fails, it means that something changed in other Plone packages which caused the response to be different. Then we check to make sure the change is expected, and if so we commit the changes to the response files.

In this case the new context variable is expected from something I added in plone.rest. I need to check where the locking thing comes from but that also sounds familiar.

(This check is not new, btw. It's been an important part of the plone.restapi test suite for years.)

PR #1947, opened 2025-09-21
Added expandable in site endpoint

sneridagh:
@davisagli Is there any reason why this hasn’t been implemented yet? I mean, it was there but not wired up… We need this for Seven, and eventually for Volto.

----
📚 Documentation preview 📚: https://plonerestapi--1947.org.readthedocs.build/

### DIFF
--- a/news/1947.feature
+++ b/news/1947.feature
@@ -1,0 +1,1 @@
+Site endpoint `site` is now expandable. @sneridagh


sneridagh:

davisagli:
@sneridagh We intentionally removed it from the PR that added the site endpoint: https://github.com/plone/plone.restapi/pull/1465#discussion_r1209590386

My reasoning was that it is an anti-pattern because the site data should be fetched once on the SSR load, and not every time new content is fetched.

There is a tradeoff because if it's only fetched on the initial load, it might get out of date if settings that affect the site data are changed later. So I can see the argument for making it available as an expander -- but I still don't like that it does extra work on every content request that isn't needed in most cases.

Can you say more about how you plan to use it in Seven and why it's not sufficient to use the separate endpoint?

sneridagh:
@davisagli I understand your point, that's why we should put some memoization in place, so if the attributes do not change, then you don't do the work again.

Why this can't be done in Seven? Because HTTP and how modern SSR works where all requests are isolated from each other. In Volto this can be done because Redux store. I don't want such a thing or poor man replacement in Seven.

In Volto are you aware that since the TTW settings that we are pushing in there, eg. you change the square images setting. It does not take place until you reload the whole app? This would fix it.

sneridagh:
@davisagli btw, the memoization, could you please help me with it? I'd like to have this soon, so Volto and the distributions can also benefit from it.

sneridagh:
@davisagli we decided _not_ to pursue this anymore, right?
Closing, open it again otherwise.

davisagli:
@sneridagh Right, I don't think it makes sense to include sitewide configuration in every content serialization.

PR #1959, opened 2025-11-24
Pep 420 native namespace

davisagli:

-----

Closes #

----
📚 Documentation preview 📚: https://plonerestapi--1959.org.readthedocs.build/

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,14 +51,10 @@ If it is not installed, add it using requirements.txt or as a dependency of anot
 
 # Python / Plone Compatibility
 
-plone.restapi 10 requires Python 3 and works with Plone 6.2.
-This version stopped using `pkg_resources` for the `plone` namespace.
-It might be possible to use plone.restapi 10 with an older version of Plone, but if you install it as an editable package you have to use [horse-with-no-namespace](https://pypi.org/project/horse-with-no-namespace/).
+plone.restapi 10 requires Python 3 and works with Plone 6.2, 6.1, 6.0, and 5.2.
 
-plone.restapi 9 requires Python 3 and works with Plone 5.2, Plone 6.1, and Plone 6.0.
-
-plone.restapi 8 entered "maintenance" mode with the release of
-plone.restapi 9 (September 2023). It is not planned to backport any
+plone.restapi 9 entered "maintenance" mode with the release of
+plone.restapi 10 (end of 2025). It is not planned to backport any
 features to this version and we highly recommend to upgrade to
 the latest version of plone.restapi.
 
--- a/setup.py
+++ b/setup.py
@@ -1,9 +1,16 @@
 from setuptools import setup
 
 import pathlib
+import sys
 
 
 version = "10.0.0.dev0"
+
+if getattr(sys.version_info, "major") == 2:
+    raise ValueError(
+        "plone.restapi 10 requires Python 3. "
+        "Please downgrade to plone.restapi 7 for Python 2 and Plone 4.3/5.1."
+    )
 
 long_description = "\n".join(
     [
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -51,13 +51,16 @@ If it is not installed, add it using requirements.txt or as a dependency of anot
 
 # Python / Plone Compatibility
 
-plone.restapi 9 requires Python 3 and works with Plone 5.2 and Plone
-6.x.
+plone.restapi 10 requires Python 3 and works with Plone 6.2.
+This version stopped using `pkg_resources` for the `plone` namespace.
+It might be possible to use plone.restapi 10 with an older version of Plone, but if you install it as an editable package you have to use [horse-with-no-namespace](https://pypi.org/project/horse-with-no-namespace/).
+
+plone.restapi 9 requires Python 3 and works with Plone 5.2, Plone 6.1, and Plone 6.0.
```
mauritsvanrees: Let's use the natural sort order:

```suggestion
plone.restapi 9 requires Python 3 and works with Plone 5.2, Plone 6.0, and Plone 6.1.
```

mauritsvanrees (commented):
I would be in favour of dropping support for 6.1 and earlier completely, allowing to get rid of lots of backwards compatibility imports.  But I can understand your desire to keep compatibility as long as it works and the effort is reasonable.

Could you restore the testing in 6.0 and 6.1 then?  Now it is tested on 5.2 and 6.2 and not in between, which seems strange.

Maybe in the readme add: "`plone.restapi` 10 is not expected to work on Plone 5.2 when you use `zc.buildout`."  You would need buildout 5 for that, which requires Python 3.9+.

BTW, locally on my Mac 5.2 did not work because `AccessControl` 4.4 could not be compiled.

davisagli:
> I would be in favour of dropping support for 6.1 and earlier completely, allowing to get rid of lots of backwards compatibility imports. But I can understand your desire to keep compatibility as long as it works and the effort is reasonable.

I plan to keep supporting Plone 5.2 until we have Plone 7, unless we run into a real blocker for doing that. The goal is to be able to use a single version of plone.restapi across two major backend versions to help make upgrades simpler.

> Could you restore the testing in 6.0 and 6.1 then? Now it is tested on 5.2 and 6.2 and not in between, which seems strange.

Yeah, I was also thinking about this. Done.

> Maybe in the readme add: "plone.restapi 10 is not expected to work on Plone 5.2 when you use zc.buildout." You would need buildout 5 for that, which requires Python 3.9+.

I tested it in Plone 5.2 with buildout 3.0.1 and 3.3 and it worked fine. In this case plone.restapi was the only editable package in the plone namespace, so sys.path was set up entirely by the buildout script and pkg_resources.declare_namespace was never invoked. It might not work out so well if other plone.* packages were editable, but then I think horse-with-no-namespace would make it work.

Anyway, we don't need to support developing plone.restapi on Plone 5.2, just installing the release. So I took out the mention of horse-with-no-namespace from the README to avoid confusion.

> BTW, locally on my Mac 5.2 did not work because AccessControl 4.4 could not be compiled.

That's been a separate issue for a long time where old versions of AccessControl used some C that is not allowed by default in recent versions of clang. Here are the flags I usually have to set when running old Plone 5 buildouts on recent macOS:
```
export _PYTHON_HOST_PLATFORM="macosx-11-x86_64"
export CFLAGS="-Wno-error=int-conversion -Wno-error=incompatible-function-pointer-types -Wno-error=implicit-function-declaration -I/opt/homebrew/opt/zlib/include -I/opt/homebrew/opt/openssl@1.1/include"
export LDFLAGS="-L/opt/homebrew/opt/zlib/lib -L/opt/homebrew/opt/openssl@1.1/lib"
```

PR #1963, opened 2025-11-29
Remove default limit of 1000 from QuerystringSearch

### LINKED ISSUE #1955: The `querystring-search` endpoint limits the result to 1000 items
**Describe the bug**

When we use the `querystring-search` endpoint without the `limit` parameter, it limits the result to 1000 items. However, the who requesting the search doesn't know how many items there are.

**To Reproduce**

1. Create more than 1000 Documents on the Portal and publish.
2. Execute the command:

```bash
curl -i -X POST http://localhost:8080/Plone/@querystring-search -H "Accept: application/json" -H "Content-Type: application/json" --data-raw '{"query": [{"i": "portal_type", "o": "plone.app.querystring.operation.selection.any", "v": ["Document"]}]}'
```

**Expected behavior**

All documents must be returned.

**Software (please complete the following information):**

- OS: Ubuntu
- Browser: Chrome
- Plone Version: 6.1.3
- Volto Version: 18.29.0

**Additional context**

We are having this problem in the Volto lists block.

The limitation occurs here:

https://github.com/plone/plone.restapi/blob/fda897320ee56c258123bc974ed8630573ec5c1e/src/plone/restapi/services/querystringsearch/get.py#L41

wesleybl:
Previously, QuerystringSearch imposed a hardcoded limit=1000 on all queries, even when the user did not specify a limit. This caused unexpected truncation of results.

Now the limit parameter is only passed to querybuilder when explicitly provided in the request data.

Closes #1955 

----
📚 Documentation preview 📚: https://plonerestapi--1963.org.readthedocs.build/

### DIFF
--- a/docs/source/endpoints/querystringsearch.md
+++ b/docs/source/endpoints/querystringsearch.md
@@ -138,7 +138,7 @@ The sort_order can be either `ascending` or `descending`.
 
 ### Limit
 
-Querystring `query` with a `limit` parameter:
+Use the `limit` parameter to set a maximum number of results that will be returned:
 
 ```
 {
@@ -154,7 +154,7 @@ Querystring `query` with a `limit` parameter:
 ```
 
 The `limit` parameter is optional.
-The default value is `1000`.
+The default value is no limit (but a single page of results will still have a size determined by the Batch Size).
 
 ### Query
 


davisagli (changes requested):
The default limit is there for a reason, to prevent accidental long-running requests that return too many results. I would rather keep the limit but fix the response so that it reports the total result count correctly.

wesleybl:
@davisagli But isn't that hide the results?

davisagli:
@wesleybl Yes, unless you specifically ask for more. That's intentional. (But setting items_total to the limit, instead of the actual total, is not intentional.)

wesleybl:
> Yes, unless you specifically ask for more. That's intentional.

@davisagli But the frontend doesn't know how many records exist, so it can't pass a value. This limits access to the records the user might want. And with pagination, it will never return 1000 items.

Note that a normal search doesn't have this limit: https://github.com/plone/plone.restapi/blob/main/src/plone/restapi/search/handler.py

wesleybl:

wesleybl:
> But setting items_total to the limit, instead of the actual total, is not intentional.

If we change this, it will cause the pagination to return pages that are not reachable, causing an error. Here's the PR that inserted this:

https://github.com/plone/plone.app.querystring/pull/17

wesleybl:
@davisagli So what's your suggestion for fixing this problem? Could we create a setting in Volto to increase this limit? But honestly, I don't think that would be the best solution.

davisagli:
For one thing, we should make sure that the batch class used for pagination is not broken by having an actual_result_count that is higher than the limit. The whole point of the catalog returning actual_result_count is to provide this information.

I need to take a closer look at all the parts of this. It may be that we can remove the default limit but I need to check some things.

wesleybl:
@davisagli I still don't quite understand your concern. Are you worried that "accidentally" more than 1000 items might be returned? Would that be through a call without parameters? Can you give an example of a call that "accidentally" returns multiple items? Today we have a default value for `b_size`:

https://github.com/plone/plone.restapi/blob/main/src/plone/restapi/services/querystringsearch/get.py#L38

So calls without parameters won't return 1000 items, even without the limit. To return 1000, we would have to pass `b_size=1000`. In my opinion, this is not accidental.

davisagli (approved):
Okay, coming back to this after some time away. Yeah, I think you're right that it's safe to remove the limit, because the default results will still be limited by the batch size.

Earlier I was worried that removing the limit would no longer give the catalog enough information to know when it can optimize sorting for a smaller result set. But it turns out that worry was unnecessary:
- The catalog picks a sort_limit based on the b_start and b_size if the limit isn't passed explicitly
- As it turns out, plone.app.querystring was not passing the limit along to the catalog query anyway! (It was just using it to override actual_result_count on the lazy resultset.)

wesleybl:
@davisagli Can we have this in restapi 9?

davisagli:
@wesleybl Sure, I think we can call it a bugfix rather than a breaking change.

davisagli:
@wesleybl But also: plone.restapi 10.x is still tested with the same versions of Plone. It should work fine unless you have to install it as an editable package. Then you might need horse-with-no-namespace in order to make the plone namespace package work with both pkg_resources and native namespaces.

Issue #1792, opened 2024-06-15
@aliases endpoint should support filtering by date (from, to)


tisto:
PR for adding aliases: https://github.com/plone/plone.restapi/pull/1398

The @aliases endpoint allows to filter by path:

http://localhost:8080/Plone/++api++/@aliases&q=/de/ft/my-path

However, there is no way to filter by a date range. This is something the ClassicUI controlpanel allows. I propose to add two new parameters (from, to) to be able to filter for a date range:

http://localhost:8080/Plone/++api++/@aliases&q=/de/ft/my-path&from=2021-11-04&to=2022-01-01

stevepiercy:
For new parameter names, Plone REST API uses `start` and `end` attributes for event content types, instead of `from` and `to`.

https://6.docs.plone.org/plone.restapi/docs/source/endpoints/content-types.html#event

I don't feel strongly about parameter names, but I think consistency is helpful.

tisto:
@stevepiercy good point! Consistency is indeed crucial. I was under the assumption that plone.app.querystring and zcatalog use from/to. Seems ZCatalog uses just a query tuple without specifying a name for those parameters though.

However, I am not sure if the Event content type is sufficient for the argument. Those are attributes of the Event content type in Plone not necessarily the way date ranges are queried.

In any case. I am undecided and would appreciate more people to chime in. @davisagli @thet @sneridagh maybe?

stevepiercy:
Indeed, there may be many instances of date range parameters, and I lack familiarity with the code base to be authoritative. Someone with more familiarity than I should chime in.

Also hopefully interpretation of the two date parameters would allow the omission of one or the other parameter, effectively allowing a minimum or maximum date without an upper or lower bound.

Faakhir30:
I see that these `start` and `end` filter params are not supported by `RedirectControlPanal` in `plone/ProductCMFPlone`.

Another solution could be dealing with this filtering in `plone.restapi` directly once we get a response from `RedirectControlPanel`, but as other params like `query` are being dealt with in `Products.CMFPlone`, so I'll be adding there first, then will integrate these params here in `plone.restapi` also. 

I am open to suggestions. Thanks, @tisto @stevepiercy.

Issue #1960, opened 2025-11-26
Inconsistent usage of boolean URL parameters (true, false, yes, no, True, False, 1, 0)


tisto:
Claude Code Prompt (on plone.restapi checkout):

> analyse the code base of plone.restapi which url parameters are used for
boolean values (e.g. True, False, true, false, yes, no, 0, 1).

⏺ I'll analyze the plone.restapi codebase to identify how boolean URL
  parameters are handled and what formats are accepted.

⏺ Explore(Analyze boolean URL parameters)
  ⎿  Done (36 tool uses · 71.4k tokens · 1m 46s)

⏺ I've completed a comprehensive analysis of boolean URL parameter handling
  in the plone.restapi codebase. Here are the key findings:

  Main Boolean Conversion Function

  The primary function is boolean_value() located in
  src/plone/restapi/deserializer/__init__.py:20-31:

  def boolean_value(value):
      return value not in {False, "false", "False", "0", 0}

  This uses a blacklist approach - only these specific values return False:
  - False (boolean)
  - "false" (lowercase string)
  - "False" (uppercase string)
  - "0" (string)
  - 0 (integer)

  Everything else returns True, including: True, "true", "True", "1", 1, or
  even "yes", "no", "foobar".

  URL Parameters Using Boolean Values

  1. include_items (Folders & Collections)

  - Files: src/plone/restapi/serializer/dxcontent.py:176,
  src/plone/restapi/serializer/collection.py:19
  - Uses: boolean_value() function
  - Example: GET /folder?include_items=false excludes children items

  2. fullobjects (Search)

  - File: src/plone/restapi/search/handler.py:78
  - Uses: Presence check (if key exists in query)
  - Example: GET /@search?fullobjects=1 returns full objects

  3. use_site_search_settings (Search)

  - File: src/plone/restapi/search/handler.py:86
  - Uses: Presence check (if key exists in query)
  - Example: GET /@search?use_site_search_settings=1 applies Plone search
  settings

  Special Case: Boolean Index Queries

  For catalog boolean index queries, there's a more restrictive parser in
  src/plone/restapi/search/query.py:219-230:

  Only accepts (case-sensitive, lowercase only):
  - "true" or "1" → True
  - "false" or "0" → False

  Any other format raises a QueryParsingError.

  Key Inconsistency ⚠️

  There are two different approaches in the codebase:
  1. boolean_value(): Permissive blacklist - accepts "True", "False",
  "true", "false", "0", "1"
[...]

PR #1966, opened 2025-12-02
Improvements to the @aliases endpoint

### LINKED ISSUE #1791: @aliases endpoint should allow to filter by how an alias has been added (automatically, manually)
PR for adding aliases: https://github.com/plone/plone.restapi/pull/1398

The @aliases endpoint allows to filter by path:

http://localhost:8080/Plone/++api++/@aliases&q=/de/ft/my-path

However, there is no way to filter by how an alias has been added, e.g.:

http://localhost:8080/Plone/++api++/@aliases&q=/de/ft/my-path&manual=true

I'd suggest adding a new parameter "manual" to the GET endpoint so that we can filter by "manual=true" or "manual=false"

jnptk:
Closes #1791, #1792

## Todo:

- [x] implement batching
- [x] update existing test & create new one for non-root
- [ ] make docs clearer

----
📚 Documentation preview 📚: https://plonerestapi--1966.org.readthedocs.build/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,7 @@ local.cfg
 /sources/
 /venv/
 .installed.txt
+/.mxdev_cache/
 
 /docs/styles/Microsoft/
 


docs/source/endpoints/aliases.md
```diff
@@ -135,7 +134,18 @@ Response:
 
 ## Filter aliases
 
-To search for specific aliases, send a `GET` request to the `/@aliases` endpoint on site `root` with a `q` parameter:
+### Parameters
+
```
stevepiercy, line 138: I'd pull ` (optional)` out of the table for all parameters and state that all parameters are optional in a preceding sentence.

```suggestion

All the following parameters are optional.

```

docs/source/endpoints/aliases.md
```diff
@@ -135,7 +134,18 @@ Response:
...
+| Name                 | Type    | Description                                            |
+| -------------------- | ------- | ------------------------------------------------------ |
+| `query` (optional)   | string  | Full-text search. Can match paths or text fields.      |
+| `manual` (optional)  | boolean | Filter by manual or automatically created redirects.     |
+| `start` (optional)   | string  | Filter redirects created **after** this date.          |
+| `end` (optional)     | string  | Filter redirects created **before** this date.         |
+| `b_start` (optional) | integer | Batch start index (offset).                            |
+| `b_size` (optional)  | integer | Batch size (maximum items returned).                   |
+
+To search for specific aliases, send a `GET` request to the `@aliases` endpoint with one or more of the above named parameters:
```
stevepiercy:
```suggestion
To search for specific aliases, send a `GET` request to the `@aliases` endpoint with one or more of the above named parameters as shown in the following example.
```

src/plone/restapi/services/aliases/get.py
```diff
@@ -22,38 +23,53 @@ def __init__(self, context, request):
...
+                "path": redirect,
+                "redirect-to": info[0],
+                "datetime": datetimelike_to_iso(info[1]),
+                "manual": info[2],
+            }
+            aliases.append(redirect)
+
+        batch = HypermediaBatch(self.request, aliases)
+
+        breakpoint()
```
davisagli: You have a leftover breakpoint here (and that's the reason some of the tests are failing)

src/plone/restapi/services/aliases/get.py
```diff
@@ -83,16 +99,14 @@ def __call__(self, expand=False):
...
             return result
-        if IPloneSiteRoot.providedBy(self.context):
-            if self.request.getHeader("Accept") == "text/csv":
-                result["aliases"]["items"] = self.reply_root_csv()
-                return result
-            else:
-                items, items_total = self.reply_root()
+        if self.request.getHeader("Accept") == "text/csv":
+            result["aliases"]["items"] = self.reply_root_csv()
+            return result
```
davisagli, line 146: I'd say if the client is requesting CSV but the context is not the site root, we should return a 406 Not Acceptable response (with a message explaining why)

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/406

docs/source/endpoints/aliases.md
```diff
@@ -135,7 +134,20 @@ Response:
...
 ## Filter aliases
 
-To search for specific aliases, send a `GET` request to the `/@aliases` endpoint on site `root` with a `q` parameter:
+### Parameters
+
+All of the following parameters are optional.
+
+| Name      | Type    | Description                                            |
+| --------- | ------- | ------------------------------------------------------ |
+| `query`   | string  | Full-text search. Can match paths or text fields.      |
```
davisagli, line 143: This doesn't match the code. The parameter is called `q` (and should remain so for backwards-compatibility; it gets complicated if Volto needs to send different parameters depending on what version of the REST API is available)

src/plone/restapi/tests/http-examples/aliases_root_filter.req
```diff
@@ -1,3 +1,3 @@
-GET /plone/@aliases?q=/fizzbuzz HTTP/1.1
+GET /plone/@aliases?query=/fizzbuzz HTTP/1.1
```
davisagli, line 1:
```suggestion
GET /plone/@aliases?q=/fizzbuzz HTTP/1.1
```

src/plone/restapi/services/aliases/get.py
```diff
@@ -22,38 +23,53 @@ def __init__(self, context, request):
...
-        path        - path
-        redirect    - full path with root
-        """
-        batch = RedirectsControlPanel(self.context, self.request).redirects()
-        redirects = [entry for entry in batch]
 
+        if not IPloneSiteRoot.providedBy(self.context):
+            storage = [
+                path for path in storage._paths.items() if path[1][0] == context_path
+            ]
```
davisagli: This part doesn't look right to me.
1. The code below is assuming `storage` is still an IRedirectionStorage, but here you're setting it to a list.
2. We should keep using storage.redirects(context_path) to get the redirects for a non-root item, because it's more efficient. (It looks up the path in storage._rpaths instead of iterating through all paths in storage._paths)

src/plone/restapi/services/aliases/get.py
```diff
@@ -22,38 +23,53 @@ def __init__(self, context, request):
...
+
+        query = form.get("q")
+        if query and query.startswith("/"):
+            min_k = f"{portal_path}/{query.strip('/')}"
+            max_k = min_k[:-1] + chr(ord(min_k[-1]) + 1)
+            redirects = storage._paths.keys(min=min_k, max=max_k, excludemax=True)
+        elif query:
+            redirects = [path for path in storage._paths.keys() if query in path]
+        else:
+            redirects = storage._paths.keys()
```
davisagli: If we're not on a root item, we need to filter down the redirects from above, instead of starting with the full set of redirects from the storage.

src/plone/restapi/services/aliases/get.py
```diff
@@ -22,38 +23,53 @@ def __init__(self, context, request):
...
+
+        aliases = []
         for redirect in redirects:
-            del redirect["redirect"]
-            redirect["datetime"] = datetimelike_to_iso(redirect["datetime"])
+            info = storage.get_full(redirect)
```
davisagli: This is less efficient than it could be. You're getting all the keys above, then doing a lookup here to get the corresponding value (which is a tuple including target path, datetime, and manual). But you could just get both together above using storage._paths.items() instead of .keys(). That way we just have to scan the BTree once instead of doing a full scan followed by individual lookups.

src/plone/restapi/services/aliases/get.py
```diff
@@ -22,38 +23,53 @@ def __init__(self, context, request):
...
+            if form.get("manual") and info[2] != form["manual"]:
```
davisagli: I'd recommend refactoring this to separate the input/output logic (reading form values and writing the response) from the core logic that finds the right items in the storage. (I mean in two functions, where the first uses the second)

I forget what this design principle is called, but it makes it easier to reuse the core logic if you need to provide a different way of accessing it.

news/1791.breaking
```diff
@@ -0,0 +1 @@
+Refactor `@aliases` endpoint to support filtering from root and page context. @jnptk
```
davisagli: `@aliases` service: Add support for filtering aliases for a non-root content item. @jnptk

And this should be .feature, not .breaking

src/plone/restapi/tests/http-examples/aliases_get.resp
davisagli, line 1: It would be good to add tests for filtering on a non-root item.

src/plone/restapi/services/aliases/get.py
davisagli: There's some logic missing to convert from strings in the request form to the correct data types to compare with the data in the storage (i.e. boolean and DateTime) -- compare https://github.com/plone/Products.CMFPlone/blob/master/src/Products/CMFPlone/controlpanel/browser/redirects.py#L192 and https://github.com/plone/Products.CMFPlone/blob/master/src/Products/CMFPlone/controlpanel/browser/redirects.py#L207

That conversion should be done in the first function before passing the converted types to the 2nd function.

davisagli: The boolean conversion should handle both the old values (yes/no) and the new ones we want to support (true/false). We should add a general utility for this because we also need it for other services (see https://github.com/plone/plone.restapi/issues/1960). I suggest copying https://github.com/plone/plone.base/blob/main/src/plone/base/utils.py#L598 (don't import it from there, because plone.base isn't included in Plone 5 which we still support).

davisagli (commented):
I'm making lots of comments here but I think you're basically on the right track!

src/plone/restapi/services/aliases/get.py
jnptk: > We should keep using storage.redirects(context_path) to get the redirects for a non-root item, because it's more efficient. (It looks up the path in storage._rpaths instead of iterating through all paths in storage._paths)

I am doing this because storage._rpaths only stores the path and the redirect like so: `'/Plone/path': OOSet(['/Plone/some/path'])` which is not enough information to filter redirects for a non-root item

docs/source/endpoints/aliases.md
jnptk: `q` will be kept for backwards-compatibility but `query` is preferred

src/plone/restapi/services/aliases/get.py
```diff
@@ -80,19 +109,43 @@ def reply_root_csv(self):
         return content
 
     def __call__(self, expand=False):
+        try:
+            data = json_body(self.request)
```
davisagli: Why are you trying to parse the request body as JSON here? This is a GET endpoint and there is no body. The parameters are in request.form.

Which makes me wonder why there aren't tests failing because of this. Shouldn't test_aliases_root_filter be failing, since it's passing a `q` param which is ignored? Maybe I'm missing something.

(The `@querystring-search` endpoint is unusual in handling both GET and POST methods. Let's not copy that here.)

src/plone/restapi/services/aliases/get.py
```diff
@@ -80,19 +109,43 @@ def reply_root_csv(self):
...
+        if query and not isinstance(query, str):
+            raise BadRequest('Parameter "query" must be a string.')
+
+        if manual and not (is_truthy(manual) or is_falsy(manual)):
+            raise BadRequest('Parameter "manual" must be a boolean.')
+
+        for value in (start, end):
+            if value:
+                try:
+                    value = DateTime(value)
```
davisagli: You're assigning to `value` but not using it below. This doesn't update the value of `start` and `end`. You'll need to assign to each one separately. You can factor the conversion and exception handling out into a helper function if you want to avoid repetition.

And where is the test to prove that the start and end filters are working as intended? If you had written the test it would have caught this problem.

src/plone/restapi/services/aliases/get.py
```diff
@@ -22,40 +31,60 @@ def __init__(self, context, request):
...
+        elif query:
+            redirects = [path for path in storage.items() if query in path]
+        else:
+            redirects = storage.items()
 
-        self.request.form["b_start"] = "0"
-        self.request.form["b_size"] = "1000000"
-        self.request.__annotations__.pop("plone.memoize")
+        aliases = []
+        for redirect in redirects:
```
davisagli: I would write this as 
```suggestion
        for path, info in redirects:
```

That way Python destructures the items for you so you don't have to do `info = redirect[1]` and `redirect[0]` below. It's just a bit easier to read.

davisagli:
@jnptk There's also a failing test test_querystringsearch_fullobjects which is surprising because you haven't touched that endpoint. See if you can find out in the debugger what's happening there.

src/plone/restapi/services/aliases/get.py
jnptk: I have no clue why it even worked like this, anyways, fixed it

src/plone/restapi/tests/http-examples/aliases_non_root_filter.resp
```diff
@@ -0,0 +1,15 @@
+HTTP/1.1 200 OK
+Content-Type: application/json
+
+{
+    "@id": "http://localhost:55001/plone/front-page/@aliases",
+    "items": [
+        {
+            "datetime": "2022-05-05T00:00:00",
```
davisagli, line 8: We have a test in CI which checks to make sure none of the .resp files have changed unexpectedly. So we have to make sure they don't include times that change each time the test is run. I took care of this in https://github.com/plone/plone.restapi/pull/1966/commits/587a2f34909ff57bb58b8890a87a32c05c10b79d

davisagli (commented):
Really close now. I'd still like to have tests demonstrating that the `manual`, `start`, and `end` parameters are parsed and executed as expected, to make sure we don't accidentally break them in the future.

jnptk:

davisagli (approved):
@jnptk Thank you!

PR #1969, opened 2025-12-17
fixing missing parameter factory error

Shyam-Raghuwanshi:
How to reproduce

go to - https://demo.plone.org/controlpanel/dexterity-types/Document/schema
Try adding any new field using - Add field
And fill required field for text or full name
Try to save

- [X] I signed and returned the [Plone Contributor Agreement](https://plone.org/foundation/contributors-agreement), and received and accepted an invitation to join a team in the Plone GitHub organization.
- [X] I verified there aren't any other open pull requests for the same change.
- [X] I followed the guidelines in [Contributing to Plone](https://6.docs.plone.org/contributing/index.html).
- [X] I successfully ran code quality checks on my changes locally.
- [X] I successfully ran tests on my changes locally.
- [X] If needed, I added new tests for my changes.
- [X] If needed, I added [documentation](https://6.docs.plone.org/contributing/documentation/index.html) for my changes.
- [X] I included a [change log entry](https://6.docs.plone.org/contributing/index.html#contributing-change-log-label) in my commits.

----
📚 Documentation preview 📚: https://plonerestapi--1969.org.readthedocs.build/

### DIFF
--- a/news/1969.bugfix
+++ b/news/1969.bugfix
@@ -1,0 +1,1 @@
+Fixed field creation API error. @Shyam-Raghuwanshi


davisagli (changes requested):
This is not the right way to fix this. If the client requests a field factory that doesn't exist, it's the job of the API to throw an error, not magically turn it into something else.

Shyam-Raghuwanshi:
> This is not the right way to fix this. If the client requests a field factory that doesn't exist, it's the job of the API to throw an error, not magically turn it into something else.

Hi, you’re absolutely right. I wanted to clarify one point and get your thoughts on it.

The client has requested adding a field that already exists on the frontend. For example, if we take the Full Name field: if we remove it from the frontend, then there’s nothing additional we need to handle here. However, if the Full Name field is selected and we send a value for a field that doesn’t exist, it may result in an error, which could be confusing from a user perspective.

I wanted to check with you whether this understanding is correct and how you’d recommend handling this case.

davisagli:
@Shyam-Raghuwanshi It's a bug that Full Name is shown as a choice in the frontend here. It's meant for when the SchemaWidget is used to set up a volto-form-block form, but not when it is used to set up a content type schema. I asked @robgietema to take a look at it.

Shyam-Raghuwanshi:
> @Shyam-Raghuwanshi It's a bug that Full Name is shown as a choice in the frontend here. It's meant for when the SchemaWidget is used to set up a volto-form-block form, but not when it is used to set up a content type schema. I asked @robgietema to take a look at it.

ok, we will proceed with this after the reply

davisagli:
This needs to be fixed in Volto:
1. the types control panel needs to pass a filterFactory to SchemaWidget so that it only allows field types that make sense for a content type
2. the SchemaWidget needs to send the factory to the backend as a string, not an object

PR #1971, opened 2026-01-12
Add the ability to import/export users via CSV

jnptk:
- [x] As site administrator I can create Plone users by uploading a CSV file
- [x] As site administrator I can export all existing Plone users as CSV file
- [x] Update documentation
- [x] More tests!!

----
📚 Documentation preview 📚: https://plonerestapi--1971.org.readthedocs.build/


src/plone/restapi/services/users/get.py
```diff
@@ -75,22 +85,19 @@ def isDefaultPortrait(value):
     )
 
 
-@implementer(IPublishTraverse)
-class UsersGet(Service):
-    def __init__(self, context, request):
-        super().__init__(context, request)
-        self.params = []
+@implementer(IExpandableElement)
```
davisagli: @jnptk This is your problem causing the tests to break.

Implementing this interface causes this adapter to be included when building all the expandable elements (that is, the `@components` section when serializing content) here: https://github.com/plone/plone.restapi/blob/main/src/plone/restapi/serializer/expansion.py#L17

Then it's called with only context and request but not params, so it breaks:
```
  File "/Users/davisagli/.local/share/uv/python/cpython-3.12-macos-aarch64-none/lib/python3.12/unittest/case.py", line 703, in debug
    self._callTestMethod(testMethod)
  File "/Users/davisagli/.local/share/uv/python/cpython-3.12-macos-aarch64-none/lib/python3.12/unittest/case.py", line 589, in _callTestMethod
    if method() is not None:
  File "/Users/davisagli/Plone/buildout.coredev/src/plone.restapi/src/plone/restapi/tests/test_dxcontent_deserializer.py", line 269, in test_serialize2deserialize_relation
    serialization_value = self.serialize("test_relationchoice_field")
  File "/Users/davisagli/Plone/buildout.coredev/src/plone.restapi/src/plone/restapi/tests/test_dxcontent_deserializer.py", line 263, in serialize
    return serializer()[field]
  File "/Users/davisagli/Plone/buildout.coredev/src/plone.restapi/src/plone/restapi/serializer/dxcontent.py", line 128, in __call__
    result.update(expandable_elements(self.context, self.request))
  File "/Users/davisagli/Plone/buildout.coredev/src/plone.restapi/src/plone/restapi/serializer/expansion.py", line 19, in expandable_elements
    for element in elements:
  File "/Users/davisagli/Plone/buildout.coredev/eggs/v5/zope.interface-8.2-py3.12-macosx-15.6-arm64.egg/zope/interface/registry.py", line 386, in getAdapters
    adapter = factory(*objects)
TypeError: Users.__init__() missing 1 required positional argument: 'params'
```

Why did you make it expandable? We don't want to return users with every content response, but only when the @users endpoint is requested explicitly.

jnptk: Aha, I see

> Why did you make it expandable?

Honestly I just like how it's done in the [at]aliases endpoint and wanted to adopt that

jnptk:
**DO NOT MERGE YET**

Notes:

- `test_list_users_via_csv` seems to be flaky
- roles don't seem to be applied
- location response header when multiple user are created?

docs/source/endpoints/users.md
```diff
@@ -56,6 +56,20 @@ The server will return a {term}`401 Unauthorized` status code.
 :language: http
 ```
 
+### List all Users via CSV
```
stevepiercy:
```suggestion
### List all users via CSV
```

docs/source/endpoints/users.md
```diff
@@ -56,6 +56,20 @@ The server will return a {term}`401 Unauthorized` status code.
...
+
+To download all users of a Plone site as a CSV file, send a `GET` request to the `/@users` endpoint from site root.
```
stevepiercy:
```suggestion
To download all users of a Plone site as a CSV file, send a `GET` request to the `/@users` endpoint from the site root.
```

stevepiercy (changes requested):
Minor docs tweaks. Looking good so far. Thank you!

stevepiercy (commented):
Docs LGTM so far.

stevepiercy (approved):
Additional new docs look good!

https://plonerestapi--1971.org.readthedocs.build/en/1971/endpoints/users.html#list-all-users-via-csv

Needs a technical review.

jnptk:

docs/source/endpoints/users.md
```diff
@@ -147,6 +147,47 @@ The `Location` header contains the URL of the newly created user, and the resour
...
+| Column          | Type   | Example                         |
+|-----------------|--------|---------------------------------|
+| `id` (required) | string | jdoe                            |
+| `username`      | string | jdoe                            |
+| `fullname`      | string | John Doe                        |
+| `description`   | string | Software Developer from Berlin. |
+| `email`         | string | jdoe@example.com                |
+| `roles`         | list   | "Member, Contributor"           |
+| `groups`        | list   | AuthenticatedUsers              |
+| `location`      | string | Berlin, DE                      |
```
stevepiercy: I think this needs to be quoted because it's a CSV and contains a comma. What do you think?

```suggestion
| `location`      | string | "Berlin, DE"                    |
```

docs/source/endpoints/users.md
```diff
@@ -147,6 +147,47 @@ The `Location` header contains the URL of the newly created user, and the resour
...
+| `home_page`     | string | jdoe.dev                        |
+| `password`      | string | pwd1234                         |
+
+```{note}
+If you want a user to have more that one role, note that you have to put the roles in quotes (see table above).
```
stevepiercy:
```suggestion
When a user has more than one role, put the roles in quotes, as shown in the table above.
Additionally, values that contain commas should be placed in quotes.
```

docs/source/endpoints/users.md
```diff
@@ -147,6 +147,47 @@ The `Location` header contains the URL of the newly created user, and the resour
...
+```
+
+Example of a minimal CSV file:
+
+```
+id,fullname,description,email
+jdoe,John Doe,Software Developer from Berlin,jdoe@example.com
```
stevepiercy: I'd suggest expanding this to display a quoted example and change it from "minimal" to a "CSV file with quoted values."

docs/source/endpoints/users.md
```diff
@@ -147,6 +147,47 @@ The `Location` header contains the URL of the newly created user, and the resour
...
+```
+
+Response:
+
+```{literalinclude} ../../../src/plone/restapi/tests/http-examples/users_add_csv_format.resp
+:language: http
+```
```
stevepiercy: The response should immediately follow the request.

stevepiercy (changes requested):
This looks really good. Thank you! Please see my suggestions.

https://plonerestapi--1971.org.readthedocs.build/en/1971/endpoints/users.html#create-users-via-csv

jnptk:
While writing the docs, I noticed that the roles included in the [request](https://github.com/plone/plone.restapi/pull/1971/changes#diff-700c0129d4ee936cf35aca7178b9b79af69010d5747f68be9678804d13c41926) are not part of the [response](https://github.com/plone/plone.restapi/pull/1971/changes#diff-8076f2854871ec5fde5a537ecc8bf4996c6117e875610e315dad81d7bf919bec), I will investigate...

stevepiercy (approved):
Docs changes look great! I'm glad you caught the empty `roles` response. I missed that.

jnptk:
@davisagli we set the `Location` header with the location of the newly created user, but in the case of a csv one might create more than one user. Should we set the header in this case or abandon it?

stevepiercy (approved):
Roles now appear in the response.

https://plonerestapi--1971.org.readthedocs.build/en/1971/endpoints/users.html#create-users-via-csv

Thank you!

docs/source/endpoints/users.md
```diff
@@ -133,6 +147,48 @@ The `Location` header contains the URL of the newly created user, and the resour
 
 If no roles have been specified, then a `Member` role is added as a sensible default.
 
+### Create users via CSV
+
+To create a new user from a CSV file, send a `POST` request to the `/@users` endpoint with a `multipart/form-data` body containing a CSV file with the user details.
```
davisagli: This makes it sound like this can create only one user.

docs/source/endpoints/users.md
```diff
@@ -133,6 +147,48 @@ The `Location` header contains the URL of the newly created user, and the resour
...
+The endpoint expects the CSV file to be under a "file" part of the `Content-Disposition: multipart/form-data`:
```
davisagli: I wasn't really sure what that meant without looking at the example. Maybe it would be clearer to say:
```suggestion
The endpoint expects a request body with `Content-Type: multipart/form-data` including a file upload named `file`.
```

src/plone/restapi/tests/http-examples/users_searched.resp
```diff
@@ -1,34 +1,38 @@
...
-            "items": [
-                {
-                    "id": "AuthenticatedUsers",
-                    "title": "AuthenticatedUsers"
-                },
-                {
-                    "id": "Reviewers",
-                    "title": "Reviewers"
-                }
+{
```
davisagli, line 4: I see the format changed from a list of users to an object with `@id`, `items`, and `items_total`. That's more consistent with what we try to do in the REST API, but unfortunately it's a backwards-incompatible change that will break existing scripts. We can only change this in a major release of plone.restapi.

src/plone/restapi/tests/test_services_users.py
```diff
@@ -1323,41 +1348,45 @@ def test_user_with_datetime(self):
...
+        # Inject only what _get_users actually needs
+        users.acl_users = MockAclUsers()
+        users.portal_membership = MockPortalMembership()
+
+        result = users._get_users(foo="bar")
+
+        # Sorted by normalized fullname; None users filtered out
+        self.assertEqual(len(result), 2)
+        self.assertEqual(result[0].userid, "user1")
+        self.assertEqual(result[1].userid, "user2")
```
davisagli: To be honest, I would rather have this test rewritten to test with real users rather than mock objects.

src/plone/restapi/batching.py
```diff
@@ -16,8 +15,8 @@ def __init__(self, request, results):
 
         try:
             data = json_body(request)
-        except DeserializationError as e:
-            raise BadRequest(e)
+        except DeserializationError:
+            data = {}
```
davisagli, line 19: We should mention this change as a separate bugfix release note

src/plone/restapi/services/users/get.py
```diff
@@ -208,10 +213,11 @@ def reply(self):
...
                         )
                         result.append(serializer())
-                    return result
+                    return result, len(result)
                 else:
-                    self.request.response.setStatus(401)
-                    return
+                    raise Unauthorized(
+                        "You are not authorized to access this resource."
+                    )
```
davisagli, line 234: Good idea to switch to raising exceptions for the error responses.

src/plone/restapi/services/users/get.py
```diff
@@ -239,13 +244,96 @@ def reply(self):
...
+            raise Unauthorized("You are not authorized to access this resource.")
+
+    def reply_root_csv(self):
+        if len(self.params) > 0:
+            raise BadRequest("You may not request a CSV reply for a specific user.")
+
+        if self.has_permission_to_enumerate():
+            file_descriptor, file_path = tempfile.mkstemp(
+                suffix=".csv", prefix="users_"
+            )
```
davisagli: Instead of creating a temporary file on disk, I would write the file in memory using io.BytesIO. Should be something like this:
```
buffer = io.BytesIO()
text_wrapper = io.TextIOWrapper(buffer, encoding='utf-8', newline='')
csv_writer = writer(text_wrapper)
# write the CSV normally
content = buffer.getvalue()
```

src/plone/restapi/services/users/get.py
```diff
@@ -239,13 +244,96 @@ def reply(self):
...
-                return
+                raise NotFound(f"User ${self._get_user_id} does not exist.")
             serializer = queryMultiAdapter((user, self.request), ISerializeToJson)
             return serializer()
         else:
-            self.request.response.setStatus(401)
-            return
+            raise Unauthorized("You are not authorized to access this resource.")
+
+    def reply_root_csv(self):
```
davisagli: I'm not sure why you have root in the name here. It's the same list of users as the JSON response. It's not context-dependent.

src/plone/restapi/services/users/get.py
```diff
@@ -75,22 +82,17 @@ def isDefaultPortrait(value):
     )
 
 
-@implementer(IPublishTraverse)
-class UsersGet(Service):
-    def __init__(self, context, request):
-        super().__init__(context, request)
-        self.params = []
+class Users:
```
davisagli: Since this is _not_ an expander, I don't think there's much reason to have a Users class that is separate from the UsersService (especially since it still directly handles I/O like reading from the request and serializing the results in the right format).

I'm okay with it if there's some remaining reason, like making it easier to test.

src/plone/restapi/services/users/add.py
```diff
@@ -131,12 +137,45 @@ def reply(self):
...
-        registration = getToolByName(self.context, "portal_registration")
+        if form := self.request.form:
+            if not form.get("file"):
+                raise BadRequest("No file uploaded")
+
+            file = form["file"]
+            if file.headers.get("Content-Type") not in ("text/csv", "application/csv"):
+                raise BadRequest("Uploaded file is not a valid CSV file")
+            if len(self.params) > 0:
+                raise NotAcceptable(_(""))
```
davisagli: 406 Not Acceptable is specifically supposed to be for when the client makes a request with an `Accept` header that the server doesn't know how to provide. I think BadRequest or NotFound is more appropriate for this error when the service is called with an unexpected path element.

src/plone/restapi/services/users/add.py
```diff
@@ -131,12 +137,45 @@ def reply(self):
...
+                encoding="utf-8",
+                newline="",
+            )
+            try:
+                reader = csv.DictReader(stream)
+                for row in reader:
+                    # convert to lists
+                    for key in ("roles", "groups"):
+                        if row.get(key):
+                            row[key] = [r.strip() for r in row[key].split(",")]
```
davisagli, line 161: You could do:
```suggestion
                            row[key] = [r.strip() for r in row[key].split(",") if r.strip()]
```
and then you don't have to remove empty values afterward

src/plone/restapi/services/users/add.py
```diff
@@ -152,11 +191,7 @@ def reply(self):
 
         # Add a portal member
         if not self.can_add_member:
-            return self._error(
-                403,
-                "Forbidden",
-                _("You need AddPortalMember permission."),
-            )
+            raise Forbidden(_("You need AddPortalMember permission."))
```
davisagli, line 193: Does this still get translated okay? The `_error` function was explicitly handling the translation, and I'm not sure if the default serialization of an exception in plone.rest does it. (If it doesn't, that would be a nice improvement.)

davisagli (changes requested):
> we set the Location header with the location of the newly created user, but in the case of a csv one might create more than one user. Should we set the header in this case or abandon it?

No, I don't think it makes sense to redirect to one particular user if multiple users were created.

docs/source/endpoints/users.md
jnptk:
```suggestion
To create new users from a CSV file, send a `POST` request to the `/@users` endpoint.
```
@davisagli ?

src/plone/restapi/services/users/add.py
jnptk: This is only for roles and groups, keys with empty values are removed below

src/plone/restapi/services/users/add.py
jnptk: For reference: https://github.com/plone/plone.rest/pull/201

src/plone/restapi/services/users/get.py
jnptk: done! please take a look

src/plone/restapi/tests/http-examples/users_searched.resp
jnptk: So should I undo the format change or do we just wait until the next major release?

davisagli: @jnptk Now I remember that we currently are on 10.0.0a1, so we can make other breaking changes in a new alpha release. Make sure we add release note with the .breaking extension for this change, separate from the other new feature.

src/plone/restapi/tests/http-examples/users_add_csv_format.resp
```diff
@@ -0,0 +1,78 @@
+HTTP/1.1 201 Created
+Content-Type: application/json
+
+[
```
davisagli: Can we make this consistent with the GET response? (a dict with `items`)?

davisagli (commented):
I think this is just about ready!

src/plone/restapi/tests/http-examples/users_searched.resp
davisagli: @jnptk Hmm, but this change will break everything in Volto that uses the `@users` endpoint to list users.

We have two options:
1. Go ahead with the change in plone.restapi -- but then we have to first make changes in Volto (at least Volto 19) to handle both the old and new format, before we can merge this.
2. Change this PR back to returning the old format with just a list.

jnptk: @davisagli https://github.com/plone/volto/pull/7895

jnptk:

src/plone/restapi/tests/http-examples/users_searched.resp
jnptk: > @davisagli [plone/volto#7895](https://github.com/plone/volto/pull/7895)

Merged.

davisagli:

davisagli (approved):

src/plone/restapi/services/users/get.py
```diff
@@ -193,6 +205,10 @@ def has_permission_to_access_user_info(self):
         )
 
     def reply(self):
+        result = {"@id": f"{self.context.absolute_url()}/@users"}
+        self.request.response.setStatus(200)
+        self.request.response.setHeader("Content-Type", "application/json")
+
```
erral, line 211: We had a project with the `@users` endpoint overriden changing the `reply` method.

After upgrading to Plone 6.2 (plone.restapi 10), we see that the endpoint is not returning the results as `application/json`.

In these changes I see that we introduced the `render` method which is called before the `reply` and this has some kind of effect on this.

I just wanted to leave a note here, as I don't know whether this can be considered a bug...

PR #1974, opened 2026-01-23
Return 401 when user lacks permission for control panel access

### LINKED ISSUE #1949: Error fetching data from a control panel that the user does not have permission in form
**Describe the bug**

I want a user with the Site Administrator role to manage users, but not have access to the `security` control panel (http://localhost:3000/controlpanel/security). This control panel requires `Plone Site Setup: Security` permission. So I granted this permission only to Manager. But in Volto, the user management screen needs access to data from this control panel. So when accessing the user management screen with a Site Administrator user, the url http://localhost:3000/@controlpanels/security returns the error:

```json
{ 
"message": "'SecurityControlpanel' object has no attribute 'title'", 
"type": "AttributeError"
}
```

In the Plone log we have the error:

```python
2025-10-01 15:30:30 ERROR [Zope.SiteErrorLog:35][waitress-0] AttributeError: http://localhost:3000/@controlpanels/security
Traceback (innermost last): 
Module ZPublisher.WSGIPublisher, line 181, in transaction_pubevents 
Module ZPublisher.WSGIPublisher, line 390, in publish_module 
Module ZPublisher.WSGIPublisher, line 284, in publish 
Module ZPublisher.mapply, line 98, in mapply 
Module ZPublisher.WSGIPublisher, line 68, in call_object 
Module plone.rest.service, line 21, in __call__ 
Module plone.restapi.services.controlpanels.get, line 50, in reply 
Module plone.restapi.services.controlpanels.get, line 73, in reply_panel 
Module plone.restapi.serializer.controlpanels, line 87, in __call__
AttributeError: 'SecurityControlpanel' object has no attribute 'title'
```
**To Reproduce**

1. In the Plone Site permissions management screen (http://localhost:8080/Plone/manage_access) set the `Plone Site Setup: Security` permission only for Manager.
2. Access the user management screen (http://localhost:3000/controlpanel/users) with a Site Administrator user.

**Expected behavior**

The url http://localhost:3000/@controlpanels/security should not return an error.

**Software (please complete the following information):**

- Browser chrome
- Plone Version 6.1.2
- Volto Version 18.27.2
- Plone REST API Version 9.15.1

**Additional context**

This error does not occur in Plone Classic.

The error occurs here:

https://github.com/plone/plone.restapi/blob/22574d574fe5e44f9f76059f861a26a62af855b7/src/plone/restapi/serializer/controlpanels/__init__.py#L87

`self.controlpanel` becomes `None`, because of the lack of permission. Maybe in this situation, we should set `title` and `group` to an empty string?

Dev10-sys:
Fixes #1949

Return a 401 Unauthorized response when the current user does not have
permission to access a control panel, instead of raising an AttributeError
during serialization.

----
📚 Documentation preview 📚: https://plonerestapi--1974.org.readthedocs.build/

### DIFF
--- a/src/plone/restapi/services/controlpanels/get.py
+++ b/src/plone/restapi/services/controlpanels/get.py
@@ -9,6 +9,9 @@ from zope.component import getAdapters
 from zope.interface import alsoProvides
 from zope.interface import implementer
 from zope.publisher.interfaces import IPublishTraverse
+from zope.security import checkPermission
+from zExceptions import Unauthorized
+
 
 
 @implementer(IPublishTraverse)
@@ -65,6 +68,9 @@ class ControlpanelsGet(Service):
             self.request.response.setStatus(404)
             return
 
+        # 🔐 Permission check (FIX for #1949)
+        if not checkPermission("cmf.ManagePortal", panel):
+            raise Unauthorized()
         # Panel child request
         if len(self.params) > 1:
             return IJsonCompatible(panel.get(self.params[1:]))


davisagli:
I suggested a different solution for #1949. And it's not always correct to check the Manage portal permission for access to control panels. In the classic UI each Plone control panel has a different permission so that management can be delegated to specific roles. We should try to use those same permissions here. And we should not only raise an exception when a specific control panel is requested, but also leave them out of the listing if the current user doesn't have permission.

PR #1979, opened 2026-02-07
Fix relatedItems [null] for deleted relations

### LINKED ISSUE #965: null is returned for RelationList fields if the reference is gone
Steps to reproduce:

1. Create an item, then create a second item
2. Add the first item as a related item for the second one
3. Delete the first item
4. If you edit the second item from Plone, the related items field is empty
5. If you ask for the second item via restapi (with fullobjects), its `relatedItems` field will be `[null]` instead of `[]`

I would expect the restapi to skip null items and return `[]`. What do you think?

aryan7081:
Fixes #965 

When a related item was deleted, the API returned relatedItems: [null] instead of [] because broken relations were still included. The fix filters by `to_object` so only resolvable relations are serialized and the API returns an empty list.

----
📚 Documentation preview 📚: https://plonerestapi--1979.org.readthedocs.build/

### DIFF
--- a/news/965.bugfix
+++ b/news/965.bugfix
@@ -1,1 +1,1 @@
-Return empty list for relatedItems when related content was deleted instead of [null]. Filter broken relations in RelationListFieldSerializer.get_value() by to_object so only resolvable relations are serialized. @aryan7081
+When serializing RelationList fields, don't include a null for broken relations. @aryan7081


aryan7081:

news/965.bugfix
```diff
@@ -0,0 +1 @@
+Return empty list for relatedItems when related content was deleted instead of [null]. Filter broken relations in RelationListFieldSerializer.get_value() by to_object so only resolvable relations are serialized. @aryan7081
```
davisagli:
```suggestion
When serializing RelationList fields, don't include a null for broken relations. @aryan7081
```

davisagli (approved):
Thank you!

PR #1987, opened 2026-02-16
Respect Exclude from navigation in search and contextnavigation

### LINKED ISSUE #1986: @contextnavigation returns items that are excluded from navigation.
**Describe the bug**

The `@contextnavigation` endpoint include content items that have Exclude from navigation set `exclude_from_nav = True`.

**To Reproduce**

1. Create a folder `for e.g test-folder`..
2. Inside test-folder create two documentes.
      - Document A => Exclude from navigation unchecked.
      - Document B => Exclude from navigation checked.
4. Save the item.
5. call the `@contextnavigation` endpoint: ` GET http://localhost:8080/Plone/test-folder/@contextnavigation`
6. Bug: Both Document A and Document B appear in the items array. Document B should not appear because it's excluded from navigation.

**Expected behavior**

Items with `exclude_from_nav=True` should not appear in the `@contextnavigation` response.

**Additional context**

Related Volto issue: https://github.com/plone/volto/issues/1340

aryan7081:
Fixes #1986

Content marked with `Exclude from navigation` still appeared in search and context navigation because the `@search handler`, catalogue serializer, `@contextnavigation`, and classic search did not filter for the `Show excluded items` setting. The fix adds `exclude_from_nav` filtering in all of these (including a patch to CMFPlone’s Search view for classic UI), so excluded items are no longer returned when that setting is off.

----
📚 Documentation preview 📚: https://plonerestapi--1987.org.readthedocs.build/

### DIFF
--- a/news/1986.bugfix
+++ b/news/1986.bugfix
@@ -1,1 +1,1 @@
-Fix content with "Exclude from navigation" still appearing in search (API and classic UI) and in context navigation; both now honor the "Show excluded items" navigation setting. @aryan7081
+Fix content with "Exclude from navigation" still appearing in search (API and Classic UI) and in context navigation. Both now honor the "Show excluded items" navigation setting. @aryan7081


aryan7081:

news/1986.bugfix
```diff
@@ -0,0 +1 @@
+Fix content with "Exclude from navigation" still appearing in search (API and classic UI) and in context navigation; both now honor the "Show excluded items" navigation setting. @aryan7081
```
stevepiercy:
```suggestion
Fix content with "Exclude from navigation" still appearing in search (API and Classic UI) and in context navigation. Both now honor the "Show excluded items" navigation setting. @aryan7081
```

davisagli (changes requested):
The exclude_from_nav setting is intended to exclude an item from navigation menus, but not from search or other listings. So the only part of this PR that is valid is the fix for the `contextnavigation` service.

aryan7081:
@davisagli Thanks for clarifying this. After reading this issue https://github.com/plone/volto/issues/1340, I thought Items with `exclude_from_nav = True` should also be excluded from the search results. `@contextnavigation` endpoint is working as expected, so I think it makes sense to close this PR and issue #1986. 

Before closing this, I just need your confirmation. Do you also think the same??

aryan7081:
Closing via https://github.com/plone/plone.restapi/issues/1986

PR #1996, opened 2026-03-16
Improve usage of boolean URL parameters

### LINKED ISSUE #1960: Inconsistent usage of boolean URL parameters (true, false, yes, no, True, False, 1, 0)

cekk:
- [x] I signed and returned the [Plone Contributor Agreement](https://plone.org/foundation/contributors-agreement), and received and accepted an invitation to join a team in the Plone GitHub organization.
- [] If needed, I added [documentation](https://6.docs.plone.org/contributing/documentation/index.html) for my changes.
- [x] I included a [change log entry](https://6.docs.plone.org/contributing/index.html#contributing-change-log-label) in my commits.

-----

Closes #1960

Now boolean checks are made in one single method/way.
I also added a strict flag to raise exceptions if given value is not boolean-ish

----
📚 Documentation preview 📚: https://plonerestapi--1996.org.readthedocs.build/

### DIFF
--- a/docs/source/upgrade-guide.md
+++ b/docs/source/upgrade-guide.md
@@ -28,7 +28,7 @@ See {doc}`../usage/parameters` for details.
 
 ## Upgrading to `plone.restapi` 9.x
 
-`plone.restapi` 9.x dropped support for Python 3.7, which reached its `end-of-life <https://devguide.python.org/versions/>`\_ in July 2023.
+`plone.restapi` 9.x dropped support for Python 3.7, which reached its [end of life](https://devguide.python.org/versions/) in July 2023.
 
 Upgrading to `plone.restapi` 9.x might require minor changes to your code base that are described in this section. If you are using a recent version of Volto, there are no changes necessary.
 
@@ -57,6 +57,7 @@ In `plone.restapi` 9.0.0, the following response would be returned with a `breac
 
 Pull Request: https://github.com/plone/plone.restapi/pull/1636
 
+
 ### Remove deprecated `@unlock`, `@refresh-lock` endpoints
 
 The deprecated `@unlock` and `@refresh-unlock` endpoints were removed in `plone.restapi` 9.
@@ -68,6 +69,7 @@ Send a `PATCH` request to the `@lock` endpoint to refresh a lock (replaces the `
 See the documentation of the `@lock` endpoint for more information: https://6.docs.plone.org/plone.restapi/docs/source/endpoints/locking.html.
 
 Pull Request: https://github.com/plone/plone.restapi/pull/1235
+
 
 ### Remove `plone.tiles` and the `@tiles` endpoint
 
--- a/docs/source/usage/parameters.md
+++ b/docs/source/usage/parameters.md
@@ -15,7 +15,7 @@ This chapter describes the conventions for URL parameters accepted by the Plone 
 
 (boolean-parameters)=
 
-## Boolean Parameters
+## Boolean parameters
 
 ```{versionchanged} plone.restapi 10.0.0
 Since plone.restapi version 10.0.0, boolean parameters are validated more strictly.


cekk:

src/plone/restapi/utils.py
```diff
@@ -1,34 +1,6 @@
 from zope.component.hooks import getSite
 
 
-def is_truthy(value) -> bool:
```
mamico, line 4: @cekk these two utilities might still be used, and given the minimal overhead of implementing them via the new function, I suggest keeping them.

cekk: they were introduced 3 months ago only in one endpoint: https://github.com/plone/plone.restapi/pull/1966

src/plone/restapi/deserializer/__init__.py
```diff
@@ -17,18 +17,50 @@ def json_body(request):
     return data
 
 
-def boolean_value(value):
+def boolean_value(value, strict=False):
```
davisagli: I think we should move this to utils.py, then import it here.

davisagli: Is there a good reason to NOT make it strict by default? It's not technically backward-compatible, but we are currently releasing alphas for Plone 6.2, so we can make that change if we want.

cekk: i kept it optional just for backward compatibility but i like the idea of set as default too.

If we switch to True we should handle exceptions where we use this method? Or pass strict=False

tisto: Let's not be afraid of bumping a major version. This is what semantic versioning is about. I'd rather have a breaking change release than a solution that is more complex.

mauritsvanrees: The latest release of `plone.restapi` is 10.0.0a1, so we are in alpha state, and no major version bump is needed.

tisto: @mauritsvanrees even better. What better time than now to break things and move on! ;)

src/plone/restapi/deserializer/__init__.py
```diff
@@ -45,3 +35,69 @@ def parse_int(data, prop, default):
...
+    if is_falsy(value):
+        return False
+    if default is not None:
+        logger.warning(
+            "Could not parse value %r as boolean, returning default %r",
+            value,
+            default,
+        )
+        return default
+    raise ValueError(f"Could not parse value {value!r} as boolean")
```
davisagli: Please move these from `deserializer/__init__.py` to `bbb.py` (keep the import of boolean_value here for backwards-compat)

src/plone/restapi/bbb.py
```diff
@@ -18,6 +18,7 @@
     from plone.base.interfaces import ITestCasePloneSiteRoot
     from plone.base.navigationroot import get_navigation_root
     from plone.base.utils import base_hasattr
+    from plone.base.utils import boolean_value
```
davisagli: Careful! Doesn't this cause the entire block of plone.base imports to fail when boolean_value isn't there yet?

I would add a separate try/catch for this import, and put the copied functions here inside the catch block.

news/1996.breaking
davisagli, line 1: Rename to 1996.breaking

news/1996.feature
```diff
@@ -0,0 +1 @@
+Improve usage of boolean URL parameters with boolean_value method (backward compatibility for plone.base). @cekk
```
davisagli:
```suggestion
Services which take boolean parameters now check the input more strictly, using the `boolean_value` util. @cekk
```

src/plone/restapi/deserializer/__init__.py
cekk: i think that this can be marked as resolved

src/plone/restapi/bbb.py
cekk: sorry, you're right..code moved

src/plone/restapi/bbb.py
```diff
@@ -47,3 +47,74 @@
...
+        if is_falsy(value):
+            return False
+        if default is not None:
+            logger.warning(
+                "Could not parse value %r as boolean, returning default %r",
+                value,
+                default,
+            )
+            return default
+        raise ValueError(f"Could not parse value {value!r} as boolean")
```
davisagli, line 117: We should resolve the feedback on https://github.com/plone/plone.base/pull/112 and then copy any changes here.

cekk:
@davisagli i added boolean_value validation also on querystring-search (i missed it) and updated docs as you suggested.

docs/source/endpoints/aliases.md
```diff
@@ -149,6 +149,38 @@ All of the following parameters are optional.
 
 To search for specific aliases, send a `GET` request to the `@aliases` endpoint with one or more of the above named parameters as shown in the following example.
 
+```{warning}
```
stevepiercy: This gets pulled into the main docs, so we need to use the package name. This also uses the correct admonition.

````suggestion
```{versionchanged} plone.restapi 10.0.0
````

File: docs/source/endpoints/aliases.md [hunk 8]
```diff
@@ -149,6 +149,38 @@ All of the following parameters are optional.
...
+From version 10.0.0 boolean parameters validation is more strictly.
```
stevepiercy:
```suggestion
Since plone.restapi version 10.0.0, boolean parameters are validated more strictly.
```

docs/source/endpoints/aliases.md
```diff
@@ -149,6 +149,38 @@ All of the following parameters are optional.
...
+
+Valid `True` values example:
```
stevepiercy:
```suggestion
Values that evaluate as `True`, where quoted values are strings, and unquoted values are either boolean or integer.
```

docs/source/endpoints/aliases.md
```diff
@@ -149,6 +149,38 @@ All of the following parameters are optional.
...
+- "yes"
+- "t"
+- "true"
+- True
+- "True"
+- "active"
+- "enabled"
+- "on"
+
+Valid `False` values example:
```
stevepiercy:
```suggestion
Values that evaluate as `False`, where quoted values are strings, and unquoted values are either boolean or integer.
```

docs/source/endpoints/aliases.md
```diff
@@ -149,6 +149,38 @@ All of the following parameters are optional.
...
+- "no"
+- "f"
+- "false"
+- False
+- "False"
+- "inactive"
+- "disabled"
+- "off"
+```
+
```
stevepiercy, line 157: After accepting the foregoing changes, please extract it into an include file at `docs/source/_inc/_true-false.inc`, and include that file here and elsewhere to reduce maintenance. Pay attention to the relative path if nesting changes.

````suggestion

```{include} ../_inc/_true-false.md
```

````

File: docs/source/endpoints/querystringsearch.md [hunk 12]
```diff
@@ -49,6 +49,38 @@ Parameters the endpoint will accept:
...
+- "n"
+- "no"
+- "f"
+- "false"
+- False
+- "False"
+- "inactive"
+- "disabled"
+- "off"
+```
```
stevepiercy:
````suggestion

```{include} ../_inc/_true-false.md
```

````

File: docs/source/endpoints/searching.md [hunk 13]
```diff
@@ -60,6 +60,38 @@ The `plone.restapi` `@search` endpoint will not do that for you.
...
+- "n"
+- "no"
+- "f"
+- "false"
+- False
+- "False"
+- "inactive"
+- "disabled"
+- "off"
+```
```
stevepiercy:
````suggestion

```{include} ../_inc/_true-false.md
```

````

File: docs/source/endpoints/querystringsearch.md (hunk 12 above)
Comment (stevepiercy) [reply docs/source/endpoints/querystringsearch.md]: Also I wonder if this should go under the next heading `## Parameters` and not here?

**Review by stevepiercy — CHANGES_REQUESTED** [review]
After making docs changes, verify these items.

- https://plonerestapi--1996.org.readthedocs.build/en/1996/endpoints/aliases.html#parameters
- https://plonerestapi--1996.org.readthedocs.build/en/1996/endpoints/querystringsearch.html
- https://plonerestapi--1996.org.readthedocs.build/en/1996/endpoints/searching.html

**Comment by stevepiercy** [comment]
@davisagli I like that better.

To clarify, instead of repeating the content, do you think this would be useful?

````
```{versionchanged} plone.restapi 10.0.0
Since plone.restapi version 10.0.0, boolean parameters are validated more strictly.

See {doc}`../usage/parameters` for details.
```
````

...and the modified content according to my other suggestions incorporated and moved to the new page?

**Comment by davisagli** [comment]
@stevepiercy Yeah, that's what I had in mind. I don't have a strong feeling for whether or not it's important to add that note in each place we have boolean params, but I don't mind if it's there.

**Comment by cekk** [comment]
@stevepiercy @davisagli i tried to update docs, let me know if now it's better

File: src/plone/restapi/deserializer/__init__.py [hunk 14]
```diff
@@ -17,20 +17,6 @@ def json_body(request):
     return data
 
 
-def boolean_value(value):
```
mauritsvanrees, line 20: If we remove the definition of `boolean_value` here, we should put a backwards compatibility import in place.  See the [deprecation guide](https://6.docs.plone.org/developer-guide/deprecation.html#deprecate-methods-and-properties).

src/plone/restapi/bbb.py
mauritsvanrees: The `plone.base` PR was approved and merged just now, so we can take over its code.

src/plone/restapi/deserializer/__init__.py
cekk: how? with deferred import in `__init__.py`?

```
zope.deferredimport.initialize()

zope.deferredimport.deprecated(
    "Import from plone.restapi.bbb instead",
    boolean_value="plone.restapi:bbb.boolean_value",
)
```

docs/source/usage/index.md
stevepiercy: Sorted. Looks like someone tacked on blocks recently without sorting.

```suggestion
blocks
content
customization
expansion
exploring
i18n
parameters
serialization
types-schema
```

news/1996.breaking
```diff
@@ -0,0 +1 @@
+Services which take boolean parameters now check the input more strictly, using the `boolean_value` util. @cekk
```
tisto, line 1: @cekk if this is a breaking change we need to mention it in our upgrade guide telling people what they need to do and what they need to check when upgrading to plone.restapi 10.

src/plone/restapi/deserializer/__init__.py
mauritsvanrees: Almost.  The last line should be: `boolean_value="plone.restapi.bbb:boolean_value"` as it imports `boolean_value` from the `plone.restapi.bbb` module.

The `zope.deferredimport.initialize()` seems not needed, but our deprecation guide currently has it, so let's do it that way.

docs/source/usage/parameters.md
```diff
@@ -0,0 +1,50 @@
...
+
+This chapter describes the conventions for URL parameters accepted by the Plone REST API.
+
+## Boolean Parameters
+
+```{versionchanged} plone.restapi 10.0.0
+Since plone.restapi version 10.0.0, boolean parameters are validated more strictly.
+```
+
+Values that evaluate as `True`, where quoted values are strings, and unquoted values are either boolean or integer.
```
davisagli: It doesn't make sense to mention non-string values here, because query string parameters are always strings before they are parsed by the application server. Also, this is not a complete sentence. I'll make an update.

davisagli: @stevepiercy How does this look? https://github.com/plone/plone.restapi/pull/1996/commits/e1a90b8bc5740d373853a5b6253c67acb01b1d62

stevepiercy: LGTM. Thank you!

news/1996.breaking
davisagli: @tisto Thanks, I forgot that we have an upgrade guide! Added in https://github.com/plone/plone.restapi/pull/1996/commits/9f63ab7d11d9e8e7dd59886f4b61d09a7a1b3d24

src/plone/restapi/bbb.py
davisagli: Done in https://github.com/plone/plone.restapi/pull/1996/commits/36208113016bb832ac1572d7d74033531b4017f9

src/plone/restapi/deserializer/__init__.py
davisagli: Done in https://github.com/plone/plone.restapi/pull/1996/commits/b9212f3896a9bcf4108b3ac22d28b87b36eb2d5b

The docs for zope.deferredimport.initialize say: "Modules do not need to call this directly, because the define* and deprecated* functions call it." So I didn't include it.

davisagli:

docs/source/usage/parameters.md
```diff
@@ -0,0 +1,48 @@
...
+
+(restapi-parameters)=
+
+# Parameters
+
+This chapter describes the conventions for URL parameters accepted by the Plone REST API.
+
+(boolean-parameters)=
+
+## Boolean Parameters
```
stevepiercy: Sentence case.

```suggestion
## Boolean parameters
```

docs/source/upgrade-guide.md
```diff
@@ -12,9 +12,23 @@ myst:
...
+### Boolean parameter parsing
+
+The REST API does stricter parsing of parameters that represent a boolean value.
+Make sure that you are sending one of the accepted values for boolean parameters.
+See {doc}`../usage/parameters` for details.
+
 ## Upgrading to `plone.restapi` 9.x
 
-`plone.restapi` 9.x dropped support for Python 3.7, which reached its `end-of-life <https://devguide.python.org/versions/>`_ in July 2023.
+`plone.restapi` 9.x dropped support for Python 3.7, which reached its `end-of-life <https://devguide.python.org/versions/>`\_ in July 2023.
```
stevepiercy: I've seen the `\` erroneously added to `.md` files by misconfigured editors. Make sure yours excludes `.md` files from being rewritten, as well as line breaks stripped. It was probably a good thing, anyway, because it caught this old reStructuredText formatted link.

```suggestion
`plone.restapi` 9.x dropped support for Python 3.7, which reached its [end of life](https://devguide.python.org/versions/) in July 2023.
```

davisagli: @stevepiercy Sorry, fixed in https://github.com/plone/plone.restapi/pull/1996/commits/79463736230d27b4427d6e976f7a5631b2a633ab

PR #2012, opened 2026-04-06
[WIP] Add content negotiation to support text/markdown

tisto:

-----

Closes #

----
📚 Documentation preview 📚: https://plonerestapi--2012.org.readthedocs.build/

### DIFF
--- a/src/plone/restapi/renderer/markdown.py
+++ b/src/plone/restapi/renderer/markdown.py
@@ -1,0 +1,345 @@
+"""Markdown Renderer - converts Python data structures to Markdown with YAML frontmatter."""
+
+from datetime import date
+from datetime import datetime
+from plone.restapi.interfaces import IRenderer
+from zope.component import adapter
+from zope.interface import implementer
+from zope.interface import Interface
+
+import json
+import re
+
+
+@implementer(IRenderer)
+@adapter(Interface)
+class MarkdownRenderer:
+    """Render data as GitHub Flavored Markdown with YAML frontmatter."""
+
+    content_type = "text/markdown"
+
+    # Fields to exclude from frontmatter
+    EXCLUDED_FROM_FRONTMATTER = {"blocks", "blocks_layout"}
+
+    # Fields that contain the main content body
+    BODY_FIELDS = {"text", "description"}
+
+    def __init__(self, request):
+        self.request = request
+
+    def __call__(self, data):
+        """Convert Python data structure to Markdown with YAML frontmatter."""
+        if not isinstance(data, dict):
+            # For non-dict data (lists, etc.), just return JSON representation
+            return json.dumps(data, indent=2)
+
+        # Extract metadata and body
+        frontmatter = {}
+        body_parts = []
+
+        for key, value in data.items():
+            if key in self.EXCLUDED_FROM_FRONTMATTER:
+                # Handle blocks separately
+                if key == "blocks":
+                    body_parts.extend(
+                        self._render_blocks(value, data.get("blocks_layout", {}))
+                    )
+            elif key in self.BODY_FIELDS:
+                # These go into the body, not frontmatter
+                if key == "text" and isinstance(value, dict):
+                    # RichText field structure
+                    body_parts.append(self._render_richtext(value))
+                elif value:
+                    body_parts.append(str(value))
+            else:
+                # Everything else goes to frontmatter
+                frontmatter[key] = value
+
+        # Build the document
+        parts = []
+
+        # Add YAML frontmatter
+        if frontmatter:
+            parts.append("---")
+            parts.append(self._render_yaml(frontmatter))
+            parts.append("---")
+            parts.append("")
+
+        # Add title as H1 if present
+        if "title" in data and data["title"]:
+            parts.append(f"# {data['title']}")
+            parts.append("")
+
+        # Add body content
+        if body_parts:
+            parts.extend(body_parts)
+
+        return "\n".join(parts)
+
+    def _render_yaml(self, data, indent=0):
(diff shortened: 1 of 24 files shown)


src/plone/restapi/renderer/markdown.py
```diff
@@ -0,0 +1,345 @@
...
+from zope.interface import Interface
+
+import json
+import re
+
+
+@implementer(IRenderer)
+@adapter(Interface)
+class MarkdownRenderer:
+    """Render data as GitHub Flavored Markdown with YAML frontmatter."""
```
davisagli, line 17: Unlike the JSONRenderer, this is really only designed to render a particular type of data (a content item in the format produced by an ISerializeToJson adapter). That makes me wonder if you're trying to generalize too much by using this in the base Service class instead of only the ContentGet service.

Do we need to return a markdown representation from other services as well?

src/plone/restapi/renderer/markdown.py
```diff
@@ -0,0 +1,345 @@
...
+            elif key in self.BODY_FIELDS:
+                # These go into the body, not frontmatter
+                if key == "text" and isinstance(value, dict):
+                    # RichText field structure
+                    body_parts.append(self._render_richtext(value))
+                elif value:
+                    body_parts.append(str(value))
+            else:
+                # Everything else goes to frontmatter
+                frontmatter[key] = value
```
davisagli, line 56: This would be more extensible if it was handled by field adapters rather than a hardcoded set of if/else statements based on hardcoded lists of fields. Of course, it's fine as a starting point for a proof of concept.

src/plone/restapi/renderer/markdown.py
```diff
@@ -0,0 +1,345 @@
...
+            parts.append("")
+
+        # Add body content
+        if body_parts:
+            parts.extend(body_parts)
+
+        return "\n".join(parts)
+
+    def _render_yaml(self, data, indent=0):
+        """Convert a Python dict to YAML format."""
```
davisagli, line 80: YAML serialization is hard to get right. For example, did you know that if the string "no" is serialized as `no` (unquoted), then it will be deserialized as boolean False?

So I would be inclined to use a battle-tested yaml library rather than roll our own.

src/plone/restapi/renderer/markdown.py
```diff
@@ -0,0 +1,345 @@
...
+            return block.get("description", "")
+
+        # Image blocks
+        elif block_type == "image":
+            url = block.get("url", "")
+            alt = block.get("alt", "")
+            if url:
+                return f"![{alt}]({url})"
+
+        # Return None for complex blocks
```
davisagli, line 319: Could be made extensible using adapters

src/plone/restapi/services/__init__.py
```diff
@@ -17,10 +19,53 @@ def render(self):
...
+                # Wildcard - use default JSON
+                continue
+
+            renderer = queryAdapter(self.request, IRenderer, name=mime_type)
+            if renderer is not None:
+                return renderer, renderer.content_type
+
+        # Fallback to JSON (default)
+        renderer = queryAdapter(self.request, IRenderer, name="application/json")
+        if renderer is not None:
```
davisagli: This should never be None, since you registered one in this package. So I think the fallback below is unnecessary.

src/plone/restapi/services/__init__.py
```diff
@@ -17,10 +19,53 @@ def render(self):
...
+            if mime_type.strip()
+        ]
+
+        # Try to find a renderer for each accepted type
+        for mime_type in accepted_types:
+            if mime_type == "*/*":
+                # Wildcard - use default JSON
+                continue
+
+            renderer = queryAdapter(self.request, IRenderer, name=mime_type)
```
davisagli: Maybe this should also adapt the service instance, so that it's possible to register a renderer for a specific service. (The markdown renderer here is registered for Interface, but is really only designed to work with the ContentGet service.)

src/plone/restapi/services/content/configure.zcml
```diff
@@ -7,13 +7,15 @@
 
   <plone:service
       method="GET"
+      accept="application/json,text/markdown,*/*"
```
davisagli: Does this actually do anything?

davisagli (commented):
@tisto Thanks for starting a proof of concept. I guess you want something to try out and not a detailed review at this stage. But here are a few thoughts I had from my first look at it.

tisto:
@davisagli my goal with this PR is to have a PoC that allows me to raise questions regarding architectural decisions that we would have to make to add markdown support to plone.restapi. Therefore, comments are welcome, but please do treat this as what it is: a vibe-coded PoC that we most likely will throw away and start from scratch, once we have a clear picture of where we want to go. :)

src/plone/restapi/renderer/markdown.py
tisto: That's a very good point. I am guessing that we only need the content endpoints to return markdown.

src/plone/restapi/renderer/markdown.py
tisto: I did not look at the Markdown generation code at all so far. I noticed that it does not look pretty though. Right now I am interested in looking at the architectural level. The details of the Markdown renderer is something that we can cover at a later point IMO.

src/plone/restapi/renderer/markdown.py
tisto: Good point. I am not fully sold on the YAML idea and I do not have much experience with using it on a programmatic level. Using a library is definitely something we should do if we decide on YAML.

PR #2015, opened 2026-04-09
Add backend support for plate block

### LINKED ISSUE #1998: Backend support for Plate block
There is proposal to add support for the Plate.js editor in Plone 7 (https://github.com/plone/volto/issues/8003).

This needs some backend support in plone.restapi for full feature parity with the existing slate block:

- [x] Linkintegrity. Add a IBlockFieldLinkIntegrityRetriever for the plate block that finds `url` inside an `a` node. (See https://volto.readthedocs.io/seven/reference/slate-to-plate-converters.html#links)
- [x] Resolve internal URLs to UIDs. Add a IBlockFieldSerializationTransformer and IBlockFieldDeserializationTransformer for the plate block to transform the `url` field inside an `a` node. --> already works, because of the generic transform for any field named `url`. added tests.
- [x] Text indexing. Add a IBlockSearchableText for the plate block that extracts searchable text.
- [x] Block discovery. Add an IBlockVisitor that knows how to find nested/inline blocks inside a plate block. --> already works, since it's in `obj.blocks`

davisagli:
Fixes #1998 

- [x] extract text for indexing
- [x] find links for linkintegrity
- [x] add tests to show resolveuid transforms work (they were already working because we have generic support for any field named `url`)
- [x] add IBlockVisitor to find sub-blocks inside a plate value

----
📚 Documentation preview 📚: https://plonerestapi--2015.org.readthedocs.build/


davisagli:

sneridagh (approved):
LGTM!!

no more plaintext key needed, right? I wonder if we could deprecate it too from Volto...

Now I have a doubt! Didn't we never have indexers for other blocks? for example, teasers?

davisagli:
> no more plaintext key needed, right? I wonder if we could deprecate it too from Volto...

I guess we could update the slate block indexer to collect text the same way.

> Now I have a doubt! Didn't we never have indexers for other blocks? for example, teasers?

It seems we do not. In the case of teasers, it's hard to say what is the right thing to do: including the teaser title in the searchable text could make it harder to find the actual page that is teased, and it would not necessarily be up to date when the teaser is dynamic.

btw, I realized there is one more thing I need to do: add an IBlockVisitor that can find sub-blocks inside a plate value.

PR #2020, opened 2026-05-29
Vendor code from plone.app.layout to avoid direct dependency on Classic UI

davisagli:

-----

Closes #

----
📚 Documentation preview 📚: https://plonerestapi--2020.org.readthedocs.build/


davisagli:

petschki (approved):
LGTM

jensens:
Hmm, wouldnt it be better to have one source of truth used by both, plone.app.layout and plone.restapi - deeper in the core then. Actually I am not sure where. But this smells a bit.

petschki:
That would be great, but I think its not easy because restapi is compatible with Plone 5.2 - 6.2

davisagli:
@jensens Yeah, I think you are right. Maybe they should go in CMFPlone?

@petschki is right that we still try to keep compatibility with Plone 5, but we can use conditional imports for that.

ale-rt:
Also to me this looks like something that should belong to `Products.CMFPlone`.

mauritsvanrees:
> Also to me this looks like something that should belong to `Products.CMFPlone`.

Problem with moving this to `Products.CMFPlone` is that we cannot do that now: in Plone 6.x, `plone.app.layout` still lives below `Products.CMFPlone`, so it cannot import from `Products.CMFPlone`.

But `Products.CMFPlone.browser.navigation.py` currently imports `buildFolderTree` from `plone.app.layout`.   We could mark this (with a simple comment) as the future canonical location for this function.  Then `plone.restapi` can today import if from `Products.CMFPlone`.  We can't show any deprecation warnings yet, but I don't think that is a real problem.

PR #2021, opened 2026-06-10
[codex] add default language sync for control panel

sneridagh:
## What changed
- Added a control-panel-only field deserializer for `ILanguageSchema.default_language`.
- Syncs the site language when `@controlpanels/language` changes `default_language` on non-multilingual sites.
- Leaves multilingual sites untouched.
- Split the language-specific tests into `test_services_controlpanel_language.py` and added a towncrier fragment.

## Why
- Control panel deserialization uses a fake Dexterity context, so the language control panel needed request-scoped behavior to update the actual portal language.

## Validation
- `.venv/bin/zope-testrunner --all --test-path=src -s plone.restapi -m 'test_services_controlpanel_language|test_services_controlpanels'

----
📚 Documentation preview 📚: https://plonerestapi--2021.org.readthedocs.build/

### DIFF
--- a/news/default-language-controlpanel-sync.bugfix
+++ b/news/default-language-controlpanel-sync.bugfix
@@ -1,0 +1,1 @@
+Patching `@controlpanels/language` now syncs the site language on non-multilingual sites when `default_language` changes, while multilingual sites keep their existing language state. @sneridagh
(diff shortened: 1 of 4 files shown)


sneridagh:

sneridagh:
@davisagli No clue why all these failures...

davisagli:
@sneridagh The desired feature here isn't specific to the REST API. It should happen when the default language is changed via any UI. So instead of implementing it in a deserializer, I think we should add this to what is taken care of by this event handler in plone.app.multilingual: https://github.com/plone/plone.app.multilingual/blob/master/src/plone/app/multilingual/subscriber.py#L184

PR #1921, opened 2025-05-12
Add a way for add-ons to modify the site endpoint response

davisagli:

PR #2022, opened 2026-06-11
Make the @site endpoint public

reebalazs:
## Problem

The `@site` endpoint provides the public bootstrap data a frontend needs before it can render anything: site title, logo, available/default languages, portal timezone, the `features` dict, and whatever `ISiteEndpointExpander` adapters contribute.

It is currently registered with `permission="zope2.View"`, checked against the site root. On sites that restrict the View permission on the site root to authenticated users — e.g. intranets such as the ones built with [kitconcept.intranet](https://github.com/kitconcept/kitconcept.intranet), where a restricted workflow grants View on published content only to `Authenticated` — anonymous requests to `/@site` return **401**.

This breaks frontend bootstrapping: Volto fetches `/@site` once at SSR time, typically while the user is still anonymous, and add-ons relying on expander data (e.g. `@kitconcept/volto-solr` reading `collective.solr.active` to decide which search UI to render) silently misbehave for the entire session, since the post-login redirect is a client-side route transition that never re-runs SSR.

Real-world manifestation: https://gitlab.kitconcept.io/kitconcept/kitconcept-intranet/-/work_items/127 (after Google login, the old search page shows instead of the Solr one).

## Fix

Register the `@site` service with `permission="zope.Public"`, matching its design contract: it serves only data that is needed before authentication and safe to disclose publicly.

- New test: anonymous `GET /@site` returns 200 even when View is revoked from Anonymous on the site root (fails with 401 before the change).
- Docs updated: removed the `zope2.View` mention, added a warning to the expander section that expander data is publicly visible.

Pinging @davisagli as the author of the `ISiteEndpointExpander` mechanism (#1921).

----
📚 Documentation preview 📚: https://plonerestapi--2022.org.readthedocs.build/

### DIFF
--- a/docs/source/endpoints/site.md
+++ b/docs/source/endpoints/site.md
@@ -10,7 +10,7 @@ myst:
 # Site
 
 The `@site` endpoint provides general site-wide information, such as the site title, logo, and other information.
-It uses the `zope2.View` permission, which requires appropriate authorization.
+It is accessible to anonymous users, so that a frontend can fetch the data it needs to bootstrap itself, even on sites that require authentication to view content.
 
 Send a `GET` request to the `@site` endpoint:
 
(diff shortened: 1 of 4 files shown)


davisagli (approved):
Thanks!

davisagli:

PR #2024, opened 2026-06-19
Handle VHM in the `@querystring-search` endpoint like it is done in `@search`

### LINKED ISSUE #2023: `@querystring-search` does not handle Virtual Host Monster (VHM) mapped paths
The `/@querystring-search` endpoint fails to return results when queried with paths relative to a virtual host root. While the standard `/@search` endpoint correctly resolves virtual paths to physical catalog paths using the `VirtualRootPhysicalPath` request variable, `/@querystring-search` passes the criteria directly to the query builder without normalization.

This causes modern frontend components (like the new Svelte-based `pat-filemanager`) to show empty listings when navigating into subfolders in a production environment where VHM is active.

### Steps to Reproduce
1. Set up a Plone site behind a Virtual Host Monster (e.g., mapping `http://plone.org/` to `/Plone`).
2. Create a folder at `/Plone/folder`.
3. Call `/@querystring-search` via the virtual URL `http://plone.org/@querystring-search` with the following JSON:
   ```json
   {
     "query": [
       {
         "i": "path",
         "o": "plone.app.querystring.operation.string.path",
         "v": "/folder::1"
       }
     ]
   }
4. Expected: The endpoint returns the children of /Plone/folder.
5. Actual: The endpoint returns 0 results because it looks for a physical path /folder, which does not exist in the catalog.

### Technical Details
The issue lies in plone.restapi.services.querystringsearch.get.QuerystringSearch. Unlike `plone.restapi.search.handler.SearchHandler`, it does not check for `VirtualRootPhysicalPath` in the request.
In `SearchHandler._constrain_query_by_path`, the following logic is applied:

```python
vhm_physical_path = self.request.get("VirtualRootPhysicalPath")
if vhm_physical_path:
    # ... prepends vhm_physical_path to the query ...
```

The @querystring-search implementation should adopt similar logic to ensure consistency across search services.

### Proposed Solution
Update `QuerystringSearch.__call__` in `src/plone/restapi/services/querystringsearch/get.py` to:
1. Detect `VirtualRootPhysicalPath`.
2. Iterate through query criteria.
3. For any path index, prepend the physical root to the value (handling both string and list formats).
4. Preserve the ::depth suffix if present in the string value.

erral:
- [x] I signed and returned the [Plone Contributor Agreement](https://plone.org/foundation/contributors-agreement), and received and accepted an invitation to join a team in the Plone GitHub organization.
- [x] If needed, I added [documentation](https://6.docs.plone.org/contributing/documentation/index.html) for my changes.
- [x] I included a [change log entry](https://6.docs.plone.org/contributing/index.html#contributing-change-log-label) in my commits.

-----

Closes #2023

Refs https://github.com/plone/mockup/issues/1609 where is shown that new pat-filemanager shows empty contents of a folder when deployed into production with VHM.

----
📚 Documentation preview 📚: https://plonerestapi--2024.org.readthedocs.build/

### DIFF
--- a/docs/source/endpoints/querystringsearch.md
+++ b/docs/source/endpoints/querystringsearch.md
@@ -39,6 +39,14 @@ The server will respond with the results that are filtered based on the query yo
 :language: http
 ```
 
+### Virtual Host Monster Support
+
+When accessed through a Virtual Host Monster (VHM), the endpoint automatically resolves virtual paths provided in the `query` criteria to their physical counterparts in the catalog.
+
+For example, if your Plone site is physically located at `/Plone` but served at `http://plone.org/`, a query for `v: "/folder"` will be automatically expanded to `/Plone/folder` before being passed to the catalog.
+
+This expansion applies to any value in a `path` criterion that starts with a `/`. It also correctly handles the `::depth` suffix.
+
 Parameters the endpoint will accept:
 
 - `query` - `plone.app.querystring` query, required
--- a/docs/source/endpoints/searching.md
+++ b/docs/source/endpoints/searching.md
@@ -204,6 +204,14 @@ Be aware that this might induce performance issues when retrieving a lot of reso
 Normally the search just serializes catalog brains, but with `fullobjects`, we wake up all the returned objects.
 ```
 
+## Virtual Host Monster Support
+
+When accessed through a Virtual Host Monster (VHM), the `/@search` endpoint automatically resolves virtual paths provided in the `path` parameter to their physical counterparts in the catalog.
+
+For example, if your Plone site is physically located at `/Plone` but served at `http://plone.org/`, a query for `path=/folder` will be automatically expanded to `/Plone/folder` before being passed to the catalog.
+
+This expansion applies to both string and list values for the `path` parameter, provided they start with a `/`.
+
 ## Restrict search results to Plone's search settings
 
 By default, the search endpoint does not exclude any types from its results.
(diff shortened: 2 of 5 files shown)


erral:

MrTango (approved):
i can confirm that this fixes the roblems with pat-filemanager thx a lot

davisagli (changes requested):
Doesn't this belong in https://github.com/plone/plone.app.querystring/blob/master/src/plone/app/querystring/queryparser.py instead of here? The handling of paths in queries should be the same whether it's done in the REST API or Blicca. "Patching" the query like this before passing it to the main query parser feels like an anti-pattern.

erral:
That's what we are doing in the [`@search` endpoint](https://github.com/plone/plone.restapi/blob/main/src/plone/restapi/search/handler.py#L46) that's why I brought this here.

davisagli:
@erral The search endpoint isn't using a query builder that is shared with non-RESTAPI code. This is.

petschki:
It feels indeed a bit strange, that we have to do this here, whereas the "normal" querystring API usage doesn't ... I remember when working on `pat-contentbrowser` we had to implement VHM support in the pattern options to fix `getVocabulary` calls (which relateditems pattern did also btw and maybe `pat-querystring` does also some magic with its options) ... OTOH I wonder why it didn't came up earlier, that the restapi querystring endpoint doesn't support VHM paths ... so I would like to check the queryparser first, before merging this.

PR #2043, opened 2026-08-27
Fix expected test output due to new plone.app.discussion profile

mauritsvanrees:
An action run on the main branch currently fails.
See https://github.com/plone/plone.restapi/actions/runs/33066530077/job/98497855546

### DIFF
--- a/news/+cc5f434b.tests.md
+++ b/news/+cc5f434b.tests.md
@@ -1,0 +1,1 @@
+Fix expected test output due to more recent `plone.app.discussion` profile.  @mauritsvanrees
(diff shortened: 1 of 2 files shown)


mauritsvanrees:
The important check now succeeds:
https://github.com/plone/plone.restapi/actions/runs/33067211635/job/98500159067?pr=2043
I merge.

davisagli:
@mauritsvanrees Thank you.

I should probably change https://github.com/plone/plone.restapi/blob/main/constraints.txt to pin a specific version instead of 6.2-dev so we don't get surprise failures like this.

PR #2042, opened 2026-08-26
Fix the security check when a wrong permission is specified

ale-rt:
Also add a warning to support developers in fixing the permission checks.

Fixes #2041

PR #2044, opened 2026-08-31
Fix the security check when a wrong permission is specified [9.x.x]

mauritsvanrees:
Backport of PR #2042 for Plone 6.1 and earlier.

And update the `http-examples` in the documentation to be correct for Plone 6.1.

Note that the lint check may fail because `pyupgrade` wants to remove these lines from `setup.py`:

```
-if sys.version_info.major == 2:
-    raise ValueError(
-        "plone.restapi 10 requires Python 3. "
-        "Please downgrade to plone.restapi 7 for Python 2 and Plone 4.3/5.1."
-    )
-
-
```

And afterwards `flake8` then complains:

```
setup.py:5:1: F401 'sys' imported but unused
```

I think we want to keep those lines on this 9.x.x branch.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,7 +19,7 @@ jobs:
       matrix:
         include:
           - python-version: "3.13"
-            plone-version: "6.2"
+            plone-version: "6.1"
     steps:
       # git checkout
       - uses: actions/checkout@v5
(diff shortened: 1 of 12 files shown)


mauritsvanrees:
For good measure I am testing this together with the similar https://github.com/plone/plone.autoform/pull/94 on Jenkins.

mamico (approved):

PR #2045, opened 2026-09-04
Drop metadata_fields from the catalog query

### LINKED ISSUE #2031: fix plone.restapi to not pass the metadata_fields into the query
plone.restapi's @search passes the entire request dict to the catalog ([handler.py#L102](https://github.com/plone/plone.restapi/blob/main/src/plone/restapi/search/handler.py#L102)), including control parameters such as metadata_fields (e.g. every request from Volto's ObjectBrowserWidget). These are not catalog indexes.

see also https://github.com/bluedynamics/plone-pgcatalog/pull/179

cc @davisagli

kunalKumar-13:
Fixes #2031.

### Root cause

`@search` hands the whole request dict to `catalog.searchResults()`, so control
parameters that are not catalog indexes go with it. `metadata_fields` is one: it
is read from `request.form` (or the JSON body) by the summary serializer, never
from the query dict, and every request from Volto's `ObjectBrowserWidget`
carries it.

`ZCatalogCompatibleQueryAdapter` already had a name for this class of key,
`ignore_query_params`, but it only used it to suppress the "No such index"
warning — the key was still left in the query it returned. ZCatalog ignores
unknown keys, which is why nothing breaks in stock Plone; a catalog that
validates its query instead of ignoring what it does not recognise rejects it.

### Change

Pop the `ignore_query_params` keys at the top of `__call__`, so the attribute
does what its name says. That makes the guard further down redundant.

### Why this is safe

Checked before changing anything that `metadata_fields` is not read back from
this dict. `services/search/get.py` passes `request.form.copy()`, and
`SummaryJSONSerializer.metadata_fields()` reads `request.form` or re-parses
`request["BODY"]` itself — so dropping it from the query cannot affect the
metadata that comes back. The three existing end-to-end tests for partial, full
and `fullobjects` metadata retrieval confirm it.

### Tests

- `test_zcatalogcompatiblequeryadapter_drops_metadata_fields` — the adapter's
  output no longer contains the key.
- `test_search_handler_does_not_pass_metadata_fields_to_catalog` — asserts on
  the kwargs that actually reach `catalog.searchResults()`.

### Measured

Plone 6.2.1 / Python 3.13, `zope-testrunner --all`, `TZ=UTC`:

- `main`, before the change: **1406 tests, 0 failures, 0 errors, 13 skipped**
- with the fix and the two new tests: **1408 tests, 0 failures, 0 errors**
- reverting only `search/query.py` and keeping the tests: **1408 tests, 2
  failures** — the two new tests and nothing else

(Note for reproduction: `test_date_index_query` and
`test_full_metadata_retrieval` fail on an unpatched checkout when the machine is
not on UTC — they compare rendered timestamps against fixed UTC strings. Run
with `TZ=UTC`.)

### DIFF
--- a/news/2031.bugfix.md
+++ b/news/2031.bugfix.md
@@ -1,0 +1,7 @@
+Stop passing `metadata_fields` into the catalog query. It is a control parameter
+read from the request by the summary serializer, not a catalog index, and every
+request from Volto's `ObjectBrowserWidget` carries it. `ignore_query_params` in
+`ZCatalogCompatibleQueryAdapter` only suppressed the "No such index" warning; the
+key still reached `catalog.searchResults()`, where ZCatalog silently ignores it
+but a catalog that validates its query does not.
+@kunalKumar-13
--- a/src/plone/restapi/search/query.py
+++ b/src/plone/restapi/search/query.py
@@ -78,6 +78,10 @@ class ZCatalogCompatibleQueryAdapter:
         "sort_order": {list: list, tuple: list, str: str},
     }
 
+    # Control parameters that are not catalog indexes. They are read from
+    # the request where they are needed -- ``metadata_fields`` is read from
+    # ``request.form`` (or the JSON body) by the summary serializer -- and
+    # are dropped here so they never reach ``catalog.searchResults()``.
     ignore_query_params = ["metadata_fields"]
 
     def __init__(self, context, request):
@@ -100,6 +104,9 @@ class ZCatalogCompatibleQueryAdapter:
                 return future_value(idx_query)
 
     def __call__(self, query):
+        for param in self.ignore_query_params:
+            query.pop(param, None)
+
         for idx_name, idx_query in query.items():
             if idx_name in self.global_query_params:
                 # It's a query-wide parameter like 'sort_limit'
@@ -115,8 +122,7 @@ class ZCatalogCompatibleQueryAdapter:
             # that could not be serialized in a query string or JSON
             index = self.get_index(idx_name)
             if index is None:
-                if idx_name not in self.ignore_query_params:
-                    log.warning("No such index: %r" % idx_name)
+                log.warning("No such index: %r" % idx_name)
                 continue
 
             query_opts_parser = getMultiAdapter(
--- a/src/plone/restapi/tests/test_search.py
+++ b/src/plone/restapi/tests/test_search.py
@@ -10,6 +10,7 @@ from plone.app.textfield.value import RichTextValue
 from plone.dexterity.utils import createContentInContainer
 from plone.registry.interfaces import IRegistry
 from plone.restapi.bbb import INavigationRoot
+from plone.restapi.search.handler import SearchHandler
 from plone.restapi.search.query import ZCatalogCompatibleQueryAdapter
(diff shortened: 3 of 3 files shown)


kunalKumar-13:

kunalKumar-13:

kunalKumar-13:
Flagging a red Jenkins job on this PR myself, along with what I have and have not been able to rule out.

`pull-request-6.3-3.10` (build [133](https://jenkins.plone.org/job/pull-request-6.3-3.10/133/)) fails on a CMFPlone acceptance test, not a plone.restapi one:

```
Test Contentbrowser
Scenario: add contenttype Link with an internal target                | FAIL |
Text 'My Link' (str) should be 'Test Folder' (str)
```

The other three Jenkins jobs pass: `6.2-3.10`, `6.2-3.14` and `6.3-3.14`.

**Why I do not think this change causes it.** The concern would be that dropping `metadata_fields` from the query starves the summary serializer, since the content browser is exactly the widget that sends it. It does not, because the serializer never reads that key from the query:

- `SearchGet.reply()` does `query = self.request.form.copy()` — the handler works on a **copy**, so popping from it cannot affect the request.
- `ISerializeToJson`'s `metadata_fields()` (`serializer/summary.py:119`) reads `self.request.form`, or `json_body(self.request)` for a POST body, never the query dict.
- `json_body()` re-parses `request.get("BODY")` on each call and returns a fresh dict, so there is no shared object to mutate either.

So the key is still present everywhere it is read; the only thing that changes is that it no longer reaches `catalog.searchResults()`.

**What I could not establish**, and would rather say than gloss over: I have no passing comparison for that test. It did not run in the recent builds of the same job for `zodbverify#18`, `plone.folder#66` or `Products.CMFPlone#4369`, nor in the older plone.restapi builds ([118](https://jenkins.plone.org/job/pull-request-6.3-3.10/118/), [97](https://jenkins.plone.org/job/pull-request-6.3-3.10/97/), [89](https://jenkins.plone.org/job/pull-request-6.3-3.10/89/)), nor in the latest `plone-6.3-python-3.10` and `plone-6.3-python-3.10-robot-chrome` nightlies. I checked the full console text in each case rather than a summary. It has run twice, on builds 127 and 133, both of them this PR, and failed both times — so "flaky" is my reading, not something I have shown.

Re-running to get a third data point. If it fails again I will dig into the robot test properly rather than assume.

kunalKumar-13:
Third run came back green — [build 134](https://jenkins.plone.org/job/pull-request-6.3-3.10/134/) passed, so the Contentbrowser failure was flaky after all. All four Jenkins jobs are green now.

jensens (approved):

kunalKumar-13:
Thanks for the review @jensens.

GitHub had this as `BEHIND`, so I've rebased onto current `main` — one commit, and nothing upstream had touched `search/query.py` or `tests/test_search.py`, so the change is unchanged.

kunalKumar-13:
Jenkins came back green on all four:

```
pull-request-6.2-3.10   success
pull-request-6.2-3.14   success
pull-request-6.3-3.10   success
pull-request-6.3-3.14   success
```

So this is approved, rebased onto current `main`, changelog entry present, and green across both Plone versions and both Pythons. Ready to merge whenever someone has a moment — thanks again for the review @jensens.

PR #2047, opened 2026-09-09
Prevent automatic effective dates after expiration

teekuningas:
- [x] I signed and returned the [Plone Contributor Agreement](https://plone.org/foundation/contributors-agreement)
- [ ] ..., and received and accepted an invitation to join a team in the Plone GitHub organization.
- [x] If needed, I added [documentation](https://6.docs.plone.org/contributing/documentation/index.html) for my changes.

-----

I think `@workflow/publish` should not set `effective` to now when `expires` is already in the past. Otherwise, the transition succeeds but subsequent PATCH requests fail validation because `effective` is after `expires`.

The automatic effective-date assignment has been there for a long time though:

https://github.com/plone/plone.restapi/pull/761


davisagli (changes requested):
I think it makes sense to do something about this edge case, but I think it would be better to skip setting the effective date in this case, instead of raising a BadRequest which blocks the publish transition.

teekuningas:
Thank you for the answer.

After looking into how Plone Classic handles this, I found that it has exactly the same issue:

https://github.com/plone/plone.app.dexterity/issues/325

Given that, changing only plone.restapi either to reject the transition or skip setting the effective date would make the behavior inconsistent. I think this should be addressed in Plone Classic first and may require deeper consideration of the consequences.

Sorry about the noise.

